PR #22, opened 2019-04-26
install.sh and uninstall.sh updates

BIBOLV:
Please check following editions
In install.sh:
updated help text
added check of passed arguments
add --no-udev-rule argument
added check of nologin shell and fallback to /bin/false if none found 
changed useradd cmd to work on different distros

In uninstall.sh:
added missing removal of files/dirs

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,17 +1,15 @@
 #!/bin/bash
 
-if [ "$1" = "" ]
+if [ "$1" == "" ] || [ "$1" == "-h" ]
 then
-    echo Usage: install.sh [binary] [root]
-    echo Example: install.sh build/wmbusmeters /
+    echo - n "Usage: install.sh [binary] [root] [OPTIONS]
+    Example: install.sh build/wmbusmeters /
+	
+	Options:
+	--no-adduser		Do not add wmbusmeters user
+	--no-udev-rules		Do not add udev rules
+	"
     exit 0
-fi
-
-ADDUSER=true
-if [ "$1" = "--no-adduser" ]
-then
-    ADDUSER=false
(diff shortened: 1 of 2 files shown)


majekw:
I agree that install/uninstall scripts are a mess :-)
But blindly removing config and all logs is a bad idea.
During installation config should be put in file with extension like `default` (or dist, etc).
During removal script should only remove default config and remove directories **if empty**.
Another story is a great number of errors pointed by shellcheck in these scripts :-(

BIBOLV:
Agree about existing uninstall.sh, correctly would be to rename it to purge.sh as that`s what it does.
Maybe also better would be to add uninstall part to install script itself? It would make install script bigger, but everything would be in one place.

weetmuts:
Thank you very much @BIBOLV for the fixes to the install and uninstall scripts!
I can merge the pull request right away if you like.

Majekw, as for uninstall, my intent is to mimic debian uninstall of packages, I just have not
had time to do it yet. So that a simple uninstall.sh only removes the binaries.
But an "uninstall.sh --purge" to get rid of the config files and logs.

Or as you say Bibo, put them both in install. But I think it is slightly easier to have to scripts for this.
The scripts are meant to be called from the Makefile or from within the future debian package installer.

weetmuts:
I just merged it because the --system fix to adduser etc was necessary anyway.
Lets work on a new pull request that improves the uninstall/install features!

BIBOLV:
NP @weetmuts 
Could you please explain in few words your idea about uninstall.sh - I haven`t built any .deb packages, just .rpm`s ;)

PR #24, opened 2019-05-01
Initial support for Techem Vario 4 heat meters

FeatureExpert:
Hello again weetmuts. This is my first pull request ever. I made it just for fun as my C++ knowlegde is zero :) This is initial support for Techem Vario 4 Typ 4.5.1 heat meters. Some basic values can be extracted from T telegrams.

Feel free to reject this PR and manually integrate code into your project as it contains some minor changes for my own use and probably some other things are made in a wrong way.


weetmuts:
Excellent work! I could merge this directly, but I am just curious about what unit we should standardize on for heat meters. 

The existing mult302 heat meter reports values as kwh but the techem seems to use GigaJoule instead. Part of the work done by wmbusmeters is to normalize the enormous number of possible units in the wmbus protocol (read wmbus.cc vif units and vif extension types) to as few units as possible. 

I suppose the proliferation of units is because the protocol wants to use the fewest number of bits for transferring the information from the meter. But not only is there all 10 exponential versions of cubic meters but also 0.1 cubic feets and gallons and american gallons...... When reporting the values
in json we are not limited to a fixed number of bits nor a limited number of decimals. 

I think that wmbusmeters should narrow down the number of units. This way, if you have
configured your HASS/Domoticz/iobroker that gets data from wmbusmeters, and you switch
from one heat meter type to another heat meter type, you should not have to change
the unit/scale on the HASS/Domoticz/iobroker side.

So you could add previousEnergyConsumption to HeatMeter (or rather targetEnergyConsumption
so that we reuse the terminology for meter readings when this month began) and have 
multical302 always return 0 for this. Have the new techem meter inherit from HeatMeter
and always return 0 for total_volme_m3. Then convert GJ to kwh for the techem meter.
Or change HeatMeter and therefore multical302 from kwh to report GJ.

So, what do you think, kWh or GJ?

In the future we could add options to wmbusmeters to automatically scale the output
to any desired type, if you would like to read a multical21 who reports the values as liters
but get deca-gallons in the json. ;-)

weetmuts:
It might be natural to use GJ for heat meters but kWh for electricity meters.....

weetmuts:
Or maybe, all meters that measure energy should have the same unit, irrespective
if it is heat or electricity.....

FeatureExpert:
It's hard to say. Maybe it should be meter dependant. I really don't care about heat energy usage in kW/h as I'm paying for GJ. In my case it's better and easier if I get readings from my meter in GJ, because I can easly compare it with the bill without any additional calculations...

weetmuts:
Ok, I read the manual for Multical302, it can be configured to report the consumption
as GJ or kWh or MWh. So, the preference  for which unit to use, seems to
be pretty arbitrary. Better let the user of wmbusmeters decide. 

The default unit for energy is going to be kWh, for no other reason, that it happened to be implemented first and that unit is first in the wmus specification.

So, I will import your pull request and then add the option --addconversion=GJ
(and the conf file setting addconversion=GJ) which means that in addition to 
meter_total_kwh in the json and METER_TOTAL_KWH in the env variable, 
wmbusmeters will add meter_total_gj to the json and METER_TOTAL_GJ.

weetmuts:
Hi FeatureExpert! I have now added support for additional units.
Try with --addconversions=GJ on the command line
and addconversions=GJ in the conf file.

It took quite a rewrite to do this. But that was something I had thought to do anyway.
Now the source to a new meter is much shorter. :-) Check the source to vario451.

PR #47, opened 2019-11-20
Add a script to import meter information from a Kamstrup KEM file

mbursa:
A Python script that decrypts the content of Kamstrup KEM file and imports meters to wmbusmeters' config folder.

The KEM file is a (sometimes zipped) xml file that contains xml-encrypted data using the xml-enc standard (http://www.w3.org/2001/04/xmlenc). The password needed to decrypt the xml-encrypted data can either be the CustomerId or something else selected by the person that created the KEM file using Kamstrup software.

This script takes the encrypted KEM file and decrypts its content. The result is also a XML file with a list of meters. The script prints the information about each meter to the console and populates wmbusmeters' config folder with corresponding meter files. Optionally, decrypted KEM file content can be saved to a file (option -o).

Usage: python kem-decryptor.py [options] <kem_file> <password>
kem_file ... the name of the KEM file to decrypt (you need to unzip the the KEM file first, if it is zipped)
password ... original password used to encrypt the content (16 characters maximum)
options ... use -h switch to get a list of options with their descriptions

The script is Python2/Python3 compatible.

### DIFF
--- a/utils/kem-import.py
+++ b/utils/kem-import.py
@@ -1,0 +1,148 @@
+# A Python script to decrypt the content of Kamstrup KEM file.
+#
+# The KEM file is a (sometimes zipped) xml file that contains xml-encrypted data using 
+# the xml-enc standard (http://www.w3.org/2001/04/xmlenc). The password needed to decrypt 
+# the xml-encrypted data can either be the CustomerId or something else selected by the 
+# person that created the KEM file using Kamstrup software.
+#
+# This script takes the encrypted KEM file and decrypts its content. The result is also 
+# a XML file with a list of meters. The script prints the information about each meter to
+# the console and populates wmbusmeters' config folder with corresponding meter files. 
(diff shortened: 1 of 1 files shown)


weetmuts:
Very nice! Thank you!
Is it easy to do unzip as well from python?

mbursa:
Yes, it is quite easy to unzip a file in python and it can be included in the script as well. 

In my case, I received a file called "20191114_1227_selected_DownloadMeters.zip.kem" which obviously was created by the Kamstrup software. Despite of mixed-up extension, it was a zip file containing two other files, one of them being the encrypted KEM file (the other was a xml schema). The script would have to test the input file, eventually unzip it, check the file names and pick the KEM file. It is not very difficult. 

The reason I have not done it is that I only have one example of such an output and I do not know if it is generic enough. That is why I think it may be better if the user inspects the file manually and eventually extracts the zipped kem file before running the script.

I will give it a second thought.

PR #52, opened 2019-12-03
Kem file import with ZIP support

mbursa:
I have added zip file support. The script now makes a check if the input file is a zip file. If yes, it opens the zip and searches for .kem file in it.

The second commit adds a readme file into utils folder explaining the purpose of individual utilities.

The third commit updates a paragraph in README.md that refers to KEM files.


weetmuts:
Very nice! Thanks!

PR #60, opened 2020-01-19
a fix for cmdline override in daemon mode

mbursa:
tested, a small fix is needed

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -197,13 +197,13 @@ fi
 ## Create /etc/systemd/system/wmbusmeters.service
 ##
 
-if [ ! -f "$ROOT"/etc/systemd/system/wmbusmeters.service ]
+if [ ! -f "$ROOT"/etc/systemd/system/wmbusmeters@.service ]
 then
     mkdir -p "$ROOT"/etc/systemd/system/
     # Create service file
-    cat <<EOF > "$ROOT"/etc/systemd/system/wmbusmeters.service
+    cat <<EOF > "$ROOT"/etc/systemd/system/wmbusmeters@.service
 [Unit]
-Description=wmbusmeters service
+Description="wmbusmeters service on %i"
 After=network.target
 StopWhenUnneeded=true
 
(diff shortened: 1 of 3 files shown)


mbursa:
The second commit is supposed to address a situation when the usb dongle is unplugged. If that happens the daemon keeps running. The PR implements `working()` method that checks if the virtual file still exists. Idea taken from [https://stackoverflow.com/questions/44010480/c-detecting-usb-serial-device-plugged-unplugged](here).

mbursa:
This third commit deals with udev rules and systemd service. It changes the service file to a template (adding @ to its name). Udev rule then starts a separate instance for each dingle that is plugged in and passes the name of the device as a parameter of the template service. Based on [this post](https://www.stevenrombauts.be/2019/01/run-multiple-instances-of-the-same-systemd-unit/).

weetmuts:
Very nice! Thanks! I just pushed the 1 to i fix.

I just tested your new working() using statvfs, and it does indeed speed up the detection of the failing device. However on my computer, unplugging an im871a dongle properly stops the daemon, without the fix.

Unplugging the rtlsdr dongle however, causes rtlsdr to go into a tailspin, which causes the daemon
to hang. Does this solve that problem?

So I will probably include your fix, but why are you using statvfs instead of plain stat?

Can you also add:
 if (fd_ == -1) return false; 
first to your working() function.

mbursa:
When I unplug im871a the daemon stays in a loop printing "(serial) received binary" in the log (`read()` gives zero bytes) and never exits. Tested on raspbian and ubuntu.

Regarding `statvfs`, I simply followed the exaple given in the link. Thinking more about it, perhaps using `fstat` on the file descriptor will be better together with `(fd_ == -1)` test. I will test it and eventually submit a new commit. So do not mergit yet.

weetmuts:
You could reuse the util function, like this:
bool SerialDeviceTTY::working()
{
    if (fd_ == -1) return false;

    bool working = checkCharacterDeviceExists(device_.c_str(), false);

    if (!working) {
        debug("(serial) device %s is gone\n", device_.c_str());
    }

    return working;
}

weetmuts:
I pushed some helpful comments from install.sh to the person installing the new server/udev files.
It will notify the user on how to reload the systemd and/or udev if necessary.

Also, I did test the fstat and it does not work for me. There is no speedup in detection of the failing usb device. I change to checkCharacterDeviceExists instead.

Can you check that it works for you?

mbursa:
I tested the merged changes, it works well in my setup. checkCharacterDeviceExists does the job.
Thanks for your help on it!

PR #75, opened 2020-02-11
Add Techem FHKV data II/III

cinemarene:


cinemarene:
I don't know, why github is not able to show the diffs correctly. Encoding of files is the same.

cinemarene:
There was a bad mistake. I'll create a new PR.

PR #84, opened 2020-02-24
Fix wmbusmetersd symlink

skazi0:
When building the package, install script creates incorrect symlink
pointing to the build directory instead of target root.


weetmuts:
The purpose of $ROOT is to install into a subdirectory, in preparation of doing the deb.
I suppose the symbolic link should be relative?

skazi0:
@weetmuts the command syntax is following `ln -s <target> <link>` where `<link>` is the place where the link will be created (this has to include `$ROOT`) but `<target>` is used "as is" so it needs to be the path in the target filesystem (without `$ROOT`).

weetmuts:
Ok, thanks!

PR #87, opened 2020-03-02
Make easier to packaging wmbumeters for linux

dwrobel:
Makefile:
 - Allow to disable stripping of executables as this prevents from
   properly creating debug information sub packages,

 - Don't invoke any git commands when building from released
   tarball. Moreover, when creating an rpm package with additional
   patches then the created git repository is the one in which rpm
   applies patches and not the original one.

 - Allow to pass CXXFLAGS and LDFLAGS as distributions requires all
   packages to be compiled with the same set of flags.

 - Support to override DESTDIR when installing files.

 - Allow to pass additional arguments (EEXTRA_INSTALL_OPTION)
   for install.sh script from Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,14 +20,16 @@
 # make DEBUG=true
 # make DEBUG=true HOST=arm
 
+DESTDIR?=/
+
 ifeq "$(HOST)" "arm"
     CXX=arm-linux-gnueabihf-g++
-    STRIP=arm-linux-gnueabihf-strip
+    STRIP?=arm-linux-gnueabihf-strip
     BUILD=build_arm
 	DEBARCH=armhf
 else
     CXX=g++
-    STRIP=strip
+    STRIP?=strip
 #--strip-unneeded --remove-section=.comment --remove-section=.note
     BUILD=build
 	DEBARCH=amd64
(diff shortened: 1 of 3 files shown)


dwrobel:
I created this and other PRs (https://github.com/weetmuts/wmbusmeters/pull/85, https://github.com/weetmuts/wmbusmeters/pull/86) as I'm in a process of packaging `wmbusmeters` for Fedora [here](https://copr.fedorainfracloud.org/coprs/dwrobel/wmbusmeters/). I would appreciate if you could review those changes and merge if possible.

weetmuts:
Nice! Thank you! 

But /usr/lib/systemd/system does not exist on my Ubuntu system.
It seems like the location is /lib/systemd/system can you add detection of this?

Can you add an rpm target in the makefile?

weetmuts:
The rpm target should probably create a src.rpm ? I will try to change the makefile to generate a source deb.

dwrobel:
>But /usr/lib/systemd/system does not exist on my Ubuntu system.

Do you have maybe `/lib/systemd/system` - that will be also fine.

weetmuts:
Yes, /lib/systemd/system exists on my Ubuntu system.

weetmuts:
I tested your branch and it seems to work ok on my Ubuntu system. :-)

PR #92, opened 2020-03-07
Explain how to get started with the AMB8465 stick

hollie:
I did run in some issues to get started with the AMB8465 stick because the stick was not configured in default mode.

I'm adding this info for others to avoid they run into the same issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,14 @@ If you are using rtl_sdr/rtl_wmbus and you want to stop the daemon, do
 
 If you are using auto, then start manually with --debug to see how wmbusmeters goes looking for devices.
 
+## AMB8465 USB stick
+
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks! What tool do you use to send the reset string?

hollie:
Hello @weetmuts,

it is me who has to thank you for releasing this software. I'm in the process of upgrading the measurement of all my utilities at home. Your software makes that I can integrate my Kamstrup meters for rain water and drinking water in my monitoring system in half an hour, instead of having to write the code myself. So: a big thanks for that!

For the hex strings I use the best serial terminal software ever: [CoolTerm from Roger Meiers](https://freeware.the-meiers.org). Cross-platform and has all the features you would want. Sending a hex string is just 2 clicks away: 
Send_String__Untitled_0__en_Untitled_0

Have a nice day!
 Lieven

weetmuts:
Cool!

PR #103, opened 2020-04-01
Add wmbusmeters availability info on Fedora

dwrobel:


weetmuts:
Very nice! Thank you!

PR #105, opened 2020-04-21
Adds FreeBSD support

ZeWaren:
**Adds FreeBSD support:**

- Uses `/bin/sh` instead of `/bin/bash` to support systems without bash.
- Changes a few preprocessor conditions to use the same things OSX uses instead of the Linux idioms.
- Allows using the default compiler of the system instead of forcing g++.
- Add support for LLVM in debug mode

Tests pass in normal mode, using `g++` or `clang++` as a compiler.
Tests pass in debug mode using `clang++` as a compiler. Debug mode can't be build with `g++` since FreeBSD's gcc doesn't provide asan support. Generating gcov files works correctly using `llvm-cov gcov` instead of gcc's tool.

The application runs fine on my FreeBSD machine. I can read the values of my water meter correctly:

	(telegram) DLL L=19 C=44 (from meter SND_NR) M=4c30 (SAP) A=20d78c1e VER=d4 TYPE=01 (Oil meter)
	(telegram) TPL CI=a2
	(izar) log "telegram=|1944304C1E8CD720D401A2|7312001337F24C33683FB997735AB9|+24"
	(izar) Decoded PRIOS data: 4BEF9001000E8D01008124
	(izar) 00: 19 length (25 bytes)
	(izar) 01: 44 dll-c (from meter SND_NR)
	(izar) 02: 304c dll-mfct (SAP)
	(izar) 04: 1e8cd720 dll-id (20d78c1e)
	(izar) 08: d4 dll-version
	(izar) 09: 01 dll-type (Oil meter)
	(izar) 0a: a2 tpl-ci-field (Mfct specific)
	Jambon  20d78c1e        102.639 m3      101.646 m3      2020-04-21 18:42.10


weetmuts:
Thanks! Can you change the == to = in the test_...._.sh scripts inside test?
Then I think the tests in travis will work as well!

weetmuts:
If fixed the shell issues myself and merged your pull request.

Many thanks for the port! 

However, it was very educational...again...to experience the different shell behaviors
on different platforms. Well, in this case the MacOSX that caused problems.
This is why I usually standardize on bash for my scripts.

However the scripts in wmbusmeters are really simple, so I will let them
stay as /bin/sh for the moment.

Again thanks for the work! And I am happy that wmbusmeters works for you!

weetmuts:
By the way, if you do develop in c++, the libasan address sanitizer is very very useful!
It makes developing C++ as quick as developing Java. You get stacktraces when you have segfaults,
no need to even start the debugger. It detects memory overwrites, arrays out of bounds, not freed memory etc etc . It is marvellous. :-)

ZeWaren:
Great!

+1 for address sanitation. It's good that it's also built-in with LLVM, so non GNU systems can benefit from it as well.

PR #111, opened 2020-05-08
amb8465: Print checksum errors from responses

HenryNe:


weetmuts:
Thanks! Have you ever experienced a broken checksum?

HenryNe:
> 
> 
> Thanks! Have you ever experienced a broken checksum?

No. Checksum from cmd response was all tmes ok.
On startup I have seen data packages out of sync, and decoded as complete nonsense. FF from config response was detectet as command response.

PR #112, opened 2020-05-08
amb8465: Fix RSSI in responses

HenryNe:
Bugfix handling RSSI in amb8465:
* Update config RSSI_Enable = false. Was always true.
* RSSI is only in CMD_DATA_IND, not in all responses

New:
* Update rssi from last byte of transparant data
* Print rssi in dBm
* Print config UART_CMD_OUT_ENABLE


weetmuts:
Very nice! Thank you!

PR #124, opened 2020-05-27
Switch to ubuntu 18.04LTS

dwrobel:
Update compiler options:
 - remove -fmessage-length=0 as it's default,
 - remove -Wno-unused-function - does not introduce any new warnings,
 - add -Werror=format-security.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+dist: bionic
 
 os:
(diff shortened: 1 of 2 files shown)


dwrobel:
This should help to earlier catch warnings like observed in https://github.com/weetmuts/wmbusmeters/issues/119.

Issue #129, opened 2020-06-24
Techem Compact V heat meter


IzeQube:
At my home i have a Techem Compact V heat meter. The funny thing is, that I can get decoded values using vario451, but the values are totally wrong.

I also get the warning: (vario451) unexpected meter version 0x45 !

Is there any way to get my heat meter working?

weetmuts:
If you run with --debug and the meter name vario451 and then email me the log, then I can take a look.

weetmuts:
Ping @IzeQube ?

IzeQube:
Sorry for the late reply. After starting it with --debug I could see that the right values were being received, but are then converted what isn't needed for Compact V meter.

I then did a rather hacky fix with a new meter that reads the right values.

I can give you the details when I'm back home later or I can also post a pull request if you want...

weetmuts:
Cool! Either way works!

IzeQube:
I added a PR, took a while because my dev machine died

weetmuts:
Hi @IzeQube ! As you can see I am ready to merge your pull request, but we need an example
telegram for the regression tests. You can email me a telegram if you like and I will anonymize it and add it to the test suite.

weetmuts:
Closing this for now. Thanks for contributing the code @IzeQube !

PR #139, opened 2020-07-26
Added Techem Compact V heatmeter

IzeQube:
Solves #129

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,7 @@ METER_OBJS:=\
 	$(BUILD)/meter_apator08.o \
 	$(BUILD)/meter_apator162.o \
 	$(BUILD)/meter_cma12w.o \
+	$(BUILD)/meter_compact5.o \
 	$(BUILD)/meter_ebzwmbe.o \
 	$(BUILD)/meter_ehzp.o \
 	$(BUILD)/meter_esyswm.o \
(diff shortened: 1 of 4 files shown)


weetmuts:
Great work! Now we need to have a test case to prevent regressions.

If you take a look in simulations/simulation_t1.txt you will see the test
for the vario meter. Add a telegram for the compact meter after it.
Then add the expected meter in tests/test_t1_meters.sh after vario451.

If you collect the telegram from the debug output, telegram=|...|...|
is not encrypted and you can use NOKEY. If you like, you can adjust
the bcd bytes in the telegram for your meter to something that is not 
your own meter id, ie anonymize it. The id bytes are in bcd and are stored
backwards (but not the nibbles that are forward) in the telegram very early.

You can leave the json {"... and field | .... expected responses at first 
empty in simulation_t1.txt then run make test,  and you will see the 
expected output, that you can paste back into simlation_t1.txt

You also need to add the proper lines in test_listen_to_all.sh
They are also easy to see when you run make test.

//Fredrik

weetmuts:
You can also email me a telegram and I will add the necessary test case.

weetmuts:
Ping @IzeQube

weetmuts:
Ping @IzeQube ? Do you have a telegram for the test suite?

weetmuts:
I merged it but it will remain as unsupported until we get a test case. 
Thanks again @IzeQube for your work!

PR #158, opened 2020-09-01
Add support for AquaMetro / Integra Topas Es Kr water meter

avandorp:
Thank you for creating wmbusmeters. This adds a new water meter manufactured by AquaMetro / Integra.
Please be aware in your review that I am neither a C++ programmer nor an M-Bus expert, just someone with a wireless M-Bus capable water meter in the basement... I've tested the code with an Arduino and a 868 MHz transceiver running the NanoCUL firmware against my water meter. I only read from the meter and did not try sending anything to it (it belongs to the water utility after all). So far all received and decoded values make sense. The reverse volume reading wasn't tested except for the expected 0 cubic meters reading.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,7 @@ METER_OBJS:=\
 	$(BUILD)/meter_rfmamb.o \
 	$(BUILD)/meter_rfmtx1.o \
 	$(BUILD)/meter_supercom587.o \
+	$(BUILD)/meter_topaseskr.o \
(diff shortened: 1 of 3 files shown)


weetmuts:
Many thanks! I only needed to add a regression test case in simulations/simulation_t1_meters.txt
and tests/test_t1_meters.sh

This was easy since you had provided a telegram in the source!

Great work!

Issue #166, opened 2020-09-24
Expand ElectricityMeter to include apparent and reactive power 


Skeen:
Hello,

I'm currently writing the device code for the [EM24 3-phase power meter](https://www.gavazzionline.com/pdf/EM24DINDS.pdf), and it reports reactive power in `kVARh` alongside true power in `kWh`. From these quantities apparent power can be calculated in `kVAh`, however the interface for the `ElectricityMeter` does not currently expose apparent or reactive power.

Thus the question:

How would you feel about adding apparant and reactive power to the `ElectricityMeter` interface?

weetmuts:
Please add it! The interfaces and c++ inheritance are not really used in any way, yet.
But eventually it might make sense to put common code for the ElectricityMeters somehwere,
and then the interfaces might be useful.

I looked at your pull request and it is excellent. We only need a test case,
ie a telegram in simulations_c1.txt and another meter in test_c1_meters.sh
 
You can email me a telegram if you like (from the --debug log) and I can anonymize it and
add it to the test suite.

Skeen:
I have emailed you a telegram.

PR #167, opened 2020-09-24
Add support for the Gavazzi EM24

Skeen:
There is one big problem with this pull request, it spews out warnings:
```
(wmbus) warning: type 123 cannot be scaled!
(wmbus) warning: type 123 cannot be scaled!
{"media":"electricity","meter":"em24","name":"Power","id":"...","errors":"","timestamp":"2020-09-24T18:56:43Z"}
```
I believe they are related to either the fact, that ["DifDifeVifVife quadruplets are not yet implemented"](https://github.com/weetmuts/wmbusmeters/blob/da4c471afcfb95a738dacaf370a2e8a9da29b1d8/HowToAddaNewMeter.txt), althrough I cannot really decide if it only uses triplets, or quadruplets?
```
147     // 04 dif (32 Bit Integer/Binary Instantaneous value)
148     // FB vif (First extension of VIF-codes)
149     // 82 vife (Reserved)
150     // F5 vife (Cold / Warm Temperature Limit 10^-2 Celsius)
151     // 3C vife (Reserved)
```
Or if it is related to the `(Reserved)` codes, somehow throwing things off?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,7 @@ METER_OBJS:=\
 	$(BUILD)/meter_ebzwmbe.o \
 	$(BUILD)/meter_ehzp.o \
 	$(BUILD)/meter_esyswm.o \
+	$(BUILD)/meter_em24.o \
 	$(BUILD)/meter_eurisii.o \
 	$(BUILD)/meter_fhkvdataiii.o \
 	$(BUILD)/meter_hydrus.o \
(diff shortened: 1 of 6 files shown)


Skeen:
Fixes #166

weetmuts:
Yes, the scale warning is because my code not yet handles more complicated vife/combinable vifs.
In this case the reactive energy power is in the first extension interval.
I merged your code and I will now add the necessary vife decoding/scaling.

PR #170, opened 2020-10-01
New detection code

BIBOLV:


slawekzz:
Have you encountered such a problem as in the attached picture. From time to time, the counter catches strange values. This causes a bad consumption count in the Home Assistant.
woda_error2

pswiatki:
Is it possible those offending readings come from different meters somehow?

weetmuts:
What meter do you have?

slawekzz:
> 
> 
> Is it possible those offending readings come from different meters somehow?

No, it's the same meter and the neighbors don't have those meters.
woda_error3

slawekzz:
> 
> 
> What meter do you have?

APATOR AT-WMBUS-16-2

BIBOLV:
Apator uses propitiatory protocol for messages, so there is not much that @weetmuts can do, if you do not know someone from Apator that would shed some light on this issue :wink:.
But I suspect that it should be possible to parse meter readings trough math sensor at HA side and discard those out-of-order readings. If you do not succeed, you can let me know and I will take a look, or you can ask for help in HA forum. They usually are responsive and helpful and if it will be possible, you will find answer there.

slawekzz:
> 
> 
> Apator uses propitiatory protocol for messages, so there is not much that @weetmuts can do, if you do not know someone from Apator that would shed some light on this issue 😉.
> But I suspect that it should be possible to parse meter readings trough math sensor at HA side and discard those out-of-order readings. If you do not succeed, you can let me know and I will take a look, or you can ask for help in HA forum. They usually are responsive and helpful and if it will be possible, you will find answer there.

Unfortunately, I don't know anyone from Apator and we won't get any help there. I will have to do it in HA. I don't know how to do it yet. The case would probably be resolved by creating an IF condition comparing the incoming value with the previous value. If it was greater than eg 1.00 then it would not be registered. I don't know HA enough to do this so I'll be looking for help.

pswiatki:
Since Apator looks to be located in Polska, perhaps someone could contact them and ask for some help? I can try....

weetmuts:
Thanks Bibo! This was merged into NewDetection and is now in master!

PR #186, opened 2020-11-22
Add new Techem water meter MKRadio4

cinemarene:


weetmuts:
Many thanks! I merged and added the missing test code.
Great that you supplied the regression telegram in simulations!
Without the regression tests its not software, its a random bundle of bits. :-)

cinemarene:
I was happy to do that.

Unfortunately I didn't find any real information about the Radio 4 on the internet, where there was speculation that the data was now encrypted. But I found out that this is not the case.
I watched the water meter for a few days and checked if the values arrived correctly.

For the simulation I took an original telegram from my meter.

PR #189, opened 2020-11-29
Use sh instead of bash to support systems without bash

ZeWaren:
One of the test files was still using `/bin/bash` as base interpreter. Change it to `/bin/sh` to support systems without bash.

### DIFF
--- a/tests/test_pipe.sh
+++ b/tests/test_pipe.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 PROG="$1"
 


weetmuts:
Nah, that did not work. I am using a bash for loop in there. Can you rewrite it?

ZeWaren:
> Nah, that did not work. I am using a bash for loop in there. Can you rewrite it?

Oops, my bad. I somehow missed that one. Sorry about that.

Here is the fix: #190

PR #147, opened 2020-08-18
Add support for Multical 403

eborned:
This pull adds support for the Multical403 in both heating and cooling configuration. I'm unable to add the tests for the heating configuration, because my meter is currently still at 0 usage due to the summer.

I'm not sure if the unit configuration is needed, but I've added the units that the meter uses for outputting its values.

PR #202, opened 2020-12-17
Improved multical403 support + M3/H <> L/H conversion

dddevvv:
See discusion PR #147


weetmuts:
Thanks! But can you revert the change to the dockerfile?

dddevvv:
> Thanks! But can you revert the change to the dockerfile?

Sorry, I assumed it would not end up in the PR, but of course it is automatically updated.

weetmuts:
Great!

weetmuts:
Do you have a telegram where these new codes are used? I would like to add an anonymized version of that telegram to the regression tests.

dddevvv:
> Do you have a telegram where these new codes are used? I would like to add an anonymized version of that telegram to the regression tests.

Yes - see discussion in PR #147.
```
telegram=|42442D2C34232571340C8D20F9845B8021F14478|040FFA18000004FF07F173000004FF08D736000004140CA80000043B0000000002598014025D7F0C04FF2200000000|+357
```

PR #204, opened 2020-12-23
Implement Kamstrup OmniPower

janusboandersen:
Implemented Kamstrup OmniPower
src/meter_omnipower.cc
src/meters.h
 
Simulation file added (encrypted data). Works with correct AES key.
simulation/simulation_omnipower.txt
 
Regression tests added (decrypted data for long and compact telegrams). Run OK! 
simulation/simulation_c1.txt
tests/test_c1_meters.sh


weetmuts:
Thank you very much! I had to merge your code manually, since it was based on an older version of wmbusmetrs.
But its merged now! Great work!

weetmuts:
I changed backwards to production, to be consistent with the other electricy meter drivers,
and I added current_ in front of the power readings, also to be consistent with the other drivers.

weetmuts:
Yeah, I know current is unfortunate in regards to electricity.... but all other instantaneous values are prefixed with current_
(like current_alarms, current_temperature etc) to make it obvious that they are instantaneous.... :-)
There can be historical/average data for power, alarms, temp etc.

janusboandersen:
Excellent, thank you for merging it. Makes sense to align to the existing codebase - my bad ;)

PR #209, opened 2021-01-11
Update WaterStarM

makuser:
- Implement manufacturer error codes
- Add another media type 0x07


weetmuts:
Thank you very much! However I had to revert your change of current_status to status. The reason was that the regression tests failed, since they test that the json output is stable for existing telegrams in the test suite.

Now, it might seem odd that I have current_status on almos all of the meters (hydrus seems to be a mistake though). The reason is that I want to make it obvious that status is the real-time current status, and not the status at the month change, or 
other status that the meters can send. One could say that an non-prefixed status is the real-time status, but oh well. Here we are.

So for stability and keeping the regressions tests happy, it is current_status for now. Any change to the json must be handled very very carefully. Perhaps introducing a new meter type?

makuser:
> Now, it might seem odd that I have current_status on almos all of the meters (hydrus seems to be a mistake though). The reason is that I want to make it obvious that status is the real-time current status, and not the status at the month change, or
> other status that the meters can send.

But on all other meters the "current_status" is also the real-time current status, or not? So why is it named differently, I do not really understand the reason behind it?

weetmuts:
I do not quite remember why I picked current_status for the real-time status of the meter when the telegram was sent, when I wrote the first driver for my water meter 4 years ago.

But in retrospect, status might just be better. Its just when you read the wmbus descriptions, there is a lot of talk of instantaneous, so perhaps that was confusing me.

Lets use just status going forward. Anything at past dates would be name _at_xxxx so:
status (is real time when the telegram was set) status_at_set_date or status_at_end_of_last_month (would be the status at that time)
likewise we already have:
total_consumption_m3 vs total_consumption_at_set_date_m3 (usually accompanied by a set_date)

I am pusing a fix since I think you are the only user so far of the waterstar meter. Otherwise
the rule is never ever break any system that is already running. All wmbusmeters users 
should be able to upgrade wmbusmeters without worring. If the json should be changed
then lets create a new driver name instead.

weetmuts:
Ah, you are not the only user of waterstarm....@ArtyomBurda  is also using it. So to be true to the rule, never change any json keys, only add new ones. I reverted my revert of your revert and we are back to current_status.

Next meter will use status instead of current_status. :-)

PR #213, opened 2021-01-20
RTL2838 doesn't seem to exist, change to 2832

rubdos:
Looks like a typo to me, thought I'd fix that.

Additionally, would there be interest in a Wiki about what sticks are supported and what their pros and cons are? Currently, I'm not capable of writing such an article myself, but I could do the research and write something out.


weetmuts:
Thanks! Yes, please write something! We can add it to this wiki: https://github.com/weetmuts/wmbusmeterswiki

PR #215, opened 2021-01-22
Add support for Compact Ve Heat Meter

deveth0:
This PR adds a new mapping for the Techem Compact Ve heat meter. 

The Ve version of the Compact seems to use the same protocol and runs in t1 mode on a cul for me. 

I'll check with our neighbors tomorrow and verify another set of data.

This is the log for a single telegram, I'm not 100% sure, why it shows "Unknown" even though I added the Type / Version to `meters.h`?
```
97647579:(meter) Heizung compact5 handling telegram from 43330826
97647636:(telegram) DLL L=36 C=44 (from meter SND_NR) M=5068 (TCH) A=43330826 VER=22 TYPE=43 (Unknown) (driver compact5) 
DEV=cul RSSI=0
97647763-(telegram) TPL CI=a1
97647784-telegram=|36446850260833432243A1|009F2962060088560000809B410F30C0E043061858E0C001000020C00101142081041258D081062980E0C106|+4279
```

The decoded version of the telegram is:
```
{"media":"Unknown","meter":"compact5","name":"Heizung","id":"43330826","total_kwh":1.72,"current_kwh":0.086,"previous_kwh":1.634,"timestamp":"2021-01-23T10:39:39Z","device":"cul","rssi_dbm":0}
```

This matches the values I see on the device's display EXCEPT that those are diviced by 1k (correct values according to the device: 1722 kwh (previous: 1634 kwh). 

I'm not sure, if this is a bug in the compact5 meter or if the compact5 and Ve use a different scaling, what are your thoughts?

I verified the data with two neighbors, both devices show the correct values, so I'm confident, that the basic decoding is working similar to the compact5.


weetmuts:
Thanks! I merged the detection. Now we "just" need to reverse engineer Techems proprietary protocol to figure out where the scale factor is stored.

PR #221, opened 2021-01-29
Implemented energy meters GSS CC101 and GSS CC301

xaelsouth:


weetmuts:
Excellent work! I just replaced  spaces in the json keys with underscores. The reason is that I want the output to be xml compatible as wekk, and xml tags cannot contain spaces. I also made sure the new units volt and ampere translate into the suffixes _v and _a , so that they are consistent with for example _kwh etc.

PR #227, opened 2021-02-03
Fix iIllegal quoting in CSV file

don-vip:
GitHub presents this warning:
image

Apparently the illegal quoting prevents GitHub to display it properly, it should solve this problem.

### DIFF
--- a/DLMS_Flagids.csv
+++ b/DLMS_Flagids.csv
@@ -850,7 +850,7 @@ REP	Radiant Energy Solutions Pvt Ltd	India	Asia
 RES	Renesas Electronics Singapore Pte Ltd	Singapore	Asia
 RIC	Richa Equipments Pvt. Ltd.	India	Asia
 RIL	Rikken Instrumentation Limited	India	Asia
-RIM	"CJSC \""Radio and Microelectronics\"""	Russian Federation	Europe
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks, I did that change and several other quoting "problems", but github still complained. So I removed the quotes altogether,
then it says: "We can make this file beautiful and searchable if this error is corrected: It looks like row 11 should actually have 1 column, instead of 2. in line 10."

And however I look at row 11 or line 10, I cannot see any problem... besides every row/line is 4 columns, so the github logic seems a bit broken.

don-vip:
Maybe GitHub doesn't like missing columns, such as AEG which does not have any country/region?

weetmuts:
Yeah, its something. But since I download this file from the DLMS website, its not really mine to edit, just to make
github happy. :-) Thanks anyway! But I think we can leave github to its own problems here.

So I download the xlsx into LibreOffice, then save it as a csv file. The bad quoting is probably created by LibreOffice
in combination with weirdo quotes in the original source. This is why "make update_manufacturers"
performs a lot of cleanup before putting the strings into the source code.

If you want, you could create a conversion script that takes the xlslx file and converts it into a csv file that github likes!

don-vip:
OK no problem I thought it was an easy fix. No need to lose more time on such a trivial issue :)

PR #230, opened 2021-02-04
Fix IZAR previous_alarms description

don-vip:
izar current_alarms  and previous_alarms  have the same description:

```
$ wmbusmeters --listfields=izar
...
          current_alarms  Alarms currently reported by the meter.
         previous_alarms  Alarms currently reported by the meter.
```


weetmuts:
Thanks!

PR #232, opened 2021-02-05
Support RSSI on CUL

CuBiC3D:
I added RSSI readings support for CUL devices. Using mode `X21` on CUL (5. bit to `1`) will enable raw LQI and RSSI values appended to the message. This PR will extract both, output LQI only in debug mode, calculate RSSI value in dBm according to datasheet of TI CC1101 and populate the `AboutTelegram` field.

I tested using a custom made CUL with an modified alternative firmware - so would be very happy if some one could verify using a more of the shelf hard- and firmware. But should be fine since it's using the same lib from the original implementation [here](https://sourceforge.net/p/culfw/code/HEAD/tree/trunk/culfw/clib/rf_mbus.c#l393) where LQI and RSSI are appended.

### DIFF
--- a/src/wmbus_cul.cc
+++ b/src/wmbus_cul.cc
@@ -76,7 +76,8 @@ private:
 
     FrameStatus checkCULFrame(vector<uchar> &data,
                               size_t *hex_frame_length,
-                              vector<uchar> &payload);
+                              vector<uchar> &payload,
+                              int *rssi_dbm);
 
     string setup_;
 };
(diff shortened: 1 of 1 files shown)


weetmuts:
Very nice! Great work!

weetmuts:
I have tested with my nanoCUL and it seems work fine. I get an rssi of -61dBm when listening to my multical21 in C1 mode, which is ~10 meters away.

PR #234, opened 2021-02-05
rtl_wmbus S1 mode enable

alalons:
Enable S1 mode


weetmuts:
This is intersting! Have you tried this and actually seend telegrams from both C1/S1 and T1 meters at the same time?

alalons:
Tried with s1 and c1 telegrams at the same time. Received both.
Auto:c1 receives s1 too
Auto:s1 receives c1 too
Auto:c1,s1 receives both

weetmuts:
I have added support for the latest rtlwmbus and now wmbusmeters defaults to the s1,t1+c1 setting.
Closing this one for now.

PR #237, opened 2021-02-06
Fix parsing for Techem HCA with version 0x94

### LINKED ISSUE #233: Techem Heat Cost Allocator Data III using wrong fields
I have two of those Techem Data III devices on my own radiators and almost a hundred in radio range. They all have version `0x94`. I know there are probably some versions but the only I can find reversed information about are `0x69` and `0x94`. Reverse engineered information about the field is available in the source of FHEM: https://github.com/mhop/fhem-mirror/blob/master/fhem/FHEM/32_TechemHKV.pm

My problem is this: The current implementation in this tool uses offsets from version `0x69`. On version `0x94` only the temperature offsets change, I could make a PR and add logic to distinguish the version of the device using `t->dll_version` and use different offsets on the temperature values, but there is probably something else wrong with the current implementation:

The used offsets are probably taken from the source mentioned above and been divided by 2 to address bytes instead of characters, but the payload in this tool does not contain the header, in FHEM it does - so every offset seems to be +10 bytes off and I can verify at least using version `0x94` that values are not correct. Subtracting the header fixes this.

The only simulated Data III (`0x69`) I can find is here: https://github.com/weetmuts/wmbusmeters/blob/master/simulations/simulation_t1.txt#L52-L56 and already mentions problems:
```
# There is a problem in the decoding here, the data stored inside the telegram does not seem to properly encode/decode the year...
```

I can't verify the values and not sure if adding version `0x94` and fixing offsets would break `0x69`. Can anyone verify that the current implementation actually works with those devices? Related to #206 

@cinemarene I see it's probably your PR? Do you still use that?

CuBiC3D:
This fixes #233 by correcting the offset of the temperature values if version `0x94` is sending the data. This covers Techem FHKV radio 4 and EHKV vario 4 devices. Further it removes the wireless mbus header from `content` if `A0` field is used - this has been introduced by https://github.com/weetmuts/wmbusmeters/commit/34a52970933960490fe7e4951b84b4001be53d33 for the version `0x69` version, which did not interact with the header on parsing.


weetmuts:
Thanks! Great work!

PR #240, opened 2021-02-07
Add known date fields to mkradio3 and tsd2

CuBiC3D:
This adds known previous and current date fields to `mkradio3` and `tsd2` drivers. The smoke detector needs some more reversing, but it seems that some counters will reset on on those previous period dates. I can't get the current date on the `tsd2` right, so I left it out for now - the assumed day bits show weird behaviour.


weetmuts:
Nice! Thanks!

PR #241, opened 2021-02-08
Add frame preprocessing to transform Diehl addresses

don-vip:
First part required to add support to another Diehl meters.

See https://github.com/weetmuts/wmbusmeters/issues/239#issuecomment-774685536


weetmuts:
Great work! Thank you very much!

PR #243, opened 2021-02-08
 Decode IZAR transmit period 

don-vip:
Add decoding support for transmit period on IZAR devices. Requires #241 to be merged.


weetmuts:
Thanks!

Issue #239, opened 2021-02-07
Add support for Diehl/Hydrometer SHARKY 775


weetmuts:

weetmuts:
@don-vip Thank you very much for contributing! Please contact me at oehrstroem@gmail.com if you have any questions!

don-vip:
Hi @weetmuts ! Sorry for having shared something that I shouldn't have.

One explanation concerning the IZAR device types "oddities": In fact in several (and numerous) conditions, the telegrams deviate from the OMS spec concerning the A field. Instead of having Serial number on first 4 bytes, then version, then device type, many IZAR meters will transmit first version, then device type, then serial number. That's why you find so many "exotic" device types. In fact they use the proper device type but a an unexpected position. Example for my heat meter

`telegram=|534424232004256092687A37004575|2235854DEEEA5939FAD81C25FEEF5A23C38FB9168493C563F08DB10BAF87F660FBA91296BA2397E8F4220B86D3A192FB51E0BFCF24DCE72118E0C75A9E89F43BDFE370824B|`

"A field" is transmitted as `200425609268` which implies a device type of 0x68 which does not exist, a serial number that doesn't match the one printed on the device, and a version which differs between several identical meters.

I should been read as `256092682004` which implies an heat meter (0x04: correct), a version of 0x20 identical across identical meters, and a serial number that matches the ones printed on the meter.

I will try to create a first PR to handle this mode I'll call "swapped mode". Conditions to interpret how to read the A field are based on C field, M field, CI field and a flag in CFG field. Pseudo algorithm for IZAR devices:

```
if C field == 0x68 or 0x70: // SND-NR or ?
  switch CI field:
    case 113: // 0x71
      // swapped mode
    case 122: // 0x7A
      if flag set in CFG field:
        // swapped mode
      else:
        // normal mode
    case 160, 161, 162, 163, 164, 165, 166, 167: // 0xA0-0xA7
      if prios (M field in "HYD", "EWT", "DME", "SPL", "SAP"):
        // swapped mode for "HYD", "EWT", "DME", "SPL". Not sure yet for "SAP" code, custom SAP logic involved
      else:
        // normal mode
    case 176: // 0xB0
      if M field == "SAP":
        // don't know yet, custom SAP logic involved
      else:
        // normal mode
    case 177, 178, 179: // 0xB1-0xB3
        // swapped mode
[...]

PR #247, opened 2021-02-13
Add Sharky 775 + fix Diehl address swapping

don-vip:
- Fix a bug for Diehl address swapping I introduced in #241 (off-by-1 offset error, I swapped bytes by 1 instead of 2)
- Add support for Diehl/Hydrometer SHARKY 775 #239
- Required to move and update LFSR decryption code from Izar driver to common manufacturer specific files, in order to be called from core when decrypting  TPLSecurityMode::SPECIFIC_16_31 for Diehl Metering meters transmitting in "Real Data" mode
- Added TemperatureDifference type

I can finally listen to my heat meter, it was so difficult (and it would have been so simple if they just followed the standard...)


weetmuts:
Thank you very much! It is so fascinating to see someone just jump in and write great code that does everything just right, design, style and testing!

weetmuts:
>  it would have been so simple if they just followed the standard...

Yes, indeed! The wmbus standard is brilliant, even there is almost always is some sort of translation necessary to figure out what the measuremements mean,  you get framing, decryption, measurements and units for free. A lot of meta-data is encoded into those VIF(e)s. 

This prios/izar protocol seems to have come from Sappel originally (Diehl seems to have bought them at some time), maybe they were creating these radio meters before the standard was set in stone? What we see here are their attempts, that perhaps contributed to the standard eventually? Assuming someone has invested in a lot of collector devices that only understands the old protocol, they have a reason to continue sell meters using the old protocol.

I have even seen meters HCAs that seem to transmit two different telegram types regularly. One proprietary and one oms-compliant. Also probably to cater for old collector systems and new systems in parallel.

don-vip:
> This prios/izar protocol seems to have come from Sappel originally (Diehl seems to have bought them at some time), maybe they were creating these radio meters before the standard was set in stone? What we see here are their attempts, that perhaps contributed to the standard eventually?

Indeed :) https://translate.google.com/translate?sl=auto&tl=en&u=https://fr.wikipedia.org/wiki/Sappel_%28entreprise%29

Issue #223, opened 2021-01-31
Another crazy Hydrus telegram


weetmuts:

weetmuts:
Can you first run with just wmbusmeters auto:t1
to capture a bunch of:

Received telegram from: 81100028
          manufacturer: (HYD) Hydrometer (0x2324)
                  type: Bus/System component (0x0e)
                   ver: 0x64
      Concerning meter: 64745666
          manufacturer: (DME) DIEHL Metering, Germany (0x11a5)
                  type: Water meter (0x07)
                   ver: 0x70
                driver: hydrus

To see if the type/ver change...

Then collect a bunch of --debug telegrams from your 64745666 meter
and grab the debug decoding lines using grep "(hydrus) ..:"
see if you can figure out any pattern and email me/post here the grepped lines as well.

zoesn89:
Hey @weetmuts,

did you get my logfiles? If yes, did they help? ;)

Felix

don-vip:
@weetmuts @zoesn89 I'm on it. I'm playing with this telegram:
`telegram=||6644242328001081640E7266567464A51170071F0050052C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C|+5`
I'm observing odd behaviour right now, I'll submit a PR once I get things sorted out.

zoesn89:
> @weetmuts @zoesn89 I'm on it. I'm playing with this telegram:
> 
> `telegram=||6644242328001081640E7266567464A51170071F0050052C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C|+5`
> 
> I'm observing odd behaviour right now, I'll submit a PR once I get things sorted out.

Nice, thank you! 😉

don-vip:
https://github.com/weetmuts/wmbusmeters/pull/248 should allow you to monitor your Hydrus meter.

weetmuts:
@zoesn89 Can you confirm that it works for you now?

zoesn89:
Unfortunately not -.-
I did the following: 

- make clean (thanks for that hint)
[...]

Issue #216, opened 2021-01-23
Support for IZAR RADIO EXTERN 868 MBUS + Hydrus watermeter


zoesn89:
Hey guys,

I have a DIEHL Hydrus watermeter. It is a Mbus meter attached to an wMbus sender named IZAR RADIO EXTERN 868 MBUS. Unfortunately, this combination isn't supported yet.

Here is the output from the cli:

```
wmbusmeters /dev/ttyUSB0:im871a:t1               
Started config im871a[00101717] on /dev/ttyUSB0 listening on t1
No meters configured. Printing id:s of all telegrams heard!
Received telegram from: 81100028
          manufacturer: (HYD) Hydrometer (0x2324)
                  type: Bus/System component (0x0e)
                   ver: 0x64
                device: im871a[00101717]
                  rssi: -66 dBm
                driver: unknown!
```
When I try the izar and hydrus driver, I get this:
```
wmbusmeters /dev/ttyUSB0:im871a:t1 Hydrus izar 81100028 NOKEY
Started config im871a[00101717] on /dev/ttyUSB0 listening on t1
(meter) Hydrus: meter detection did not match the selected driver izar! correct driver is: unknown!
(meter) Not printing this warning agin for id: 81100028 mfct: (HYD) Hydrometer (0x2324) type: Bus/System component (0x0e) ver: 0x64
(meter) please consider opening an issue at https://github.com/weetmuts/wmbusmeters/
(meter) to add support for this unknown mfct,media,version combination
```
Here are some telegrams:
```
telegram=||6644242328001081640E7266567464A51170071F0050052C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C|+5

telegram=||6644242328001081640E7266567464A51170072000500511BB62025DDC52526495E29082A2A00D03386C80921C1F0594465FA7CE186BFEB2B0F3F91B00FA47A4017C909765720C3860CE887AF417F4B651B7E6D2BA7C96C4D6222876EADA756B9CAEAEE0D63D17|+23

[...]

PR #248, opened 2021-02-13
Improve Hydrus support

don-vip:
This PR includes the following:
- fix a nasty bug in toValueInformation function
- add OperatingTime and ActualityDuration
- add Minute time unit
- add support for default manufacturers keys
- more reliable parsing of Hydrus meter fields

_Previously (solved in comments):_

I hit something I don't understand and need your help:

```
$ ./build/wmbusmeters --debug simulations/simulation_hydrus.txt hydrus hydrus '*' NOKEY 2>&1 | grep decrypt
(TPL) decrypting "2C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C"
(TPL) decrypted  "2F2F15257A616F14139172133795300D000000008C2013917213000B3B0000000B26784601025AF5000266EF00046D1B08B7214C1338861200CC101300000000CC201338861200426C9F2C42EC7EBF2C"

$ ./build/wmbusmeters --debug simulations/simulation_hydrus.txt hydrus hydrus '*' NOKEY 2>&1 | grep decrypt
(TPL) decrypting "2C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C"
(TPL) decrypted  "2F2F15257A616F141391721337F5410D000000008C2013917213000B3B0000000B26784601025AF5000266EF00046D1B08B7214C1338861200CC101300000000CC201338861200426C9F2C42EC7EBF2C"

$ ./build/wmbusmeters --debug simulations/simulation_hydrus.txt hydrus hydrus '*' NOKEY 2>&1 | grep decrypt
(TPL) decrypting "2C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C"
(TPL) decrypted  "2F2F15257A616F14139172133715A60C000000008C2013917213000B3B0000000B26784601025AF5000266EF00046D1B08B7214C1338861200CC101300000000CC201338861200426C9F2C42EC7EBF2C"
```

As you can see consecutive AES decryptions of the same input produce different output, messing up later with frame decoding. Yet the 2F2F checksum is always found and thus the decryption appears to be OK. How is that possible?


weetmuts:
If you have run wmbusmeters with a meter key, then the telegrams printed using --logtelegrams 
(telegram=|...| are actually not the exact original telegram, they are indeed the decrypted telegrams.
I choose this solution because then users of wmbusmeters can simply share their log files 
without worrying about the key. We can jump directly to the content decoding.

In those few occasions when there is a problem in the decryption code, then they will
need to share their key and the parts of the log file which is the lower level serial data.
If you are running wmbusmeters without a meter key, then logtelegrams will log the complete
original telegram, encryption and all. This is what happens if you run for example:
"wmbusmeters --logtelegrams auto:c1"

When running from simulation_xxx.txt files, wmbusmeters will skip decryption if no key was supplied.
The 0x2f 0x2f check will still succeed, since the data was already decrypted.

weetmuts:
Ha! Thanks for catching that bug in ValueInformation! It just randomly worked for the use cases so far! :-)

don-vip:
Sorry, I should have explained more. The telegram is the first raw telegram posted in #216 / #223:
`telegram=||6644242328001081640E7266567464A51170071F0050052C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C|+5`

The telegram is ciphered using default manufacturer key. I added support to this default key, without adding anything new or sensitive ;) This way it is possible to decrypt the telegram using `NOKEY` exactly like when we decipher LFSR data using PRIOS default keys.

The problem is that the AES decryption produces different output for the same input. I don't understand why/how.
The effect of that is that it completely confuses the DV parser later, with probably invalid and random data, it's very weird.

I don't know if I did something really wrong here or if I'm hitting a rare bug in the AES decryption algorithm not yet faced.

weetmuts:
Hmm, but I see what you mean. Can you share your simulations/simulation_hydrus.txt file?

don-vip:
> Hmm, but I see what you mean. Can you share your simulations/simulation_hydrus.txt file?

It's in the PR, first file :)
```
# Test Hydrus with default AES encryption
telegram=||6644242328001081640E7266567464A51170071F0050052C411A08674048DD6BA82A0DF79FFD401309179A893A1BE3CE8EDC50C2A45CD7AFEC3B4CE765820BE8056C124A17416C3722985FFFF7FCEB7094901AB3A16294B511B9A740C9F9911352B42A72FB3B0C|
{"media":"water"}
|HydrusAES;
```

weetmuts:
We posted exactly at the same time. I will try your example myself.... decryption cannot partially succeed for AES CBC blocks.
So if some minor bytes differ, then perhaps there is a buffer overwrite somewhere.

don-vip:
OK I'm not crazy then :) Thanks!

weetmuts:
Found the problem! Its a broken initialization of the IV, and yes when the IV is wrong we can get partially broken bytes!
The new IV code was added for this particular Hydrus meter + Izar wmbus dongle.

weetmuts:
Fix pushed to master. Please merge! The randomness was due to non-initialized memory in the IV.
It should now always be completely initialized. Perhaps it is even correctly initialized. :-)
Please test!

If you need to adjust anything its in decrypt_TPL_AES_CBC_IV line 129-133
and parseLongTPL where the array tpl_a is filled in.

When only the DLL exists, then IV is mfct + id + type +version, 
which happens to be exactly the byte order in the telegram.

When the TPL exists, I am assuming the IV is still mfct + id + type + version.
However in the TPL the actual byte order is id + mfct + type + version,
but this is handled in parseLongTPL.

We have to check the spec for what it should be, if my guess is not right.

don-vip:
> Please test!

Fix confirmed! Decryption is stable now :) I can resume my work on supporting the new payload :)

don-vip:
The PR is almost ready. Only one failure with First Hydrus telegram in tests which has been added deciphered. wmbusmeters tries to decipher it again now that I added a default key, and it fails:
```
(TPL) decrypting "2F2F01FD08300C13741100007C1300000000FC101300000000FC201300000000726C00000B3B00000002FD748713025A6800C4016D3B177F2ACC011300020000"
(TPL) num encrypted blocks 4 (64 bytes and remaining unencrypted 0 bytes)
(TPL) IV A511646464647007EDEDEDEDEDEDEDED
(TPL) decrypted  "8EC0C6FEFDF3115B633ACDFE24DFB9AD9E1D78583E15A332171A0240D7928B8E6EF520E4EEC30A63E757FACB71B8E5D38FB1D1EA4923120C06914E9E4FF86702"
(wmbus) decrypted content failed check, did you use the correct decryption key? Permanently ignoring telegrams from id: 64646464 mfct: (DME) DIEHL Metering, Germany (0x11a5) type: Water meter (0x07) ver: 0x70
```

What is the recommended course of action? I see two possibilities:

1. Include the ciphered telegram in tests, not the deciphered one
2. Keep it deciphered but fix the CFG field to make it clear the payload is not AES encrypted

weetmuts:
Maybe something like this?

Add:
if (alreadyDecryptedCBC(pos)) return true; 
just before:
 if (!meter_keys) return false;

For both AES_CBC_IV and AES_CBC_NO_IV

bool Telegram::alreadyDecryptedCBC(vector<uchar>::iterator &pos)
{
    if (*pos != 0x2f || *pos != 0x2f) return false;
    addExplanationAndIncrementPos(pos, 2, "%02x%02x decrypt check bytes", *(pos+0), *(pos+1));
    return true;
}

Anonymize the id in the telegram when you add it decrypted to the simulation database. For this meter
there are two ids, the dll and the tpl. I will check with the user who supplied the telegram if its ok
to add an encrypted telegram to the database as well, since that one cannot be anonymized.

weetmuts:
Its ok, you can add the encrypted telegram to the test suite. Which is good because then we test all parts of the process.

don-vip:
Now all tests are OK. But I had to modify one value in `simulation_dll_tpl_diff.txt` (total_m3 is now expected to be set to 0 instead of 114.35283) and I'm not sure to understand why. Are you sure the contents are currently correct? If I decode this telegram using official IZAR library this is the output I get:
```
ReadingData={
	rawMessage={
		frame=[66,44,24,23,81,81,81,81,64,0E,72,46,56,46,56,A5,11,70,07,1F,00,00,05,2F,2F,15,25,7A,61,6F,14,13,91,72,13,7D,AE,3A,0C,00,00,00,00,8C,20,13,91,72,13,00,0B,3B,00,00,00,0B,26,78,46,01,02,5A,F5,00,02,66,EF,00,04,6D,1B,08,B7,21,4C,13,38,86,12,00,CC,10,13,00,00,00,00,CC,20,13,38,86,12,00,42,6C,9F,2C,42,EC,7E,BF,2C], timestamp=1612657129739.000, moduleId=0
	},
	frameDescription={ type=FrameDescMBusRadio,
		meterId=UDME700756465646000(parent=UHYD640E81818181000),
		meterName=HYDRUS_R4_DME,
		telegramType=OPEN_METERING_LONG,
		withAlarm=false,
		withError=false,
		meterDeviceNumber=56465646,
		mediumByte=0x07,
		version=112,
		manufacturer=HYD,
		errorFlagBytes=0x00,
		telegramContentType=STANDARD,
		installFrame=false,
		synchronizeFrame=false,
		lField=0x66,
		cField=0x44,
		mField=[A5,11],
		aFieldSecondaryAddress=[46,56,46,56,70,07],
		ciField=0x72,
		encMode=AES_CBC_IV },
	telegramFormatError=false,
	parsingLevel=FULL_CLEARTEXT,
	measurements={
		UDME700756465646000 -> {
 			0 -> {
Date/Time= 2021.1.23 08:27:00,
				values=[
					{ semantic=operating.time.MAX,	type=QUANTITY,	isValid=true,	value=1.2085640268593492E-26min },
					{ semantic=volume.MAX,	type=QUANTITY,	isValid=true,	value=-0.005341933㎥ },
					{ semantic=energy,	type=QUANTITY,	isValid=false,	value=0J },
					{ semantic=energy,	type=QUANTITY,	isValid=true,	value=0.000W·h },
					{ semantic=volume.tariff[2],	type=QUANTITY,	isValid=true,	value=137.291㎥ },
					{ semantic=volume.flow,	type=QUANTITY,	isValid=true,	value=0.000m³/h },
					{ semantic=operating.time,	type=QUANTITY,	isValid=true,	value=14678h },
					{ semantic=flow.temperature,	type=QUANTITY,	isValid=true,	value=24.5°C },
					{ semantic=external.temperature,	type=QUANTITY,	isValid=true,	value=23.9°C },
					{ semantic=date.time,	type=DATE,	isValid=true,	value=2021.1.23 8:27:0 }
				]
			},
			1 -> {
Date= 2020.12.31,
				values=[
					{ semantic=volume,	type=QUANTITY,	isValid=true,	value=128.638㎥ },
					{ semantic=volume.tariff[1],	type=QUANTITY,	isValid=true,	value=0.000㎥ },
					{ semantic=volume.tariff[2],	type=QUANTITY,	isValid=true,	value=128.638㎥ },
					{ semantic=date,	type=DATE,	isValid=true,	value=2020.12.31 },
					{ semantic=date.FUTURE_VALUE,	type=DATE,	isValid=true,	value=2021.12.31 }
				]
			}
		}
	}
}
```
... which has weird values in it for operating.time.MAX and volume.MAX.

Is it possible that the AES decrypted payload was invalid when you added this test case? It might be good to decrypt the raw telegram again now that you have fixed the IV bug.

weetmuts:
> Is it possible that the AES decrypted payload was invalid when you added this test case? I

Yes, most likely. You can correct the test case, I will double check later, and potentially recreate the test case.

don-vip:
OK thank you! The PR is ready for review and merge, then :)

weetmuts:
Great work!

zoesn89:
Could it be possible, that the PR courses problems with the command make? I did the following: 

1. git pull
2. ./configure
3. make

and then I got the following error:

`chip@chip:~/git/wmbusmeters$ make
Building 1.0.5-62-g05d52e0
g++ -o build/testinternals build/aes.o build/aescmac.o build/cmdline.o build/config.o build/dvparser.o build/mbus_rawtty.o build/meters.o build/meter_amiplus.o build/meter_apator08.o build/meter_apator162.o build/meter_cma12w.o build/meter_compact5.o build/meter_ebzwmbe.o build/meter_ehzp.o build/meter_esyswm.o build/meter_elf.o build/meter_em24.o build/meter_emerlin868.o build/meter_ev200.o build/meter_evo868.o build/meter_eurisii.o build/meter_fhkvdataiii.o build/meter_fhkvdataiv.o build/meter_flowiq2200.o build/meter_hydrus.o build/meter_hydrocalm3.o build/meter_hydrodigit.o build/meter_iperl.o build/meter_izar.o build/meter_izar3.o build/meter_lansendw.o build/meter_lansensm.o build/meter_lansenth.o build/meter_lansenpu.o build/meter_mkradio3.o build/meter_mkradio4.o build/meter_multical21.o build/meter_multical302.o build/meter_multical403.o build/meter_multical603.o build/meter_multical803.o build/meter_omnipower.o build/meter_q400.o build/meter_qcaloric.o build/meter_rfmamb.o build/meter_rfmtx1.o build/meter_sharky.o build/meter_supercom587.o build/meter_sontex868.o build/meter_topaseskr.o build/meter_tsd2.o build/meter_ultrimis.o build/meter_vario451.o build/meter_waterstarm.o build/meter_whe46x.o build/meter_whe5x.o build/meter_sensostar.o build/meter_gransystems_ccx01.o build/manufacturer_specificities.o build/printer.o build/rtlsdr.o build/serial.o build/shell.o build/sha256.o build/threads.o build/util.o build/units.o build/wmbus.o build/wmbus_amb8465.o build/wmbus_im871a.o build/wmbus_cul.o build/wmbus_rtlwmbus.o build/wmbus_rtl433.o build/wmbus_simulator.o build/wmbus_rawtty.o build/wmbus_rc1180.o build/wmbus_utils.o build/testinternals.o  -lrtlsdr -lusb-1.0 -lpthread
/usr/bin/ld: build/testinternals.o: in function `test_kdf()':
testinternals.cc:(.text+0x11a2): undefined reference to `bin2hex[abi:cxx11](std::vector<unsigned char, std::allocator<unsigned char> >&)'
/usr/bin/ld: testinternals.cc:(.text+0x11f6): undefined reference to `bin2hex[abi:cxx11](std::vector<unsigned char, std::allocator<unsigned char> >&)'
collect2: error: ld returned 1 exit status
make: *** [Makefile:248: build/testinternals] Error 1`

weetmuts:
Can you do "make clean" and "make" again?

zoesn89:
@weetmuts ah, ok. Thank you ;)

PR #250, opened 2021-02-14
Update build_ubuntu.yml

DanielHabenicht:
add dependencies that might not be needed on the current build infrastructure but are needed locally

PS: I found out it was needed here: https://github.com/weetmuts/wmbusmeters/issues/116#issuecomment-634901813 but I think most people building it will look in the according build pipeline (as long as there is not build guide)


weetmuts:
Thanks!

PR #251, opened 2021-02-15
Fix IZAR manufacture year parsing

don-vip:
I just spotted an IZAR meter built in 2031...
Obviously my parsing was wrong, I didn't keep the correct bits for the first byte (which I didn't see because it worked for my own meter).

Added a non-regression test to make sure this meter has been built in 2015.


weetmuts:
Thanks!

PR #294, opened 2021-06-03
Add Qundis 5.5 watermeter version 16 to lse_07_17 driver

westfalenyeti:
While trying to readout my Qundis 5.5 watermeter, I noticed that it was identifying itself as version number 16, not 17 as expected bei the driver lse_07_17. With this change, wmbusmeters finds the correct driver by itself and interprets the telegrams correctly.

This pull request just adds the different version to meter_detection.h as outlined in other discussions.


weetmuts:
Great! Do you have a telegram that I can add to the test-suite? You grab that with --debug or --logtelegrams.
You can email it to me, then I can anonymize it and put it into the test-suite.

Issue #320, opened 2021-08-09
apator162 - would there be a more flexible way to point out the offset?


to-masz:
I have bought the brand new apator AT-WMBUS-16-2 radio module on my own and after its activation I am able to receive the telegrams using nanocul and wmbusmeters 👌 

However, I have yet another pattern in the telegrams. Maybe it is because it is brand new and was not configured by any company 🤔 

```
0F 3307D8 9509020040 B2D2B5F0F4 F341555942 5850 0C00 F00101000010 E700 0000 FFFFFFFFFFFFFFFFFFFFFFF877
... a while after
0F 4167D8 9529020040 B4C0C4F0F3 F341555942 EB76 0C00 F00101000010 0E13 0000 FFFFFFFFFFFFFFFFFFFFFF49F9
0F 4467D8 9529020040 B4C0C2F0F3 F341555942 1377 0C00 F00101000010 0E13 0000 FFFFFFFFFFFFFFFFFFFFFF30C8
0F 4567D8 9529020040 B6C0BCF0F3 F341555942 2177 0C00 F00101000010 0E13 0000 FFFFFFFFFFFFFFFFFFFFFFBEEB
... few days after
0F 3068F8 9529020040 C7C1B3F0F3 F341555942 53C2 0C00 F00101000010 5714 0000 FFFFFFFFFFFFFFFFFFFFFFE883
0F 3168F8 9529020040 C7C1B3F0F3 F341555942 61C2 0C00 F00101000010 5714 0000 FFFFFFFFFFFFFFFFFFFFFF88C2
0F 3368F8 9529020040 C6C1B3F0F3 F341555942 7BC2 0C00 F00101000010 5714 0000 FFFFFFFFFFFFFFFFFFFFFFAC2B
// running some water
0F 3568F8 9529020040 C6C1B2F0F3 F341555942 96C2 0C00 F00101000010 5C14 0000 FFFFFFFFFFFFFFFFFFFFFF699F
0F 3668F8 9529020040 B2D3B3F0F4 F341555942 A6C2 0C00 F00101000010 5C14 0000 FFFFFFFFFFFFFFFFFFFFFF472D
0F 4468F8 9529020040 B2D3B3F0F4 F341555942 2BC3 0C00 F00101000010 5C14 0000 FFFFFFFFFFFFFFFFFFFFFF3657
```

As you can see the offset is 32 (2f2f + 0f + 29 other bytes). 

[...]

PR #328, opened 2021-08-21
Add possibility to explicitly point meter reading offset in apator162 telegram

to-masz:
Following the discussion in #320 it adds the possibility to use a custom offset if wmbusmeters does not return proper value for your apator162. As apator162 uses a closed/proprietary protocol and we have no idea how to confidently determine location of the water meter reading in the telegram, it opens a possibility for the users to use the correct offset without a need to change the source code.

Usage: provide extra params to the driver using an existing approach `driver(extra_option=value)` that in our case will be exactly `apator162(offset=123)` where `123` is a desired offset to be used with a given apator162 radio module.


weetmuts:
Great work!

weetmuts:
FYI @to-masz, I usually do not use exception throwing in my c++ code because of compilers....
I noticed your exception handling and thought, maybe it works this time. Lets try!

Your code compiled perfectly on my Ubunty and Macosx, but alas g++ on alpine linux used for
for docker images, the build failed with some obscure:
error: expected unqualified-id before '&' token
79 |             catch (const std::invalid_argument &e) 

I tried some workarounds without succeeding, so I switched to atoi instead. ;-)

PR #385, opened 2021-11-11
Honor the env in child processes

### LINKED ISSUE #380: Warning! Auto scan has found an rtl_sdr dongle, but you have no rtl_wmbus in the path!
Warning! Auto scan has found an rtl_sdr dongle, but you have no rtl_wmbus in the path!

vvvrrooomm:
fixes #380: execvpe replaces the original env with the given array. The
array was empty effectivly removing PATH from the subshell. Thus
rtl_wmbus is not found, even if located within the PATH, as stated in
the documentation.

### DIFF
--- a/src/shell.cc
+++ b/src/shell.cc
@@ -59,7 +59,7 @@ void invokeShell(string program, vector<string> args, vector<string> envs)
 #if (defined(__APPLE__) && defined(__MACH__)) || defined(__FreeBSD__)
         execve(program.c_str(), (char*const*)&argv[0], (char*const*)&env[0]);
 #else
-        execvpe(program.c_str(), (char*const*)&argv[0], (char*const*)&env[0]);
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks! It usually worked since it looks for rtl_wmbus in /usr/bin or in the same directory as the wmbusmeters binary.

weetmuts:
Sorry, I was too quick here. Your change broke wmbusmeters because the env should contain all the meter information in env variables. Like METER_JSON etc. 

I have reverted your commit. Can you please open an issue and explain what the problem is that you experience.
The real fix might be that we copy the PATH from the invoking shell into the env.

However, my experience is that when you run wmbusmeters from the command line, then it will find the rtl_wmbus in your PATH. When you run as a daemon, you must place rtl_wmbus in /usr/bin or in the same dir as the wmbusmetersd binary. This is a security feature.

So please  open an issue and explain in more detail the problem.

PR #400, opened 2021-11-30
Add new fields for kamstrup multical302 in mega joule

DenysFrasinich:
Only after I add this fields it get non zero energy consumption from meter.


weetmuts:
Thanks for your work, but take a look at src/meter_hydrocalm3.cc that meter can also be configured to either send the energy as kwh or mj.

The addPrint commands prepares printing without having to specify the exact unit that is going to be printed.
Instead the desired unit is the argument and one uses convert(var,Unit::KWH, u) before returning the value,
assuming var stores the value as KWH. The addPrint then suffixes the string with the desired unit.
If you need MegaJoule you add --addconversions=mj to get all energy units as mj as well in the json.

The reason for this is that Wmbumeters wants to convert all values into SI units (because the wmbus protocol supports all sorts of weird wacky units to minimize the work done by the meter software/hardware). 

However some meters can send either kwh or mj, therefore you need another convert in the processContent to normalize the incoming value into for example kwh. I suppose one could also remember which kind you unit was received with another boolean and use that boolean to switch between Unit::KWH or Unit::MJ in the addPrint lambda.
But that seems trickier somehow.

weetmuts:
A regression test will be needed, add a telegram in simulations/simulation_t1_meters.txt and update tests/test_t1_meters.sh and tests/test_listen_to_all.sh 

So that "make test" passes ok.

DenysFrasinich:
Thank you for for review. I`ll give it a try this weekend.

DenysFrasinich:
@weetmuts , could you plese review it again?

weetmuts:
Looks great! Many thanks!

PR #425, opened 2021-12-29
Add support to Sharky 774 meter.

DomAtHome:
I've added support to heat meter Sharky 774. I hope it will be useful to others.


weetmuts:
Thank you! But do you have an example telegram?

DomAtHome:
Yes, here it is:
`telegram=||5E44A5112751617241047A8B0050052F2F0C0E000000000C13010000000B3B0000000C2B000000000A5A26020A5E18020B260321000AA6180000C2026CBE2BCC020E00000000CC021301000000DB023B000000DC022B000000002F2F2F2F2F|`

DomAtHome:
@weetmuts I've updated the source code in sharky774.cc to handle your refactored code.

DomAtHome:
@weetmuts I've added tests to Sharky 774. Is there anything else to add before the pull request could be completed?

weetmuts:
Thanks! Great work! However the quantities are supposed to be unique, you only need a single Energy quantity,
then you have multiple unites for energy. The unit specified in the quantities table is only the default unit for energy.

So you can just remove EnergyGJ from LIST_OF_QANTITIES and replace EnergyGJ with just Energy in the meter code.

So wmbusmeters is supposed to standardize every energy unit into kWh (so MJ,GJ MWh etc)
then if you want GJ in the json you add --addconversion=gj and  you will get total_energy_consumption_gj
in addition to the existing total_energy_consumption_kwh

DomAtHome:
Thanks!
I didn't know about the unit conversion feature (I probably should've read the manual ;) ). I've removed the EnergyGJ form list of units and added "addconversions=GJ" to my config file. Works fine!

weetmuts:
Thanks!

PR #442, opened 2022-01-08
Fix typo in warning message

WagMat:
This is just a fix for a typo that I found during debugging the decoder for a new meter.

As I am new to GitHub I want to start with a simple change.


weetmuts:
Thanks!

PR #465, opened 2022-02-03
Fix: fixed typo in meter_qheat.cc

geektoor:
While testing wmbusmeters, i noticed a small typo in my output log.
Instead of last_year_date it was printing las_year_date. This is
now fixed.


weetmuts:
Thank you!

PR #473, opened 2022-02-08
Fix oneshot when more than one meter is defined but has not been started yet

ottmi:
Meters get only added to meters_ once a telegram has been received. If more than one meter has been defined, the check will already trigger on the first meter that receives a telegram although the others have not even been started yet.

I think this should fix it.


weetmuts:
Super! Thanks!

PR #478, opened 2022-02-17
Update meter_izar.cc

ldebomy:
Serial number starts at 2


weetmuts:
Thank you! When I rerun the tests on old telegrams, this fix will change the old serial numbers from 059196 to 159196
Ie they left pad the values with 0 to 6 digits. Now when we actually extract the most significant decimal digit, the zero turns into a one. Is this consistent with your own meter?

This will be a potentially breaking change to anyone who depends on the serial_number so I will document this in the CHANGES file. But hopefully, it might be that not so many users depend on this field.

weetmuts:
@don-vip @jacek27 Just checking that this change is ok?

don-vip:
Let me check for the meters I have. What's your model @ldebomy? Can you please share a frame?

don-vip:
@weetmuts it's ok! I didn't notice my mistake because my serial number starts with 0. Thank you @ldebomy!

weetmuts:
Thanks @Idebomy !

Issue #477, opened 2022-02-16
Diehl Sharky 774 - Partially Encrypted ? 


patou50:
I recently tried to read the data from my Diehl Sharky 774 heatmeter.

So, I first identified my Heatmeter (there are other nearby in the building). 
`wmbusmeters --t1 rtlwmbus`

```
Received telegram from: 58496405
          manufacturer: (DME) DIEHL Metering, Germany (0x11a5)
                  type: Heat meter (0x04) encrypted
                   ver: 0x41
                device: rtlwmbus[00000001]
                  rssi: 36 dBm
                driver: sharky774
```

I was dissappointed to see that it seems to be encrypted.

However, I tried my luck and tried the NOKEY option...
`wmbusmeters --logtelegrams --format=json auto:t1 heatmeter sharky774 58496405 NOKEY`

which outputs

```
telegram=|3E44A5110564495841047A700030052F2F#0C06846800000C13195364000B3B0400000C2B110100000A5A17050A5E76020AA61800004C0647630000426CBF25|+55
{"media":"heat","meter":"sharky774","name":"heatmeter","id":"58496405","total_energy_consumption_kwh":0,"total_volume_m3":645.319,"volume_flow_m3h":0.004,"power_kw":0.111,"flow_temperature_c":51.7,"return_temperature_c":27.6,"temperature_difference_c":24.1,"timestamp":"2022-02-16T18:52:29Z","device":"rtlwmbus[00000001]","rssi_dbm":30}
```

Apart from total_energy_consumption_kwh which is incorrect here, all variables correspond to what I can read with my eyes on the device.
Strangely, can I interpret this as there is no key ? However, very strangely, the total_energy_consumption_kwh is set to 0. It seems to be the same for my neighbours. The correct value is obviously not 0.
Is this somehow related to the key ? So the key would only allow to get this variable extra ?

Germar:
Same here. I can see all correct variables except `total_energy_consumption_kwh` with Diehl Sharky 774 and IM871A-USB.

patou50:
I have also seen the exact same value of 220000 for variable power_kw.

Have you ever bothered contacting your supplier to get the key ?

weetmuts:
Can you post a plain rtl_sdr | rtl_wmbus line of telegram here? It is weird that the telegram does not need to be decrypted but still reports in the header that it is encrypted......

Germar:
Works like a charm! Thanks @weetmuts
```
[...]

PR #480, opened 2022-02-18
detect negative BCD value

Germar:
This PR introduce a check for negative BCD values like discussed in #477
Fixes `"power_kw":220000.009` values with Diehl Sharky 774 heatmeter.

Please double check my changes as this is my first contribution in C++. I'm used to code in Python only.


weetmuts:
Great work! Many thanks!

PR #482, opened 2022-02-21
escape command substitution for logrotate

Germar:
The command `cat /run/wmbusmeters/wmbusmeters.pid 2> /dev/null` in `scripts/prepare_logfiles` wasn't escaped. So the resulting logrotate file had a broken postrotate command
```
postrotate
        /bin/kill -HUP  2> /dev/null || true
```
This made wmbusmetersd hang after logrotate moved the logfile


weetmuts:
Thanks!

PR #492, opened 2022-03-05
Qwater meter & update for Qheat

TheZingo:
I discovered too late that there is already support for Qwater under the name lse_07_17, so I derived another Qwater driver from Qheat. Maybe that is still interesing for the project. It runs already successfully since some weeks. 
Furthermore, I've added another media/version combination for the Qheat driver.


weetmuts:
Thanks @TheZingo ! Do you have any test telegrams that we can add to prevent regressions?

I am rewriting the drivers from meter_xxxx.cc to driver_xxxx.cc which is much more maintainable. However I need telegrams to prevent regressions.

petterreinholdtsen:
Perhaps @TheZingo can provide telegrams, if he is pointed to a recipe on how to generate the files you want?

TheZingo:
Sure, I hope these are the values you are looking for:

QHeat:
telegram=|4944934413566513180778#0DFF5F350082AB0000EE0007C113FFFF24430300BF2A39220200DF2367220300B400B100CE00B800CF00B800CF00C800CB00C400B400E0002F046D1C11DE24|+136
QWater (two different meters):
telegram=|4944934413566513180778#0DFF5F350082AB0000EE0007C113FFFF24430300BF2A39220200DF2367220300B400B100CE00B800CF00B800CF00C800CB00C400B400E0002F046D1C11DE24|+136
telegram=|4944934418292312180678#0DFF5F350082F90000800007C113FFFF27300200BF2A52390100DF23401502007F008000740051006B0057008D008400AE00940095009C002F046D1D11DE24|+141

weetmuts:
Thanks @TheZingo ! However these example telegrams do not contain the desired data that your new driver can decode.
Using wmbusmeters --analyze <hex>
or online here:
https://wmbusmeters.org/analyze/49449344135665131807780DFF5F350082AB0000EE0007C113FFFF24430300BF2A39220200DF2367220300B400B100CE00B800CF00B800CF00C800CB00C400B400E0002F046D1C11DE24

It seems like the most of the data is stored in a vendor extension.

Does the driver work for you, and does it give the right results?

weetmuts:
Hello @TheZingo ! Thank you for your pull request, but we need some working telegrams.
Also there is a new much better driver format that I am converting the meter_xxx to.
The new drivers are named driver_xxx and here is an example:
https://github.com/weetmuts/wmbusmeters/blob/master/src/driver_iperl.cc
where the test cases are included in the driver_xxx.cc source file at the end.

You do not need to rewrite your code to be the new driver format, I can do that,
but I need some example telegrams. When I look at your code it looks like
the do expect standard wmbus APL data, ie dif/vif entries, but in the two
examples you posted above, there is only a single big varlen proprietary blob.

Do you think you can collect some more telegrams? Sometimes the meters
send multiple types of telegrams, some more standard and some more proprietary.

You can use wmbusmeters.org to take easily look at the content of the output from --logtelegrams.

Thanks!

weetmuts:
Most of this pull request has been integrated into th driver_qwater.cc file. Closing this one.
Thank you for your work anyway @TheZingo !

If you can find the telegrams with the qheat type/version combo and the qwater type/version combo
for which we do not yet have test-cases. Please submit them in a new issue or pull request! Thanks!

weetmuts:
FYI this is the latest qwater update: https://github.com/weetmuts/wmbusmeters/pull/613

PR #516, opened 2022-04-18
Add bash to docker image builders

pidpawel:
CI was failing on `generate_authors.sh` thus preventing builds for Docker. I've managed to find that the issue is due to lack of Bash in default Alpine Linux images used for Docker builds. This PR adds bash to all Dockerfiles. Hope it helps!


weetmuts:
Many thanks!

PR #522, opened 2022-04-26
Update auto discover for HA

libo:
- Change the device to use auto discovery
- Don't probe /dev/ttyAMA0 that upsets Raspberry PIs

After the changes in this PR I could see on my setup:
```
[...]
Running wmbusmeters ...
No wmbus device detected, waiting for a device to be plugged in.
No meters configured. Printing id:s of all telegrams heard!
```
Then I plugged in my dongle and...

```
Started auto rtlwmbus[00000001] listening on t1
Received telegram from: [redacted]
          manufacturer: (KAM) Kamstrup Energi (0x2c2d)
                  type: Cold water meter (0x16) encrypted
```

From README.md
```
It is recommended that you use auto to find your dongle, then, when you know the exact device path, you write for example: device=/dev/ttyUSB0:im871a:c1,t1 in the configuration file or on the command line. This will skip the slow probing for all possible dongles when wmbusmeters starts up.

If you have to scan serial devices, then remember that some Raspberry PIs are upset when random data is sent to /dev/ttyAMA0 when it is configured in bluetooth mode. To solve this, add donotprobe=/dev/ttyAMA0
```

cc: @to-masz @rafalstarczak @weetmuts

### DIFF
--- a/ha-addon/config.json
+++ b/ha-addon/config.json
@@ -20,7 +20,7 @@
     "options": {
         "data_path": "/config/wmbusmeters",
         "conf":
-            "loglevel=normal\ndevice=/dev/ttyUSB0:t1\nlogtelegrams=false\nformat=json\nlogfile=/dev/stdout\nshell=/wmbusmeters/mosquitto_pub.sh wmbusmeters/$METER_NAME \"$METER_JSON\"",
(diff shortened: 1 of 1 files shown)


weetmuts:
I suppose it is an ok change. But can you do a local change to the installed ha-addon? 
Since auto is not recommended for production use.

Or will it be overwritten the next time wmbusmeters publishes a new ha-addon version?

thecem:
I would propose the accept the change an test it with the new ha-addon version. I will test.

weetmuts:
@thecem @to-masz @libo Please test this change to the HA.

PR #540, opened 2022-05-09
Update CreateDriver.md

Kosygor:
I believe there were two typos. Also added /etc directory since it is default one.


weetmuts:
Thanks!

PR #545, opened 2022-05-14
Update config.json

piio:
Adding "init": false, to ha addon configuration to avoid "s6-overlay-suexec: fatal: can only run as pid 1"


weetmuts:
Ah! I already made the change myself! But thanks for the fix!

PR #553, opened 2022-05-30
Enable CI on all branches

dwrobel:
Enables CI on every branch to make it usable on
pull requests and forked repositories.

### DIFF
--- a/.github/workflows/build_ubuntu.yml
+++ b/.github/workflows/build_ubuntu.yml
@@ -1,9 +1,7 @@
 name: Build Ubuntu
 on:
   push:
-    branches:
-      - master
-      - 'debug/**'
+  pull_request:
 
 jobs:
   build:


weetmuts:
When someone creates a pull request or a fork. Will the build run with their credentials or mine?

dwrobel:
> When someone creates a pull request or a fork. Will the build run with their credentials or mine?

When someone has its own fork then they are on their own. If someone creates a PR on your fork then it looks like the following:
- https://github.com/weetmuts/wmbusmeters/pull/552/checks _(that's the one we intentionally would like to fail)_.
- https://github.com/weetmuts/wmbusmeters/pull/553/checks _(it's for this PR)_,

As you might noticed anyone could enable CI on your repository. Here you can read specifically about authentication: https://docs.github.com/en/actions/security-guides/automatic-token-authentication.

weetmuts:
I think the intent from github is to run pull_requests without access to the secrets. So this should be safe, unless there are bugs in github.... :-)

https://blog.teddykatz.com/2021/03/17/github-actions-write-access.html

PR #574, opened 2022-07-26
Fix compilation error on gcc 12

dwrobel:
Fixes the following error:
  src/util.cc: In function ‘bool enableLogfile(std::string, bool)’:
  src/util.cc:379:22: error: ‘time’ was not declared in this scope; did you mean ‘utimes’?
    379 |         time_t now = time(NULL);
        |                      ^~~~


weetmuts:
Thanks!

PR #575, opened 2022-07-28
MQTT discovery

jnxxx:
This is a proposal for adding MQTT discovery to the HA addon.

For known meters (types that has a template file) it can automatically tell HA which sensors to create and how to read the payload.
Thus, manually adding the sensor definition to `configuration.yaml` is not needed. Rather than each user defines this, it can be done once and for all per meter and new users just get it automatically.

#### How it works:
As it is implemented in bash script it is made very simple.
It enumerates the defined meters and specificially reads the driver value. Based on the driver it reads a template file for the meter type. This file defines which sensers to create, much like the information that can manually be inserted in the configuration file. Some values are replaced, and finally it is published in MQTT and the sensers appear in HA.

Due to the dependency on the driver value, it cannot work with auto. But it seems it is not recommended for production workloads anyway.

#### Example for a simple sensor for multical21:
```
{
    "total_m3": {
        "component": "sensor",
        "discovery_payload": {
            "device": {
                "identifiers": ["wmbusmeters_{id}"],
                "manufacturer": "Kamstrup",
                "model": "{driver}",
                "name": "{name}",
                "sw_version": "{id}"
            },
            "enabled_by_default": true,
            "json_attributes_topic": "wmbusmeters/{name}",
            "state_class": "total",
            "name": "{name} total",
            "state_topic": "wmbusmeters/{name}",
            "unique_id": "wmbusmeters_{id}_{attribute}",
            "unit_of_measurement": "m³",
            "value_template": "{{ value_json.{attribute} }}",
            "icon": "mdi:gauge"
        }
    },
...
```
Curly brackets like `{driver}` is replaced with the corresponding value from the meter definition. `{attribute}` specifically does not come from the meter definition, but is just the value from the top node, in the case "total_m3". It just makes it easier to copy for reuse.

In the sensor definition I have included a device definition.
When seen from within HA it looks a lot more convincing I think, rather than just a number of unrelated sensors.

**Binary sensors**
In the first template file for multical21, I have also made some examples of binary sensors for the four warning indicators: dry, burst, leak and reversed.

**Availability**
Also it has an example of availability for the temperature readings. For example my meter read 127 all the time, and I can see in the code it does that when meter does not deliver any value. This seem to match the R value of my meters configuration. Thus, if the reading says 127 rather than showing this value it says unavailable, and for others it shows the reported value.

#### Template files source
The template files are taken directly from the repository rather than including these in the docker image.
This may not be best practice, but if it turns a success it may require quite a few additions or updates of these template files.
As the container is build locally, it puts some [wear and tear to users SD card](https://developers.home-assistant.io/docs/add-ons/publishing#locally-build-containers). Rather than cause an abundance of builds, I decided for this method (but can of course be changed). The files in mqtt_discovery are copied to ~/config/wmbusmeters/etc/mqtt_discovery where it is used from. Here it is also easy to make changes or add and test new types locally.

#### For it to work:

1. Meter must be defined with a specific driver (not auto, which isn't recommended for production anyway)
2. A template file must have been defined for the driver type.
3. MQTT discovery must be enabled in the config file.

If MQTT discovery is not enabled, cleanup will still run to remove what may have previously been published.

image

image

image

cc: @to-masz @rafalstarczak @weetmuts

### DIFF
--- a/ha-addon/Dockerfile
+++ b/ha-addon/Dockerfile
@@ -24,6 +24,9 @@ COPY --from=build /wmbusmeters/build/wmbusmeters /wmbusmeters/wmbusmeters
 COPY --from=build /rtl-wmbus/build/rtl_wmbus /usr/bin/rtl_wmbus
 COPY --from=build /rtl_433/build/src/rtl_433 /usr/bin/rtl_433
 
+COPY mqtt_discovery.sh /
+RUN chmod a+x /mqtt_discovery.sh
+
 COPY run.sh /
 RUN chmod a+x /run.sh
 
--- a/ha-addon/config.json
+++ b/ha-addon/config.json
@@ -27,6 +27,7 @@
     },
     "schema": {
         "data_path": "str",
+        "enable_mqtt_discovery": "bool?",
         "mqtt": {
           "host": "str?",
           "port": "port?",
--- a/ha-addon/config.json
+++ b/ha-addon/config.json
@@ -20,6 +20,7 @@
     ],
     "options": {
         "data_path": "/config/wmbusmeters",
+        "enable_mqtt_discovery": false,
         "conf":
             "loglevel=normal\ndevice=/dev/ttyUSB0:t1\nlogtelegrams=false\nformat=json\nlogfile=/dev/stdout\nshell=/wmbusmeters/mosquitto_pub.sh wmbusmeters/$METER_NAME \"$METER_JSON\"",
         "meters": [],
--- a/ha-addon/mqtt_discovery.sh
+++ b/ha-addon/mqtt_discovery.sh
@@ -11,18 +11,18 @@ do
     esac
 done
 
+# Get discovery_prefix
+if bashio::config.exists "mqtt.discovery_prefix"
+then MQTT_DISCOVERY_PREFIX=$(bashio::config "mqtt.discovery_prefix")
+else MQTT_DISCOVERY_PREFIX=$(bashio::services mqtt "discovery_prefix"); fi
+if [ "${MQTT_DISCOVERY_PREFIX+true}" ] || [ $MQTT_DISCOVERY_PREFIX != "null" ];
+then MQTT_DISCOVERY_PREFIX="homeassistant"; fi
+
 
 # Is MQTT discovery enabled?
 CONFIG_MQTTDISCOVERY_ENABLED="$(jq --raw-output -c -M '.enable_mqtt_discovery' $CONFIG_PATH)"
 if [ $CONFIG_MQTTDISCOVERY_ENABLED == "true" ]; then
     echo -e "\nMQTT Discovery ..."
-
-    # Get discovery_prefix
-    if bashio::config.exists "mqtt.discovery_prefix"
-    then MQTT_DISCOVERY_PREFIX=$(bashio::config "mqtt.discovery_prefix")
-    else MQTT_DISCOVERY_PREFIX=$(bashio::services mqtt "discovery_prefix"); fi
-    if [ "${MQTT_DISCOVERY_PREFIX+true}" ] || [ $MQTT_DISCOVERY_PREFIX != "null" ];
-    then MQTT_DISCOVERY_PREFIX="homeassistant"; fi
 
     # Copy template files
     templatedir="$(mktemp -d -p /dev/shm/)"
--- a/ha-addon/mqtt_discovery.sh
+++ b/ha-addon/mqtt_discovery.sh
@@ -46,7 +46,7 @@ if [ $CONFIG_MQTTDISCOVERY_ENABLED == "true" ]; then
         echo " Adding meter: ${aryKV['name']} ..."
 
         if [ "${aryKV['id']+true}" ] && [ "${aryKV['driver']+true}" ] ; then
-            file="$CONFIG_DATA_PATH/etc/discovery/${aryKV['driver']}.json"
+            file="$CONFIG_DATA_PATH/etc/mqtt_discovery/${aryKV['driver']}.json"
             if test -f "$file"; then
 
                 for attribute in $(jq --raw-output -c -M '. | keys[]' $file)
--- a/ha-addon/mqtt_discovery.sh
+++ b/ha-addon/mqtt_discovery.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/with-contenv bashio
+
+aryDiscTopic=()
+pub_args=()
+while getopts h:p:u:P:c:w: flag
+do
+    case "${flag}" in
+        h|p|u|P) pub_args+=("-${flag}" ${OPTARG});;
+        c) CONFIG_PATH=${OPTARG};;
+        w) CONFIG_DATA_PATH=${OPTARG};;
+    esac
+done
+
+
+# Is MQTT discovery enabled?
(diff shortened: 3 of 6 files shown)


Bluhme1:
Recomended. I use the PR from jnxxx and it works perfectly for Multical21 (water) ande multical603 (district heating)

weetmuts:
Very nice indeed! I am not an expert on HA but I can merge this, since @Bluhme1 confirms it is working. :-)
This will not mess up existing installations where the user has manually configured their sensors? 

It seems like the template for the sensors should be possible to autogenerate from wmbusmeters meter configuration.
The multical21 driver has not yet been converted to the new driver format, check out the driver_waterstarm.cc for example.
Is there any additional information in the driver_waterstarm.cc that would be needed to auto-generate the ha sensor file?

jnxxx:
> This will not mess up existing installations where the user has manually configured their sensors?

It is off by default, so will require users to enable it specifically.
If manually configured, it depends on how it was configured, if sensor names are the same or different.
If same name or unique_id it could interfere, but likely it is the same value, so I do not see it would harm.
If different name, if would just result in two sensors in HA. And then the manual one can be removed if desirable.

The the sensor in the example does not have long-term statistics, so by default it only has 10 days of history. So not a huge issue to have it switch to a different name, but automations, UI elements may have to be adjusted.
   
   
> It seems like the template for the sensors should be possible to autogenerate from wmbusmeters meter configuration.

I definitely see an idea in extracting that information from the new driver type. Otherwise it is a pretty large task to get all defined.
But do you mean to parse the source files with a script or to do it in the inherited base class?
Parsing could easily be error prone due to various ways of formatting.

Regarding data availability for the templates:

Driver name seems easy.

With manufacturer and model I see the first challenge. Not due to missing info, but because a driver can map to multiple manufacturers and models (di.addDetection lines).
The MQTT discovery runs and is inserted before wmbusmeters startup, and at this point we do not have an example of the data. Only knows the driver from the meters defined to retrieve data from.
Even if postponed and done in the mosquitto_pub.sh script, we do not have manufacturer in the payload and I wonder if meter would be the actual meter type or just driver?
Anyways, these are purely informational and nice to have. Not strictly required.
Maybe also not a good idea to place `{id}` in `sw_version`, but there just isn't an identifier on a device.

It seems `addNumericFieldWithExtractor` would translate into sensors.
 Json attribute name ~ vname + vquantity?
 Unit = vquantity
 Can the following be derived from vquantity?
- [state_class](https://developers.home-assistant.io/docs/core/entity/sensor#available-state-classes): needs to be `total` (for incrementing values), `measurement` or not defined.
(only when defined and a suitable device_class and unit is used, long term statistics will be collected)
- device_class: A value from [here](https://developers.home-assistant.io/docs/core/entity/sensor#available-device-classes) or not defined
- icon: some mdi:icon, or perhaps left out. device_class should provide a reasonable default.
 
 Availability could probably be detected on if the attribute is in the payload or not, and value != null.


`addStringFieldWithExtractorAndLookup` seem to be a good source for status codes.
I guess each defined error value should be defined as a binary_sensor. For example one named SW_ERROR which looks for that value in the current_status string.
It will be difficult though to set a meaningful [device_class](https://www.home-assistant.io/integrations/binary_sensor#device-class) or icon.

I do not see rssi_dbm defined, but maybe it is always there, so could also just always be added to the template.

Lastly, I think it will be difficult to determine a reasonable value for `enabled_by_default`.
The intent with this is to have values only used in rare cases disabled, but can easily be enabled from the UI.
I guess then everything, except rssi, can just be enabled by default.

And similarly with `entity_category`. This one could be left out for sensors, and I guess for binary_sensors set to diagnostic.

weetmuts:
Thank you for your thoughts on automating the templates! I am working towards a more declarative and smarter way of maintaining the drivers. I will keep these manual templates in mind for the future work. Merging now.

@Bluhme1 @thecem @to-masz @libo Please test this change to the HA if you have the time.

thecem:
Working with 603 driver. What is about the Firmware? Is there no possibility to change this to Serial?

grafik

Bluhme1:
Hasn’t shown up as upgrade yet, so I haven’t installed. Will return with comments later

thecem:
@Bluhme1 : You could first update your addons and then only hit "rebuild", then you have the latest updates.

@weetmuts : I createtd a PR https://github.com/weetmuts/wmbusmeters/pull/602 that includes the MC303 with adaptions.

I update to a final version, that fits the new driver (multical303) value and field namimg. Additional I propose to change the naming of all multical driver to a short and same naming. So not every kamstrup device must be adapted in there own value naming.

Bluhme1:
Is there a possibility to make a manual update. Can’t see an automatic update yet

thecem:
You will not see an update only if the version of
the addon is changed. Hit “
Neu Bauen” will update yours to the latest.

weetmuts:
I was thinking about how to deprecate old drivers, so that we can normalize field names in the json.
It should not happen often in the future, but we need to cleanup a couple of drivers ie all the multical heat drivers and others. This is definitely not something that is going to happen often in the future, but we need a polite way of doing it now.
I like backwards compatibility very much, but sometimes you have to break it.

Does the HA-addon currently listen to the alarm? The alarm shell is designed to post a message to for example HA to notify the user with a underlying problem in the data collection. I.e. a dongle no longer responds etc.

But we could also use the alarm to notify the user that the driver they are using is going to be deprecated.
A warning once per week, then one per day the last week should give the user enough time to do something about it.
So a deprecated driver could be removed 3 months after the replacement driver is in place.

I could also add a deprecation message in every telegram json...

Is there a way to auto-update the HA config? Or can we just create a new discovery rule that creates a new HA config with the new driver? Then the user work is perhaps as simply as a click to accept the new auto-discovered meter?

@BIBOLV @thecem @to-masz @rafalstarczak @jnxxx

thecem:
What I did for testing:

doubled the two heat meters in wmbusmeters config. One with the multical303 one with the multical603. So I save the manual HA configured meters and they workings as expected, the second pair config in wmbusmeters are with 603 driver and discovered by HA as expected ( beside the values, but of wrong naming in the 603 driver).
So both are working in parallel. Even the naming of mine in config are different to the HA discovered.
So with the discovery Funktion it would really easy to switch to the new, beside the name changing for HA entity.
Since the main sensor includes all values as attributes, it should be easy to stay on same entity name and get the value you need out of the attributes.

May this could be an smart way to switch to the new code To use the discovery function.

jnxxx:
I got version 0.2.23 offered, and it is working for me.

Now that the template files are in the `weetmuts` repo, I suggest to update the file `mqtt_discovery.sh` to fetch the templates from there.
Line 29: replace `jnxxx` with `weetmuts`.


@thecem, unfortunately in HA a device does not have an identifier or serial property. At least I have not been able to find it.
An alternative could be a Diagnostic sensor named Id or so.

@weetmuts, No. The HA add-on only has an entry for shell.
Looks like it is something that has to be added, but I am not sure where to put it. Except maybe a binary sensor that could indicate an issue, and then more details can be found in the log.
Is there an example of the argument values somewhere?

I think, if version number is updated many HA users will update it within long. Some may even have enabled auto update.

But if field names in the json data is updated, it may be a bad idea to read templates directly from the repository. Because then the update of the driver and template is not necessarily aligned.
Reason for taking templates directly from the repo, was to allow easier updates and additions of templates. Then it would only require a restart and not a full update (incuding git clone, compile source code and build docker image which does put some strain on an SD card).

PR #601, opened 2022-09-04
Keep environment in shell subprocesses

kleinweby:
Currently we replace the environment of subprocesses with our own,
primarily only consisting of METER_* entries. Besides clearing
some very useful system dependent variables (like PATH, LDPATH, etc.)
with also prevent the user from passing environment variables to
invoked commands.

This patch will now copy the current environment and extends it with
the variables we want to set.

---

This is essentially another take on #385 although my motivation is different. I simply wanted to pass down some configuration variables to the shell command I was running. This is now possible, while also setting the intended `METER_*` variables.

### DIFF
--- a/src/shell.cc
+++ b/src/shell.cc
@@ -26,6 +26,39 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+// Posix says that this variable just exists.
+// (On some systems this is also declared in unistd.h)
+extern char **environ;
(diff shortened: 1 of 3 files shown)


weetmuts:
Yes, and it should solve the problem with the missing HOME directory for mqtt_pub that has been annoying!
Thank you very much!

Issue #352, opened 2021-10-05
Qundis Q module 5.5 water Modularis is supported by lse_07_17


scy:
Hi there!

In my apartment, there are water meters supplied with [Q module 5.5 add-on modules](https://qundis.com/products/modules/) that make them C-mode enabled.

These send a manufacturer of `(QDS) Qundis, Germany (0x4493)`, type `Radio converter (meter side) (0x37)` and `ver: 0x33`. I’ve tried setting the driver to `lse_07_17` (the QWater 5.5 introduced in #277), and it worked instantly 😁 🎉 

The cold water meter correctly reports `media: water` and the correct `total_m3` value; the warm water meter (same mfct/media/version numbers) reports `media: warm water` and the correct `total_m3`, too.

This is awesome, I got my RTL-SDR today and just wanted to track my heat cost allocator values, but I got my water meters at no additional cost, too 😉  Thanks for this project!

weetmuts:
Great to hear that! I have added the new mapping to the LSE_07_17 driver from QDS 0x37 0x33.
Please test if you have the time!

weetmuts:
If you have a telegram, please post it here!

We need it for the regression tests! Thanks!

scy:
I’ll test it and supply you with a telegram, but it will take a few days until I find the time, sorry :)

weetmuts:
No worries! Thanks!

scy:
Took me some time, but here are your telegrams. :) Note that I did _not_ update wmbusmeters to a version with your change from Oct 16, so the meters are still detected as “unknown”.

First, two random meters from somewhere inside the house.

```console
$ rtl_sdr -f 868.95M -s 1.6M - | rtl_wmbus -p s -a | wmbusmeters --logtelegrams stdin:rtlwmbus
[…]
Received telegram from: 35207125
          manufacturer: (QDS) Qundis, Germany (0x4493)
                  type: Radio converter (meter side) (0x37)
                   ver: 0x33
      Concerning meter: 35207125
          manufacturer: (QDS) Qundis, Germany (0x4493)
                  type: Warm Water (30°C-90°C) meter (0x06)
                   ver: 0x33
                device: rtlwmbus[]
                  rssi: 16 dBm
                driver: unknown!
telegram=||3C449344257120353337722571203593443306420000200C13870200004C1364000000426C9F2CCC081387020000C2086CBE2B326CFFFF046D0B00BE2C|+15
[…]
[...]

PR #609, opened 2022-09-11
ha-addon: Add discovery for qcaloric & lse_07_17

patrislav1:
This is a patch to add MQTT discovery support for the Qundis heat cost allocator (qcaloric) and water meter (lse_07_17)

* Only the main measurand (total consumption of water / heat) and the RSSI is included - not sure how any of the other available values would be useful
* For the heat cost allocator, I called the measurement unit for total consumption "hca" - Not sure if there is a better name for it

Note that I have the Q module 5.5 water and had to manually set the driver to lse_07_17, even though it should have auto-detected it, if I understand #352 correctly. Should I submit some packets, so the auto-detection can be improved?

### DIFF
--- a/ha-addon/mqtt_discovery/lse_07_17.json
+++ b/ha-addon/mqtt_discovery/lse_07_17.json
@@ -1,0 +1,49 @@
+{
+    "total_m3": {
+        "component": "sensor",
+        "discovery_payload": {
+            "device": {
+                "identifiers": [
+                    "wmbusmeters_{id}"
+                ],
+                "manufacturer": "Qundis",
+                "model": "{driver}",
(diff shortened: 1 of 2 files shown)


weetmuts:
Thanks!

weetmuts:
"Should I submit some packets, so the auto-detection can be improved?"

Yes please! Open an issue!

patrislav1:
> "Should I submit some packets, so the auto-detection can be improved?"
> 
> Yes please! Open an issue!

Ok I attached [my log](https://github.com/weetmuts/wmbusmeters/issues/352#issuecomment-1245157167) to #352 as it seems to be the same issue.

Issue #598, opened 2022-08-31
Missing driver for Apator E-ITN 30.51 HCA, A lot of Unknown tpl-ci-field b6 in logs


kitor:
Hello,
I live in a multi-storey building with a lot of wmbus devices. Those are:
- `apator162` water meters
- `amiplus` power meters
- `Apator E-ITN 30.51` HCAs (not supported yet)

I use rtl-sdr, wmbusmeters config is (frequency was fine-tuned with spectrum analysis):
```
loglevel=normal
device=rtlwmbus:868.993M:t1
logtelegrams=false
format=json
meterfiles=/var/log/wmbusmeters/meter_readings
meterfilesaction=append
logfile=/var/log/wmbusmeters/wmbusmeters.log
```

`wmbusmeters` is spamming a ton of `(wmbus) Unknown tpl-ci-field b6` messages, doesn't matter if I have any of `apator162` or `amiplus` sensors configured in `/etc/wmbusmeters.d/*` or not (running in `logtelegrams=true` mode). 

What causes this message? Is this related to any of non-supported devices? I guess so, as those usually prepend HCA messages [1]

There's a sensor documentation (Polish only) but it doesn't tell anything about the protocol itself.
http://www.apator.com/uploads/files/Produkty/Podzielnik_kosztow_ogrzewania/i-pl-021-2016-e-itn-30-51-30-6.pdf
At least it lists that 30.51 meter transmits:
* serial number
* season start date
* read date
* electronic seal break date
* previous season average temperature
* this season average temperature

(there's also 30.6 with a bit different data transmitted, but I don't have access to that sensor)

I already partially reverse engineered protocol for `Apator E-ITN 30.51`:
Looks to be a proprietary protocol, which seems to be a theme for Apator devices. At least data is not encrypted in any way.

```
  ?  season      ts?    ?     ?   last     ?  this    current avg   ? 
     start?                       season      season? date    temp
F545 0186   F41B A62B  A0A1 0000  1002   0000  0000   192D    9815 B81A
F545 0186   F41B 985D  A0A1 0000  0000   0000  0000   192D    7415 881A
F545 0186   F41B A358  A0A1 0000  0000   0000  0000   192D    AC15 BC19
F545 0186   F41B 9F3F  A0A1 0000  6400   0000  0000   192D    D015 1C1B
```

`last season` is just BE uint16_t with reading.

[...]

PR #617, opened 2022-09-20
Full implementation for Apator E-ITN 30.51

kitor:
Full implementation of protocol for Apator E-ITN 30.51, as discussed in #598 

So far tested on 4 devices I have access to, and a dozen ones I don't have access but I can receive telegrams (this is how I got `esb_date` by the way)

Value for `season_start_date_` may be wrong. But this is the only part of telegram that matches season start data in any familiar format.

Not sure how to deal with date properly, including the fact that electronic seal break date may (and well, should) be `0000`.

Driver "Test" section needs to be updated, I'm not sure how to generate that?

and I guess I should add `t->addSpecialExplanation` for all fields, right?

Data example (outdated)
```
{'current_date': '2022-09-20T01:00:00Z',
 'current_hca': 0,
 'device': 'rtlwmbus[00000001]',
 'esb_date': '',
 'id': 'zzzzzzzz',
 'media': 'heat cost allocation',
 'meter': 'apatoreitn',
 'name': 'heat',
 'previous_hca': 100,
 'rssi_dbm': 9,
 'season_start_date': '2016-05-01T01:00:00Z',
 'temp_room_avg_c': 23.484375,
 'temp_room_prev_avg_c': 21.8125,
 'timestamp': '2022-09-20T11:33:10Z'}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,7 @@ Waterstar M (waterstarm)
 Zenner Minomess (minomess)
 
 Supported heat cost allocators:
+Apator E-ITN 30.51 (apatoreitn)
 Innotas EurisII  (eurisii)
 Qundis Q caloric (qcaloric)
 Sontex 868 (sontex868)
--- a/src/driver_apatoreitn.cc
+++ b/src/driver_apatoreitn.cc
@@ -1,5 +1,6 @@
 /*
  Copyright (C) 2022 Fredrik Öhrström (gpl-3.0-or-later)
+ Copyright (C) 2022 Kajetan Krykwiński (gpl-3.0-or-later)
 
  This program is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


weetmuts:
Nice! You run the test with "make test" which runs all tests.
You can also only ring the source code embedded tests for this driver only:
"make testdriver DRIVER=apatoreitn"

If you often do "make clean" then you can speed up the development cycle with:
"make DRIVER=apatoreitn"
this will create a build/wmbusmeters binary that only supports one driver.

weetmuts:
You add tests like this:
// Test: HCA apatoreitn 37373737 NOKEY
// telegram=|25441486373737370408B60AFFFFF5450186F41B9D58A0A100007809000000001F2D6416C819|
// {"media":"heat cost allocation","meter":"apatoreitn","name":"HCA","id":"37373737","allocation_hca":0,"room_c":22.390625,"timestamp":"1111-11-11T11:11:11Z"}
// |HCA;37373737;22.390625;0;1111-11-11 11:11.11

You can have multiple triplets  with telegram,{,| for the same meter
and you can have multilpe different meter setups (ie "// Test:....")

kitor:
As described in #598 I am waiting for info how to deal with B6 field properly.
Meter sends optional, variable length header (if B6 header is present, length is given in next byte). In each case observed so far, last 16 bytes contain regular, expected data.

kitor:
@weetmuts I'm not sure why `fields` format is missing some data.
I hope handling for `tpl-ci-field` is also OK.

Other than that, driver should work for all known formats so far (including variable length B6 header)

weetmuts:
Looks nice! Dealing with the tpl_ci oddity looks good enough to me. :-)

PR #618, opened 2022-09-21
Add support techem devices autodiscovery

jsuchal:
Adds support for Techem devices autodiscovery.


jsuchal:
Also I believe I've fixed the wrong path to download current discovery definitions. This is weird btw - should this be used from the actual code, not downloaded from internet?

weetmuts:
Thanks! Yes I suppose it is an artefact of the mqtt discovery first pull request. 
@jnxxx Right?

jnxxx:
Right. It should be changed to weetmuts/wmbusmeters repo.

PR #637, opened 2022-10-13
feat: Add quoting to prevent word splitting

RiddleMan:
Adds single quotes to the `mosquitto_pub.sh` file to prevent issues while having special shell characters in any values provided to `mqtt` settings.


weetmuts:
Thanks!

PR #641, opened 2022-10-18
Add QHeat 5.5 US(R) driver (ver 0x46)

MightyPork:
This is a version of the QHeat 5.5 driver that works with a payload from meter with version 0x46

related issue https://github.com/weetmuts/wmbusmeters/issues/625

Sample telegram:

```
41449344796550674637727965506793444604dc0000200c0d000000004c0d00000000426cffffcc080d00000000c2086cdf2802fd170000326cffff046d3a0ddb29
```

There are some things that need resolving:

- What driver name to use? The super secret PDF with specification calls it "Q OMS / Q AMR" or "Q heat 5.5 (US)R (Version 60/70)"
- Why did FieldMatcher not work here? (I used DifVifKey as a workaround)
- The error flags decoding might be wrong

### DIFF
--- a/src/driver_qheat_55_usr.cc
+++ b/src/driver_qheat_55_usr.cc
@@ -1,0 +1,185 @@
+/*
+ Copyright (C) 2021-2022 Fredrik Öhrström (gpl-3.0-or-later)
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


MightyPork:
I'm now testing this branch and it seems that the consumption might be in the wrong unit, JSON says _gj but the value looks more like kWh?

PR #644, opened 2022-10-21
added MQTT discovery for sensostar and extended statuses supported by driver

ovettel:
- Added the required file to support MQTT discovery for sensostar meters (first commit).
- In the second commit, I extended the statuses supported by the sensostar driver based on the vendor spec + added them to MQTT discovery. Note that I was not able to test/build this new source, so I hope this is ok.


weetmuts:
Thanks! Great work in figuring out the meaning of the error bits!

PR #651, opened 2022-10-29
HA add-on: Retain last message in MQTT broker

lassebm:
This is required to restore last state in Home Assistant after restarts.


weetmuts:
Cool, this looks like something we should always use for mosquitto_pub? 

Should we update the mosquitto_pub usages in README as well?

lassebm:
As HA doesn't cache values from MQTT data sources, it's required to restore last state, but it might not be desirable for other use cases.

weetmuts:
Thanks!

PR #687, opened 2022-11-21
Parse history for Lorenz water meters

bb-froggy:
Hi,

driver_waterstarm.cc parses Lorenz water meters already. There was one id missing, but then it worked for the ones I have tried.

However, the current driver only parses some very basic data, but the manufacturer told me that the water meters I have have been configured for the long message format and indeed there is more information that driver_waterstarm.cc did not yet parse. I have added that code, which parses the consumption history for the last 15 months from the telegram.

But now the tests fail as the new code doesn't work correctly anymore for the short type telegrams that were included in the tests already. I don't even know what is supposed to be returned in the case of a short telegram. Should it be "0.0" m³ consumption for the historic dates? This is not correct, since we actually just don't know the consumption, which is different from "0.0" m³. But it seems that I have to add which fields are there before actually parsing the telegram, so at a point where I don't know whether it is a short or long telegram.

In addition, addNumericFieldWithExtractor seems not to handle situations well when the field does not exist or I have a bug in the code ... if it is a short message, there is always one historic value that contains garbage from memory, i.e. very high consumptions. I think I can avoid this if I just call "addPrint" in the constructor and fill the actual values in the processContent method. I have the code in my commit already, but commented out.

So, at this point I would be glad to receive some help with the question:
**How can I deal with fields that may or may not be present in the telegram and I only know whether they are present in the telegram after parsing the message?**

### DIFF
--- a/simulations/simulation_t1.txt
+++ b/simulations/simulation_t1.txt
@@ -193,8 +193,8 @@ telegram=|4644B4097172737405077AA5000610_1115F78184AB0F1D1E200000005904103103208
 # Test waterstarm
 
 telegram=|3944FA122162092002067A3600202567C94D48D00DC47B11213E23383DB51968A705AAFA60C60E263D50CD259D7C9A03FD0C08000002FD0B0011|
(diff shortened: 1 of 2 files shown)


weetmuts:
I think you can close this draft. :-)

PR #693, opened 2022-11-23
	new file:   src/driver_mkradio4a.cc

thisandre:
new file:   src/driver_vario411.cc
Drivers for Techem Radio 4 (warm & cold water) & Vario411 (heat) as used by Techem in Germany.

### DIFF
--- a/src/driver_mkradio4a.cc
+++ b/src/driver_mkradio4a.cc
@@ -1,0 +1,79 @@
+/*
+ Copyright (C) 2017-2022 Fredrik Öhrström (gpl-3.0-or-later)
+ Copyright (C) 2018 David Mallon (gpl-3.0-or-later)
+
+
+ This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


weetmuts:
One of the mfct,type,version combos that you specified was already in use in the weh_07 driver.
@thisandre do you think the drivers should be merged? I believed I picked the weh_07 name because
the combo was used in many different brandings of the same meter.

thisandre:
Frederik, you are right, the mfct specific block contains all other values but at the moment, I'm not able to decode this info.  I asked Techem but did not get a response yet (their responsetime is awfully long - around 1 month).  As such, according to their published information, the values one can get after decrypting the telegram, are the previous year values (value & time).  
I would prefer NOT to merge the drivers weh_07 & mkradio4a in order to extend the mkradio4a when I get more information on how to decode the mfct specific block.

thisandre:
Frederik, test of mkradio4a & vario411 successfull !

PR #698, opened 2022-11-24
Add MQTT discovery info for qheat driver

kroimon:


ha-addon/mqtt_discovery/qheat.json
```diff
@@ -0,0 +1,50 @@
...
+                    "wmbusmeters_{id}"
+                ],
+                "manufacturer": "Qundis",
+                "model": "{driver}",
+                "name": "{name}",
+                "sw_version": "{id}"
+            },
+            "enabled_by_default": true,
+            "json_attributes_topic": "wmbusmeters/{name}",
+            "device_class": "energy",
```
jannickfahlbusch, line 16:
```suggestion
            "device_class": "gas",
```

While technically incorrect I had to change this in order to being able to include my qHeat meter into the Energy dashboard as main source for Heat instead of the Energy section that usually tracks electrical energy.

kroimon: Yeah I was aware of that HA shortcoming, but decided that this is a workaround that anyone could do on their own by changing the default file if needed.
I am not using that for now and instead track heating as a separate device because it looks wrong to me to have my heating listed as gas.

Please support https://community.home-assistant.io/t/energy-dashboard-change-gas-into-heating/391647 to add a separate heating category to the energy dashboard.

weetmuts:
@kroimon @jannickfahlbusch @to-masz I am not an expert on the ha-addon stuff. Should I merge this as is?

kroimon:
I would love to have it merged as is, as changing the device class would be wrong.
For reference, the [HA developer docs](https://developers.home-assistant.io/docs/core/entity/sensor/) say about the 'gas' device class:

> Volume of gas, statistics will be stored in m³. Gas consumption measured as energy in kWh instead of a volume should be classified as energy.

jannickfahlbusch:
I also agree that this can be merged as it is right now. This is something that should be addressed in HomeAssistant. 

While this is a minor annoyance, I've now added this customization configuration which should for now make everything work as I like:
```yaml
homeassistant:
  customize:
    sensor.heat_energy:
      device_class: gas
```

weetmuts:
Great! Thanks for the quick answer.

PR #701, opened 2022-11-25
Add support for a different version of flowiq2200

opajonk:
I apparently have a different version of the Kamstrup flowIQ 2200 (KWM2231). With this little modification, it works flawlessly.

### DIFF
--- a/src/driver_flowiq2200.cc
+++ b/src/driver_flowiq2200.cc
@@ -32,6 +32,7 @@ namespace
         di.addLinkMode(LinkMode::C1);
         // FlowIQ2200
         di.addDetection(MANUFACTURER_KAW,  0x16,  0x3a);
+	di.addDetection(MANUFACTURER_KAW,  0x16,  0x3c);
         // FlowIQ3100
         di.addDetection(MANUFACTURER_KAM,  0x16,  0x1d);
 


weetmuts:
Thanks! But can you provide an example telegram? Often the new version numbers means that there can be a change in the content, that might not be visible in the decoded data, but might be relevant for further driver versions.

opajonk:
I got this when setting `logtelegrams=true`:

```
telegram=|4744372C763081233C168D2056D1ED11205A3C78_04FF2300080000441300000000523B000006FF1B08900008F0FF426CC12B61670A51671B023B000004131F0F00008101E7FF0F13|+149
telegram=|3244372C763081233C168D2057D2ED112072257905095480_0008000000000000000008900008F0FFC12B0A1B00001F0F000013|+165
telegram=|3244372C763081233C168D2058D3ED112072257905095480_0008000000000000000008900008F0FFC12B0A1B00001F0F000013|+180
telegram=|3244372C763081233C168D2059E0ED112072257905095480_0008000000000000000008900008F0FFC12B0A1B00001F0F000013|+196
```

Does that help?

weetmuts:
Thank you for the telegrams! I could add more fields that are inside your telegrams. You can check the content here:
https://wmbusmeters.org/analyze/4744372c763081233c168d2056d1ed11205a3c7804ff2300080000441300000000523B000006ff1B08900008f0ff426cc12B61670a51671B023B000004131f0f00008101e7ff0f13

weetmuts:
If you can give any input on what the error flag mean, then we can update the driver there as well.

opajonk:
> If you can give any input on what the error flag mean, then we can update the driver there as well.

Unfortunately not yet. I know it has some features like leakage detection (the person installing it said that), but I got no user manual or anything beyond the stuff that Kamstrup has on their website...

If I figure out something new about it I will open a ticket / PR!

PR #715, opened 2022-12-03
Add device class water to qwater autodiscovery

jannickfahlbusch:
This enables the Sensor to be used as part of the Home Assistant Energy dashboard.


weetmuts:
Thanks!

PR #718, opened 2022-12-04
Docker update

BIBOLV:
Moving docker building from CircleCI to docker actions. Tested built images on amd64 and arm64. Adding also dependabot checks for used github action script updates.


BIBOLV:
Also to enable dependabot, you need to go to Insights -> Dependency Graph -> Dependabot and enable it.

weetmuts:
Many thanks!

BIBOLV:
Also looks line you will need to remove hook for circleci, or it will show up in commit checks.

weetmuts:
Fixed!

PR #726, opened 2022-12-07
MQTT discovery improvements

kroimon:
Adds missing `device_class` properties and fixes discovery files for new `kamheat` driver.


weetmuts:
Thanks!

PR #729, opened 2022-12-08
Snap update

BIBOLV:
Snap has been upgraded from core18 to core22.
Build process moved from snapcraft to github actions - now built snaps are being pushed to appropriate channels automatically. Tested on amd64 and arm64 with imst and rtl_sdr.
Webhook should be removed in order to prevent parallel builds in github and snapcraft. If login to snapcraft fails during build then store login token should be regenerated.


weetmuts:
Many many thanks! This is great work!

PR #751, opened 2022-12-16
Update DOCS.md

chpego:
Since HA 2022.12, it is necessary to declare the sensors in mqtt in yaml (see [here](https://www.home-assistant.io/blog/2022/12/07/release-202212/#breaking-changes))

Change the unit_of_measurement (available [here](https://developers.home-assistant.io/docs/core/entity/sensor/) )

Add the state_class option with **total_increasing** value (necessary to be able to add in energy dashboard)

### DIFF
--- a/ha-addon/DOCS.md
+++ b/ha-addon/DOCS.md
@@ -66,20 +66,25 @@ mqtt:
 
 ## Home Assistant integration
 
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks!

PR #752, opened 2022-12-16
Add ports to forward for netcat to ha-addon. User can use rtlwmbus:CMD as device

SzczepanLeon:
Added ports to ha-addon (UDP and TCP) to allow use of:
`device=rtlwmbus:CMD(/usr/bin/nc -lku 9011)`
or
`device=rtlwmbus:CMD(/usr/bin/nc -lk 9022)`

### DIFF
--- a/ha-addon/config.json
+++ b/ha-addon/config.json
@@ -11,6 +11,14 @@
     "devices": ["/dev/ttyUSB0", "/dev/ttyAMA0"],
     "usb": true,
     "uart": true,
+    "ports": {
+      "9011/udp": null,
+      "9022/tcp": null
+    },
+    "ports_description": {
+      "9011/udp": "Netcat udp-listen port",
+      "9022/tcp": "Netcat tcp-listen port"
+    },
     "services": [
       "mqtt:need"
     ],


weetmuts:
Thanks!

weetmuts:
Hi @SzczepanLeon ! 
I have managed to start a HA inside a kvm on Linux. Added the latest wmbusmeters as addon (with your fix)
change the config to use loglevel=debug and device=rtlwmbus:CMD(/usr/bin/nc -lk 9022)

Then I first change into the hass vm: virsh console hass
login as root
Then I move into the docker for wmbusmeters:
docker exec -it 7cd7ad1c9f67 /bin/bash
Then I send a telegram to wmbusmeters:
echo "T1;1;1;2019-04-03 19:00:42.000;97;148;33225544;0x1844AE4C4455223368077A55000000041389E20100023B0000"  | nc -q1 localhost 9022

And voila! The HA wmbusmeters log prints:
Received telegram from: 33225544
          manufacturer: (SEN) Sensus Metering Systems, Germany (0x4cae)
                  type: Water meter (0x07)
                   ver: 0x68
                device: rtlwmbus[cmd_0]
                  rssi: 97 dBm
                driver: iperl

Alas, I cannot send to the 9022 port outside of this inner docker.
The hass vm is very stripped down and has neither nc nor telnet, as far as I can see?

The hass vm exports a web interface on 
http://192.168.122.89:8123/

So I was hoping that 192.168.122.89 should offer the 9022 port as well.
But apparently not.... connection refused. Do I need to do some more?

Many thanks!

weetmuts:
It just happens to work to use loglevel=debug and -lkv for netcat to get some extra debug info in the log.

weetmuts:
I have also added support for piping hex or raw bytes using a command, ie these are also possible:
device=hex:CMD(/usr/bin/nc -lk 9022)
device=rawtty:CMD(/usr/bin/nc -lk 9022)

SzczepanLeon:
Please check my short [HowTo](https://github.com/SzczepanLeon/esphome-components/blob/main/docs/wmbusgw.md):

weetmuts:
A very nice! By enabling the hidden disabled ports it works nicely! Thanks!

weetmuts:
But is the mqtt config always the one you specify?
mqtt:  
  host: 10.1.2.44  
  port: "1883"  
  user: myUser  
  password: myPass

Or can I find the mqtt setting somewhere in my hass?

SzczepanLeon:
This is only example of mqtt. In that part You should put address, user, etc. mtqq broker that You want to use. It can run on separate machine or as HA addon.

weetmuts:
Finally! I started my own mosquitto service in ubuntu, outside of the hass vm with an ip and username/pwd. Made an integration to it and added the mqtt setting to the wmbusmeters conf  and added the meter manually to the /mnt/data/supervisor/homeassistant/configuration.yaml inside the hass vm. 

Then echo "T1;1;1;2019-04-03 19:00:42.000;97;148;33225544;0x1844AE...." | nc -q1 192.168.122.89 9022

and the sensor showed up in hass! Horray!

weetmuts:
Aha, there is an HA add-on that starts an mqtt server?

weetmuts:
Can I find the received meter values in some log inside HA? 
I want to use this to setup an automatic HA test system, so I cannot look at the UI for checking that the values arrive....

SzczepanLeon:
Addon is here:
https://github.com/home-assistant/addons/blob/master/mosquitto/DOCS.md

weetmuts:
@SzczepanLeon Why do you write that the auto-discovery feature is very limited?
Is it because only a few meters have auto-discovery scripts?

SzczepanLeon:
> Why do you write that the auto-discovery feature is very limited?
> Is it because only a few meters have auto-discovery scripts?

Yes, exacly (at least in Poland, where Apator16-2 and Izar are popular ;-) ). As I see only 13 meters are available.
Maybe it is possible to autogenerate those mqtt discovery json files from driver C++ files, Ff not with full information, then only with "volune" data.

weetmuts:
> Maybe it is possible to autogenerate those mqtt discovery json files from driver C++ file

Indeed! That is my intent.

PR #761, opened 2022-12-21
HA-addon improvement

chpego:
Hi,

Maybe it's useful (or not), i think maybe it's cool to using lib of [bashio](https://github.com/hassio-addons/bashio)

What do you think ?


BIBOLV:
Hi, 
Looks nice, but doesn`t it also requires docker base image change? Currently default one is being used, but is bashio included there?

chpego:
In run.sh file, bashio is already used

weetmuts:
Thanks!

PR #771, opened 2022-12-29
Added Robin device designator 

jacek27:


weetmuts:
Thank you! But do you have an example telegram as well? Sometimes more information is sent in new versions of the telegrams, even though the driver works as is, we could miss the new data.

jacek27:
Hi, 
Please see the example telegram 
3044B05C11012200041B7A2B0000002F2F0A6617020AFB1A100602FD971D00002F2F2F2F2F2F2F2F2F2F2F2F2F2F2FC83FDF
The only change in reference to munia is in the marked area 
3044B05C11012200**04**1B7A2B0000002F2F0A6617020AFB1A100602FD971D00002F2F2F2F2F2F2F2F2F2F2F2F2F2F2FC83FDF

weetmuts:
Great! Then we know for sure! Many thanks for the fix!

weetmuts:
Can you move the test telegram in to the driver itself?

weetmuts:
Thanks again! Merged it and removed the unnecessary # in the driver test case that caused an error.

PR #780, opened 2023-01-02
Update test of fhkvdataiii to new year

george-hopkins:
Update the test because the current year isn't actually transmitted in the telegram.


weetmuts:
Thank you very much! I just rewrote the driver from scratch and noticed the problem and fixed it.
It is a bit silly, the telegram does not contain the year, but the driver just prepends the current year.

If you modify tests/test_drivers.sh so that when it grabs the json from the driver source,
it also translates "__YEAR__" into the current year, eg 2023. Then we can write __YEAR__
in the test json and it will work. :-)

Closing this one though.

PR #781, opened 2023-01-02
Propagate timestamps from rtlwmbus

george-hopkins:
I'd like to post-process some logs from rtlwmbus and run wmbusmeters over them. However, the field "timestamp" does not reflect the date the telegram was received by rtlwmbus but rather when it was processed (which is always the current time). To propogate the timestamp, I added a new field to `AboutTelegram`. The change is backwards-compatible in the sense that other input sources will produce the same results as before.

Please feel free to close this PR if don't agree with the approach. I just went ahead because the patch is quite small.


weetmuts:
Great idea and implementation!
Have you run "make test" ?

Can you add a test case tests/test_rtlwmbus_timestamps.sh
(Use for example test_aes.sh as a template)
Make a new simulation/serial_rtlwmbus_with_timestamps.msg for the test data.

george-hopkins:
Thank you for the review. I added a test to ensure the timestamps are parsed correctly.

PR #783, opened 2023-01-03
Infer current year in fhkvdataiii

george-hopkins:
Thinking about #780 a bit more, I think we can actually infer the current year from the previous year. This might be the reason it is not transmitted in the first place.


weetmuts:
Of course it is! Great find!

PR #791, opened 2023-01-08
Add Diehl IZAR RC 868 I R4 PL water meter to MQTT auto-discovery

convicte:
Hi all,

I've made an attempt to create an auto-discovery file for the relatively popular IZAR driver. Since it's my first commit every, please excuse any obvious issues or breaches of coding etiquette. 

I am awaiting some assistance with an MQTT auto-discovery error (https://github.com/weetmuts/wmbusmeters/issues/550), so the above has not been tested yet, but if someone would like to try it out, it's as simple as creating a .json file in your mqtt_discovery folder.

Thank you!

### DIFF
--- a/ha-addon/mqtt_discovery/izar.json
+++ b/ha-addon/mqtt_discovery/izar.json
@@ -1,0 +1,148 @@
+{
+  "total_m3": {
+    "component": "sensor",
+    "discovery_payload": {
+      "device": {
+        "identifiers": ["wmbusmeters_{id}"],
+        "manufacturer": "Diehl",
+        "model": "Water meter IZAR RC 868 I R4 PL",
+        "name": "{name}",
+        "sw_version": "{id}"
+      },
+      "name": "{name} total",
+      "unique_id": "wmbusmeters_{id}_{attribute}",
+      "state_topic": "wmbusmeters/{name}",
+      "value_template": "{{ value_json.{attribute} }}",
+      "json_attributes_topic": "wmbusmeters/{name}",
+      "icon": "mdi:gauge",
+      "unit_of_measurement": "m³",
+      "state_class": "total_increasing",
+      "device_class": "water",
+      "enabled_by_default": true
+    }
+  },
+
+  "last_month_total_m3": {
+    "component": "sensor",
+    "discovery_payload": {
+      "device": {
+        "identifiers": ["wmbusmeters_{id}"],
+        "manufacturer": "Diehl",
+        "model": "Water meter IZAR RC 868 I R4 PL",
+        "name": "{name}",
+        "sw_version": "{id}"
+      },
+      "name": "{name} last month total",
+      "unique_id": "wmbusmeters_{id}_{attribute}",
+      "state_topic": "wmbusmeters/{name}",
+      "value_template": "{{ value_json.{attribute} }}",
+      "json_attributes_topic": "wmbusmeters/{name}",
+      "icon": "mdi:gauge",
+      "unit_of_measurement": "m³",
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks! I am not an expert on the HA stuff, but it looks like the other auto-discovery scripts! :-)
Please confirm when you have tested it and there is no longer any errors.

convicte:
@weetmuts 

Since @jnxxx assisted in resolving the issue (kudos once again for that), I've been tested the above PR on my HA and it works like a charm:

image


If someone would like to change it to their liking after having it deployed, it's super simple.
There is many ways to skin a cat (or call a sensor in this case) and I think these look simple, nice and align with all the other meter layouts.

Feel free to incorporate in due course.

chpego:
I see some errors on line 116 and 135, about manufactured.
It's inverted with model, right ?

convicte:
@chpego, fair point!

Too much copying and pasting, so I failed to spot the omission.

Corrected per your comment.

chpego:
One more question but the model here should be the value obtained by the **prefix** attribute, right? IZAR RC 868 i W R4 is not really the type of meter but rather the support that is added to the water meter?

convicte:
> One more question but the model here should be the value obtained by the **prefix** attribute, right? IZAR RC 868 i W R4 is not really the type of meter but rather the support that is added to the water meter?

I don't know that this is an established convention.
Personally, I would prefer to know what actual model it is reading and not a prefix number of the unit, being referred to as model. I've also seen this being done for other units, so it's not really a precedence, I believe.
We could pull the prefix as a separate sensor or read it as an attribute somewhere.

chpego:
> Personally, I would prefer to know what actual model it is reading and not a prefix number of the unit, being referred to as model. 
There is several model of **IZAR RC 868 i**, i don't have the same reader of you.

> I've also seen this being done for other units, so it's not really a precedence, I believe.
I think that there are things to review on the other units, for me it is not necessary to use the field MQTT **sw_version** since we do not have this information on the statements

convicte:
@chpego,

I am not sure if I understand what you'd like to do next.

Can you please make the adjustments in the PR or be more specific?

Thank you in advance!

ha-addon/mqtt_discovery/izar.json
```diff
@@ -0,0 +1,148 @@
+{
+  "total_m3": {
+    "component": "sensor",
+    "discovery_payload": {
+      "device": {
+        "identifiers": ["wmbusmeters_{id}"],
+        "manufacturer": "Diehl",
+        "model": "Water meter IZAR RC 868 I R4 PL",
+        "name": "{name}",
+        "sw_version": "{id}"
```
chpego, line 10: it is not necessary to use the field MQTT **sw_version** since we do not have this information on the statements

ha-addon/mqtt_discovery/izar.json
```diff
@@ -0,0 +1,148 @@
...
+
+  "last_month_total_m3": {
+    "component": "sensor",
+    "discovery_payload": {
+      "device": {
+        "identifiers": ["wmbusmeters_{id}"],
+        "manufacturer": "Diehl",
+        "model": "Water meter IZAR RC 868 I R4 PL",
+        "name": "{name}",
+        "sw_version": "{id}"
```
chpego, line 33: it is not necessary to use the field MQTT **sw_version** since we do not have this information on the statements

convicte:
> @convicte, could you please change `sw_version` to `hw_version` in all places in your commit?

Done!

BIBOLV:
Thanks!

PR #800, opened 2023-01-13
Update for registers size and description

shevu:


weetmuts:
Thanks! It seems like we could add some more data to the json output from the driver_apator162.cc now?

shevu:
Yes. I think "life days" could be added for now. This parameter is always in the default frame. 
There is also capacity history but I think we don't need them.
Rest of the parameters are "on demand". f.ex. To have battery level master should send special frame - I don't know how to do it for now but I have it on my TODO list :) 

I also know couple of Error flags which we could find in 0x01 parameter ( f.ex. magnetic field, water leak, low battery etc.) but it will be done by another PR.

Thanks for your awsome work! ;)

PR #804, opened 2023-01-15
Docker fix

BIBOLV:
Fix tags for armhf docker image reported in https://github.com/weetmuts/wmbusmeters/issues/762


weetmuts:
Super!

PR #814, opened 2023-01-19
Update driver_minomess.cc

BIBOLV:
Adding two new meters to minomess driver as agreed in https://github.com/weetmuts/wmbusmeters/issues/810


BIBOLV:
Test cases added

weetmuts:
Thanks!

PR #818, opened 2023-01-21
mqtt discovery sharky774

chpego:


chpego:
@cukoun , what do you think?

cukoun:
@chpego cosmetic things thanks for update :-). Power sensor - do you know what is the exact purpose of this sensor? Actual power usage for which period of time? I was not able to find any description so I added and using just total power sensor.

chpego:
> @chpego cosmetic things thanks for update :-). Power sensor - do you know what is the exact purpose of this sensor? Actual power usage for which period of time? I was not able to find any description so I added and using just total power sensor.

On the manufacturer's document, there is not really more information, but I think that it is the power measured at the time t

cukoun:
Agree. It could be there this sensor also and anyone could use it on its own. I decided to use total energy sensor which is correctly calculated on the energy dashboard for every day.

chpego:
as discussed [on another pull request](https://github.com/weetmuts/wmbusmeters/pull/791#issuecomment-1401060214), I removed the **sw_version** attribute which is not appropriate if you follow [the documentation](https://www.home-assistant.io/integrations/sensor.mqtt/#sw_version)

cukoun:
@chpego did you try to find informations about intervals/days/hours etc. in which sharky774 sending updates through its wireless module defaultly? And if there is option to change its configuration.

chpego:
https://www.diehl.com/cms/files/20211110_InstallationGuide_SHARKY_774_Compact_DE_EN_FR_EN.pdf?download=1

cukoun:
Hmm,
on the documentation is written:

The module transmits every 8 ... 256 s (send period max. 0.1 % 
of duty cycle (min. 8s); variable, depending on protocol length and 
programming)

But from the wmbus logs it seems that real communication interval is slightly different:

image

BIBOLV:
Thanks!

PR #822, opened 2023-01-22
Included Techem vario 3 type 3.2.1 in Readme.md

Boldfor:
Even though https://github.com/weetmuts/wmbusmeters/issues/333 is still open (from what I see only some trivial mapping needs to be done), the meter got included into the readme, since it does already work with some workaround.


weetmuts:
Thanks!

PR #832, opened 2023-02-01
Update driver_qualcosonic.cc

BIBOLV:
Adding new meter version to driver - https://github.com/weetmuts/wmbusmeters/issues/831


weetmuts:
Many thanks!

PR #856, opened 2023-02-15
Fixed the repo is not showing the addon.

thecem:
Fixed the lints and puter the example config to the right place.


BIBOLV:
Thanks!

PR #858, opened 2023-02-15
Update the link to the DOCS

convicte:
@BIBOLV 

Not sure if a direct link is how it should be done or if one can link it to a generic `docs` link, which will always follow the specific `docs` file location, as it moves. I took a stab at fixing it the simplest way for now.


BIBOLV:
Thanks for noticing and fix!

PR #868, opened 2023-02-16
Sensostar storageNr fix

supaklass:
The sensostar heatmeter has 30 storage positions for storing energy consuption values. Every semi-month a new value is written to the storageNr 2 and the following values are pushed down one position. This applies even when the meter in configured for monthly storage. In this case the telegrams contain values from every second storage position, these are (in the first half of the month) storageNr 2, 4, 6, etc. and in the second half storageNr 3, 4, 5, etc. This fix corrects the storageNr handling in the sensostar driver to get a subsequent list of values in the first and second semi-month.


BIBOLV:
Hi, please check code - test cases are chopped of at the end.

BIBOLV:
Fixed issues.

weetmuts:
Thanks for your insight in this meter! I had to develop the new dynamically loadable drivers infrastructure before this could be fixed in the proper way. The driver can now be found here: https://github.com/wmbusmeters/wmbusmeters/blob/master/drivers/src/sensostar.xmq

I had to make a few incompatible changes to the sensostar json and fields. Please check if it looks ok.

dkadioglu:
Thanks for merging this! I just updated and the new driver seems to work, except one thing: I now have the fields enegry_consumption_semi_monthly_2_kwh to enegry_consumption_semi_monthly_28_kwh with only even numbers. As far as I understood, it is ok, if only every second field is transmitted respectively received. However, as the manual of the meter mentions 30 fields, the last field seems to be missing. Could that be due to your decoding implementation or is it normal? How could I test that?
And another small thing: the fields enegry_consumption_semi_monthly_*_kwh all contain a typo in the first word enegry.
However, again, thank you very much @supaklass and @weetmuts!

weetmuts:
Can you post a link to wmbusmeters.org where your telegram is decoded? Then we can take a look at the contents together.

weetmuts:
energy spelling error fixed.

weetmuts:
A link like this, but for your telegram:
https://wmbusmeters.org/analyze/a444c5146025750200047ac20000202f2f046d2e26c62a040643160000041310f0050001fd1700426cbf2c4406570e00008401061f160000840206f6150000840306f5150000840406f3150000840506ea150000840606bf1500008407065214000084080692120000840906c5100000840a06570e0000840b06ca0b0000840c06da090000840d06ca080000840e06c8080000840f06c608000003fd0c05010002fd0b2111

dkadioglu:
Sure, I copied the telegram from the debug log and used the line starting with "(TPL) appended", I hope this was the right one:
https://wmbusmeters.org/analyze/2F2F046D102A14360406FB0600000413D54D010001FD1700426CFF2C4406D6020000C40106FB060000C40206FB060000C40306A8060000C40406CE050000C40506E4040000C40606D6020000C4070619010000C4080604000000C4090600000000C40A0600000000C40B0600000000C40C0600000000C40D0600000000C40E0600000000C40F06000000002F2F2F2F2F03FD0C05010002FD0B0111

weetmuts:
ah that is only part of the telegram. use logtelegrams=true or --logtelegrams to get an easy to find telegram= line.

weetmuts:
I pasted a header in front of your tpl data to make a fake telegram, but it is good enough to look at the contents.
https://wmbusmeters.org/analyze/a944C5144404752100047AC10000202F2F046D102A14360406FB0600000413D54D010001FD1700426CFF2C4406D6020000C40106FB060000C40206FB060000C40306A8060000C40406CE050000C40506E4040000C40606D6020000C4070619010000C4080604000000C4090600000000C40A0600000000C40B0600000000C40C0600000000C40D0600000000C40E0600000000C40F06000000002F2F2F2F2F03FD0C05010002FD0B0111

weetmuts:
I can see that you have a storage nr 31 which is not decoded, this is easily fixed by changing 2,30 to 2,31 in the driver file. Or 2,129 if you think you will have even more fields.

But how are these semi monthly interpreted, which dates do they correspond to?

dkadioglu:
> ah that is only part of the telegram. use logtelegrams=true or --logtelegrams to get an easy to find telegram= line.

Is this better (now, wmbusmeters.org shows output):
https://wmbusmeters.org/analyze/A944C5144473352200047A6D0090252F2F046D152A14360406FB0600000413D54D010001FD1700426CFF2C4406D6020000C40106FB060000C40206FB060000C40306A8060000C40406CE050000C40506E4040000C40606D6020000C4070619010000C4080604000000C4090600000000C40A0600000000C40B0600000000C40C0600000000C40D0600000000C40E0600000000C40F06000000002F2F2F2F2F03FD0C05010002FD0B0111

weetmuts:
Yep, that is working. As you can see all content is decoded except storage 31. Do you understand the meaning  of these semi monthly measurements? Do we count months back from the meter_datetime?

dkadioglu:
> I can see that you have a storage nr 31 which is not decoded, this is easily fixed by changing 2,30 to 2,31 in the driver file. Or 2,129 if you think you will have even more fields.

Thank you for that hint. Changed to 2,31 and put the adapted driver file into /etc/wmbusmeters.drivers.d/sensostarfix.xmq. I also changed the name of the driver to sensostarfix. And I changed the line driver in my meter specific config files to sensostarfix. However, when a telegram is received, I get a segfault:
```
wmbusmeters  | Started config im871a[01001652] on /dev/im871a listening on t1,c1
wmbusmeters  | (meter) WMZunten: meter detection did not match the selected driver ! correct driver is: sensostarfix
wmbusmeters  | (meter) Not printing this warning again for id: 23151989 mfct: (EFE) Engelmann Sensor, Germany (0x14c5) type: Heat meter (0x04) ver: 0x00
wmbusmeters  | Segmentation fault (core dumped)
wmbusmeters exited with code 0
```

weetmuts:
Do you have an answer to this question?

I.e. if target_date is 2023-12-31 (using storage nr 1) then the value using storage nr 3 would be measured on the date 2023-10-31 ?

dkadioglu:
> Do you have an answer to this question?
> 
> I.e. if target_date is 2023-12-31 (using storage nr 1) then the value using storage nr 3 would be measured on the date 2023-10-31 ?

Unfortunately not so far. Still trying to find documentation.

weetmuts:
Ok, I think we should rename the field to "target_3_kwh" for the kwh at storage nr 3.
No subtracting 1 from the storage nr, this only confuses the user. target_1_kwh is just target_kwh.

weetmuts:
Ok, fix pushed. Now try your link to wmbusmeters.org again. Check if it looks ok.
If you pull, build and install the latest wmbusmters you will get this driver as a builtin. No need to have your own xmq.

wmbusmeters now support generating a key name that is calculated not only from storage_nr but from another date field.

For example, you can now write the field name for historical values:
     name     = 'target_{storage_counter}_{target_date - (storage_counter * 1month)}' 
instead of just:
     name     = 'target_{storage_counter}' 

This will generate fields like:
   "target_2_2021-10-31_kwh":5663

But unfortunately we do not exactly know which month storage_nr 2 corresponds to relative the target_date.
It could also be 2021-11-31. 
Also I cannot add this to the driver because some other sensostar meters do not send the target_date!
Only a meter_datetime.

Anyway for other future meters which are more stable, then this is a nifty feature that perhaps i useful.

weetmuts:
To get around the segfault please rebuild from a clean version. There is a lot of changes in the code.
Also do not use your own xmq, try the builtin. Then when this works you can start developing your own xmq.

There is a misfeature that spelling errors in the xmq will trigger a seg-fault, fortunately after it has printed
a good error message describing the problem. :-)

dkadioglu:
It seems to work, thank you very much. Now I have target_kwh and  target_3_kwh to target_31_kwh, so only uneven numbers. As far as I understand the comment https://github.com/wmbusmeters/wmbusmeters/pull/868#issue-1588386781 from @supaklass, at the first of July I should start to see target_2_kwh to target_30_kwh, until the first half of the month ends and the second half starts, where I will then see again target_3_kwh to target_31_kwh. What I don't get is: @supaklass as well as the manual say, that the device has 30 storage spaces. Shouldn't we than have target_1_kwh to target_29_kwh and target_2_kwh to target_30_kwh? And the sixteenth field is maybe the energy consumption exactly at the target date, which would always be transmitted whether it is the first or second half of the month?

dkadioglu:
> It seems to work, thank you very much. Now I have target_kwh and target_3_kwh to target_31_kwh, so only uneven numbers. As far as I understand the comment [#868 (comment)](https://github.com/wmbusmeters/wmbusmeters/pull/868#issue-1588386781) from @supaklass, at the first of July I should start to see target_2_kwh to target_30_kwh, until the first half of the month ends and the second half starts, where I will then see again target_3_kwh to target_31_kwh. What I don't get is: @supaklass as well as the manual say, that the device has 30 storage spaces. Shouldn't we than have target_1_kwh to target_29_kwh and target_2_kwh to target_30_kwh? And the sixteenth field is maybe the energy consumption exactly at the target date, which would always be transmitted whether it is the first or second half of the month?

As expected, starting on first of July I got target_2_kwh to target_30_kwh and the values of target_3_kwh to target_29_kwh are not included anymore - the uneven ones should appear again starting July 15th. When I compare the values transmitted via WMBUS with the values stored in the device, I can see the following: the device stores 30 values, which are stored on the 15th of the month and on the last day of the month. So, if I go through the values today, the first value is from June 30th, the second is from June 15th and so on and so forth. On July 15th all values will be moved one space up, so the first value in the list will then be from July 15th, the second one from June 30th, and so on and so forth. According to the manual, there are two possibilities, how the values could be transmitted, if the device is configured to send long telegrams: 15 monthly values or 30 half-monthly values
(compact mode). In my case the device is configured to send monthly values, so is not configured with compact mode. This means, as far as I understand, that the 15 values I receive, are the respective values of the last day of the previous months, starting with the last month transmitted in storage space 1, the month before the last one stored in storage space 2 and so on and so forth. So, in my configuration I should only get target_1_kwh to target_15_kwh. If the device is configured with compact mode, then I don't know exactly what will be transmitted - as I do not own the device, I cannot reconfigure it.

So, to make it short: For my configuration, the current driver is wrong. Instead, it should interpret the 15 storage spaces as the values of the last 15 months. And if we take it like that, the current naming of the values, target_x_kwh, is maybe not the best one. Instead it should be named like 1_month_ago_kwh, 2_months_ago_kwh and so on and so forth.
And as an improvement, the driver could include the derived months as additional values, e.g. 1_month_ago_date, 2_months_ago_date and so on and so forth.

I hope, this was understandable!? If not, please ask.
Maybe someone else can contribute, how the half-monthly values are transmitted , when the device is configured for long telegrams and compact mode.

Do you think, the driver could be improved to adapt its interpretation according to the configuration of the device, i.e. the user has to tell the driver, if the device is configured to send long telegrams with compact mode enabled or disabled?

If you need further info, please ask. I will try to help.

dkadioglu:
@weetmuts Could you have a look into my previous comment? That would be great! Thank you very much.

PR #867, opened 2023-02-16
Create trigger_ha_addon.yml

BIBOLV:
Action for ha-addon container build trigger in wmbusmeters/wmbusmeters-ha-addon repo. 
This should be accepted after secrets has been configured in all related repos. 
Starting from now all ha-addon changes should go to wmbusmeters/wmbusmeters-ha-addon repo and should not be accepted here. 
After next release I will make changelog changes to inform users to switch to new repo.

PR #869, opened 2023-02-17
 More addon options available

chpego:
According to the [doc](https://github.com/weetmuts/wmbusmeters/blob/master/README.md#running-without-config-files-good-for-experimentation-and-test), added some options optional

### DIFF
--- a/ha-addon/config.json
+++ b/ha-addon/config.json
@@ -56,13 +56,21 @@
           "password": "str?"
         },
         "conf": {
+          "alarmexpectedactivity": "str?",
+          "alarmshell": "str?",
+          "alarmtimeout": "str?",
           "loglevel": "str",
           "device": "str",
           "donotprobe": "str?",
           "logtelegrams": "bool?",
           "format": "str",
           "logfile": "str",
-          "shell": "str"
+          "shell": "str",
+          "meterfiles": "str?",
+          "meterfilesaction": "list(overwrite|append)?",
+          "meterfilesnaming": "list(name|id|name-id)?",
+          "meterfilestimestamp": "list(never|day|hour|minute|micros)?",
+          "ignoreduplicates": "bool?"
         },
         "meters": [
           {


KrzysztofHajdamowicz:
From a quick look at this- looks fantastic!

BIBOLV:
Firstly - let us not create any PR`s for ha-addon in this repo anymore unless it is urgent and critical, otherwise I will need to transfer changes manually - please see https://github.com/weetmuts/wmbusmeters/pull/867
Secondly - I must confess that I`m not that familiar with ha-addon configuration format implementation, but in our case it is restrictive. Is there a way to allow user to add whatever key: value parameters in each sections so we would not need to manually describe them? Like creating open section or multiple `str?: str?` elements?

chpego:
> Firstly - let us not create any PR's for ha-addon in this repo anymore unless it is urgent and critical, otherwise I will need to transfer changes manually - please see #867 Secondly - I must confess that I'm not that familiar with ha-addon configuration format implementation, but in our case it is restrictive. Is there a way to allow user to add whatever key: value parameters in each sections so we would not need to manually describe them? Like creating open section or multiple `str?: str?` elements?

So I close this PR and make a new one on the other repository?
I've looked at a lot of official HA addon and I'm not sure if it was possible.

thecem:
@chpego your solution is the right one. str?: str? Is not possible.

chpego:
I found [this](https://github.com/hassio-addons/addon-grafana/commit/1fd163c7a6c14a5b6ff0a6927f3ebbb869684f24), Does it look good?

BIBOLV:
Yeah, to me more and more looks like only way to achieve what we need is to disable built in config manager and write one on our own and give access to it trough Ingress.
@chpego - you can create one in parallel to second repo when finalized with this one, or in worst case - I will transfer changes manually.

BIBOLV:
This is my idea - https://github.com/BIBOLV/wmbusmeters-ha-addon 
configuration is moved to separate page that is being served over Ingress functionality. Install plugin -> enable "Show in sidebar" -> go to sidebar -> change and save config.
Only quirk is that if multiple elements of configuration parameters are needed, for example
```
donotprobe=/dev/ttyACM2
donotprobe=/dev/ttyUSB0
donotprobe=/dev/ttyUSB1
```
then in web it should be written in one line with `;` delimiter because multiple key=value elements is not valid json that is being used in this case, example `donotprobe=/dev/ttyACM2;/dev/ttyUSB0;/dev/ttyUSB1`
It should be workaround for this PR, for https://github.com/weetmuts/wmbusmeters/pull/880 and other opened issues.
Any feedback would be appreciated.

BIBOLV:
I will cancel this PR as web config interface is available at https://github.com/wmbusmeters/wmbusmeters-ha-addon in edge version. 
Future fixes please push to https://github.com/wmbusmeters/wmbusmeters-ha-addon

PR #880, opened 2023-02-23
Device and dontprobe arrays

juliekoubova:
The last HA addon update broke our use case; we have two dongles, each listening on a different frequency. This PR changes the `device` and `dontprobe` config fields to be be arrays

### DIFF
--- a/ha-addon/config.json
+++ b/ha-addon/config.json
@@ -58,7 +58,7 @@
         "conf": {
           "loglevel": "str",
           "device": [ "str" ],
-          "donotprobe": [ "str" ],
+          "donotprobe": [ "str?" ],
           "logtelegrams": "bool?",
           "format": "str",
           "logfile": "str",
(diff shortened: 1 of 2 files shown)


thecem:
Pls leave the “?” After str (“str?”), since not all systems need a do not probe config.

juliekoubova:
@thecem it is an array now, so you can pass an empty array instead of an empty string. and you probably shouldn't be using an empty string in the array

juliekoubova:
ie., this is fine:
```yaml
donotprobe: []
```
but you probably don't want to do this:
```yaml
donotprobe:
  - ""
```

thecem:
I don’t know if you are right since HA is awaiting now at least one string in the array, with “str?” the array could be empty.

juliekoubova:
OK, I've put it back

BIBOLV:
Thanks for PR, but I will close it as addon is moving to https://github.com/wmbusmeters/wmbusmeters-ha-addon . Currently edge version is usable. Migration is described in documentation. In your case - when multiple device fields are necessary, it must be entered in one line using `;` as delimiter.
If any fixes needed - please push to  https://github.com/wmbusmeters/wmbusmeters-ha-addon

PR #901, opened 2023-03-05
Added github workflow for building Debian package.

petterreinholdtsen:
As github lack Debian workers, the latest ubuntu worker is used.


weetmuts:
Thanks!

PR #902, opened 2023-03-05
Speed up Ubuntu github workflow using eatmydata.

petterreinholdtsen:


weetmuts:
Super cool! Thank you for teaching my about eatmydata.

PR #904, opened 2023-03-05
Make binary deb package suggest rtl-wmbus.

petterreinholdtsen:
The package can be used to receive broadcasted radio messages.


weetmuts:
Thanks!

Issue #917, opened 2023-03-15
Add new Device: DIEHL AERIUS Gas Meter


Stefan8615:
Hi All,

I have a gas meter from DIEHL (Model AERIUS) which is currently not supported:

(meter) Gas: meter detection could not find driver for id: 42093300 mfct: (DME) DIEHL Metering, Germany (0x11a5) type: Gas meter (0x03) ver: 0x30
(meter) please consider opening an issue at https://github.com/wmbusmeters/wmbusmeters/ (meter) to add support for this unknown mfct,media,version combination

What is required to get this meter supported?

Thanks in Advance!

BIBOLV:
In order to log telegrams, set `logtelegrams=true` in configuration and gather at least 10 telegrams from meter. Check that telegrams are not encrypted, it can be verified in logs or by pasting them [here](https://wmbusmeters.org/). If telegram is encrypted you need to get encryption key, decrypt messages and them post them as we can not do anything with encrypted ones.

Stefan8615:
Thanks for your support. 
here are some readings:

telegram=|2E44A5110033094230037A060020052F2F_0C933E842784060A3B00000A5A5901C4016D3B37DF2CCC01933E24032606|+11
telegram=|5E44A5111630727070077AD50050052F2F_0C13314504004C1330010100CC101300000000CC201330010100426CDF2C0B3B00000002FD74AF12025A6100C4016D3B17EE23CC01135240040001FD085102FD170000EB013B0400F02F2F2F2F2F|+22
telegram=|2E44A5110033094230037A060020052F2F_0C933E912784060A3B60040A5A5901C4016D3B37DF2CCC01933E24032606|+31
telegram=|5E44A5111630727070077AD50050052F2F_0C13314504004C1330010100CC101300000000CC201330010100426CDF2C0B3B00000002FD74AF12025A6100C4016D3B17EE23CC01135240040001FD085202FD170000EB013B0400F02F2F2F2F2F|+37
telegram=|5E44A5111630727070077AD50050052F2F_0C13314504004C1330010100CC101300000000CC201330010100426CDF2C0B3B00000002FD74AF12025A6100C4016D3B17EE23CC01135240040001FD085302FD170000EB013B0400F02F2F2F2F2F|+51
[...]

PR #918, opened 2023-03-15
Create driver_aerius.cc

BIBOLV:
Add support for DIEHL AERIUS Gas Meter - https://github.com/wmbusmeters/wmbusmeters/issues/917

### DIFF
--- a/src/driver_aerius.cc
+++ b/src/driver_aerius.cc
@@ -1,0 +1,106 @@
+/*
+ Copyright (C) 2017-2023 Fredrik Öhrström (gpl-3.0-or-later)
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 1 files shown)


weetmuts:
Very nice! You can remove
.set(StorageNr(1))
.set(SubUnitNr(0))

Because the default is to match with tariff/subunit/storage 0.
Then the set first sets 1 and then 3, only the three sticks.
(In the analyze the storagenr is built up succesively over the dif bytes,
without forward looking debug info, it cannot know at the first byte that there will 
be more storage nr bits, this is why it prints storage 1 there and then 3 at the next byte.)

The datetime now has its own unit. Use:
        addNumericFieldWithExtractor(
            "target",
            "Date time when previous billing period ended.",
            DEFAULT_PRINT_PROPERTIES,
            Quantity::PointInTime,
            VifScaling::Auto,
            FieldMatcher::build()
            .set(MeasurementType::Instantaneous)
            .set(VIFRange::DateTime)
            .set(StorageNr(3))
            );

I think we can remove current_ from current_flow and current_temp,
just use flow and temperature.

BIBOLV:
Done

weetmuts:
In the past I had a lot of default fields, however it turned out, when rewriting the drivers that sometimes the default fields were wrong. Positional fields are problematic, since any change in the number of fields will cause significant incompatibilities.

So from now on, I recommend that only put a single, or very few fields as the default fields.
Having few fields is better since then the user is force to explicitly select the desired fields:

```
wmbusmeters --format=fields --selectfields=total_m3,temperature_c 
2E44A5119870659930037A060020052F2F_0C933E842784060A3B00000A5A5901C4016D3B37DF2CCC01933E24032606 
Gas auto ANYID NOKEY
```
which results in this output
```
6842.784;15.9
```

weetmuts:
So I made a tiny change to the driver after the merge! 
Many thanks for the driver!

PR #937, opened 2023-04-02
amiplus - add maximum power consumption field. This is calculated as …

DeDragonSlayer:
…the maximum value of 15-minute consumption averages, reset on 1st of each month. If the calculated max power consumption exceeds the value on the contract with power company, the end user will pay penalty fees.


weetmuts:
Great work! Could you just not add the new value to the default fields?

Historically the default fields (ie --format=fields which prints semicolon separated values) 
were all of the content in the telegrams, however as time went on, it turned out that
the telegrams are often dynamic and changing the content. This creates
a problem for a line like a;b;c;d;e where the ordering matters.
Adding a new field a;b;c;d;f;e will break backwards compatibility.

In new drivers I try to keep the default fields to the absolute minimum, i.e. just total consumption for an electricity meter.
This will force users to use --selectfields=a,b,c to pick the fields they want in the --format=fields output and this is much better.

--format=json does not have this problem since new fields can be added at any time.

DeDragonSlayer:
Thanks for the detailed explanation. I wasn't really aware of the default field list functionality and just carpet bombed everything with the new field. :) Should be OK now!

weetmuts:
Super! Many thanks!

Issue #898, opened 2023-03-05
Add driver for Sontex Supercal 739


raknjane:
### Type of request

New device support

### Meter description

Hi,

I am using a AMB8465 USB stick trying to read data from a Sontex Supercal 739 (wM-Bus EN13757-4) meter via (wmbus).

This is all new to me but I have managed to retrieve data.
I followed the HowToAddaNewMeter.txt which Simulation suggested to use "Best driver  : microclima 25/31" 

I did a test using the "supercom587" driver having the same result.

This driver does not appear to work correctly and below message show up in the debug log: 

```
(meter) B1102UFH: meter detection did not match the selected driver microclima! correct driver is: unknown!
(meter) Not printing this warning again for id: 12345678 mfct: (SON) Sontex, Switzerland (0x4dee) type: Heat meter (0x04) ver: 0x1b
(meter) please consider opening an issue at https://github.com/weetmuts/wmbusmeters/
(meter) to add support for this unknown mfct,media,version combination
```

Please find details below, not sure if more details is needed if so let me know.

Thank you very much.

https://sontex.ch/en/applications/building-automation/#supercal-739

### Logged telegrams from meter in decrypted format

```shell
Received telegram from: 12345678
          manufacturer: (SON) Sontex, Switzerland (0x4dee)
                  type: Heat meter (0x04)
                   ver: 0x1b
                device: amb8465[00000000]
                  rssi: -74 dBm
                driver: unknown!
telegram=|_4444EE4D850166281B047AFB000000046D3612E423820A6CE1F1040628010000840A060000000004144A110000840A14000000000259E30F025D2A0C023B1D01032C3A0100|+6

wmbusmeters --analyze simulation_12345678.txt
Auto driver  : not found!
Best driver  : microclima 25/31
Using driver : microclima 25/31
000   : 44 length (68 bytes)
001   : 44 dll-c (from meter SND_NR)
002   : ee4d dll-mfct (SON)
004   : 85016628 dll-id (12345678)
008   : 1b dll-version
009   : 04 dll-type (Heat meter)
010   : 7a tpl-ci-field (EN 13757-3 Application Layer (short tplh))
011   : fb tpl-acc-field
012   : 00 tpl-sts-field (OK)(OK)
013   : 0000 tpl-cfg 0000 ( )
015   : 04 dif (32 Bit Integer/Binary Instantaneous value)
016   : 6D vif (Date and time type)
[...]

PR #949, opened 2023-04-22
Update driver_microclima.cc

BIBOLV:
Adding Sontex Supercal 739 - https://github.com/wmbusmeters/wmbusmeters/issues/898


BIBOLV:
Mostly all devices transmit encrypted telegrams, so decryption key is needed. When decryption is successful, then decoding can be done and that is where drivers are needed.

MarekVigas:
I found that in Slovakia Sontex Supercal 739 are white labeled and sold as Enbra Supercal 739. I called Enbra and asked for decryption key for my device. I was told that they don't have unique key for each device. They have agreement with Sontex to flash all Enbra devices with the same key. At least they are willing to share the key but I would have to sign an NDA to get it.

Issue #955, opened 2023-05-09
Engelmann Heat cost allocators HCA e2


Skysurfer-14:
### Type of request

New device support

### Meter description

**Manufacturer**: Engelmann [https://www.engelmann.de/en/hca-e2/](url)
**Model**: HCA e2
**Radio mode**: C-mode OMS 4 Mode 5 with own AES key, Radio time: 7 am - 7 pm daily

For more information see #820 

### Logged telegrams from meter in decrypted format

```shell
(serial) received binary "A5C203CE44C5149126412331087A6B0BC02562235F3EA78C0A58BBB4284B7A9AEBE11943EB52740634A29C9E08CD2C3D90717360D2E342A39DACE7A849D03DCA80F4C424C0DE1A79A2A40F0EAC886C9417FA084BDFB8BE21A8E904AA41946E51D1B5E2D929127C6D1470EAF093497B7574CFB22FF13A866812FD780D9EBE9EB4DE2E8CADEF92F01FBBFFD42B1D7FE0FD8093D76F6015029EF9F6F96E11C4094C79D0CB547DAF69FD7F7CC98F57E6CB4CF147ADD5BAF247E03EF7AFE51EAF58AC905E756C17B077E58FA2E7B3C359BA8867A2938934"
(im871a) checkIM871AFrame "A5C203CE44C5149126412331087A6B0BC02562235F3EA78C0A58BBB4284B7A9AEBE11943EB52740634A29C9E08CD2C3D90717360D2E342A39DACE7A849D03DCA80F4C424C0DE1A79A2A40F0EAC886C9417FA084BDFB8BE21A8E904AA41946E51D1B5E2D929127C6D1470EAF093497B7574CFB22FF13A866812FD780D9EBE9EB4DE2E8CADEF92F01FBBFFD42B1D7FE0FD8093D76F6015029EF9F6F96E11C4094C79D0CB547DAF69FD7F7CC98F57E6CB4CF147ADD5BAF247E03EF7AFE51EAF58AC905E756C17B077E58FA2E7B3C359BA8867A2938934"
(im871a) has_timestamp=0 has_rssi=1 has_crc16=1
(im871a) endpoint 2
(im871a) msgid 3
(im871a) rssi 147 (-49 dBm)
(im871a) got crc16 3489 expected 3489
(im871a) received full frame
(wmbus) parseDLL @0 207
(wmbus) parseELL @10 197
(wmbus) parseNWL @10 197
[...]

Issue #820, opened 2023-01-21
Telegram Decryption for Engelmann Heat cost allocators HCA E2 failed


GitKunde:
Hi,

I'm using heat cost allocators "HCA E2" from Engelmann Sensors (Manufacturer ID = 14c5). I can receive telegrams (see example below) but I'm struggling to decrypt the data. I'm pretty sure to use the right key (standard key). So I tried to analyze the telegram at you nice web page (https://wmbusmeters.org) and got same result (failed decryption. Wrong key?).
Are there any hints to find my mistake? Or is there anything special to regard with these type of sensors?

Thanks for your great work.

Example: cf44c5145041219029087a4800c025b1ca6bee1bba1e4b791ae0a22869dc66b523192799bc8a8a5a866a87a76450bd978f9a84db9aea4ac67c82f708784e16804eb783a38bc481653f002eed922a21c84ed8e2dc228af45b49822abb382ba798db0a9ea0eb40b5382b682a8e40466a5224137555860d35b56bfa726b17fcf6cb1f2f00ff174974f007bf440a27b6811cd20428c2f039248a8e2abbfec3c3908f580e36519ef1b66ac63f59ac8c5fbe2966ac18ad8b953ff51679db9f78e06c2e0b1d7931c8b6e0c219808adafe4532e816

AES-128 Standard-Key:
00112233445566778899aabbccddeeff

IV:
c5145041219029084848484848484848

JensHoRi:
I have ordered Engelmann 2 years ago, no one of them I was able to get running with different software. 

But I got a device specific key for everyone of them.

I ordered somewhere else Lorenz then. Got individual keys, everything worked out of the box after weemut (THANKS) added some additional changes for further values :-)

weetmuts:
How did you receive these telegrams? It is a bit odd that when wmbusmeters tries to decrypt the telegram
the number of encrypted blocks  add up to 192 which leaves 2 unencrypted bytes at the end.
This is sort of allowed, but usually there is more bytes at the end, with dif+vif+data codes....

Could the telegram have some dll-crc bytes not yet removed?

GitKunde:
Thanks for your comments!
[...]

PR #956, opened 2023-05-09
Add files via upload

BIBOLV:
Uploading file created in https://github.com/wmbusmeters/wmbusmeters/issues/820 - related to https://github.com/wmbusmeters/wmbusmeters/issues/955

### DIFF
--- a/src/driver_hcae2.cc
+++ b/src/driver_hcae2.cc
@@ -1,0 +1,143 @@
+/*
+ Copyright (C) 2019-2023 Fredrik Öhrström (gpl-3.0-or-later)
+
+ This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


BIBOLV:
Regarding deprecated fields - is that correct?

weetmuts:
For a new driver you do not need to include the deprecated field error_flags.

A field is marked as deprected if you are modifying an existing driver and want to move towards a more standardized naming of the fields but keep backwardscompatibility for some time. Exactly how the deprecated warning should
be transferred to the ha-addon is not clear. :-) yet. 

Here the future proof standardized field is "status":"OK" if no error flags are set neither inside the difvif or inside the tpl_sts bits etc. The old standard was "error_flags":"".

Issue #965, opened 2023-05-20
Incorrect handling of negative values


demetz:
### Type of request

Bug

### OS version

Linux

### wmbusmeters version

1.13.1

### Your message goes here

Negative values are incorrectly parsed.

Example: 
https://wmbusmeters.org/analyze/3b44a7320556637104047afa2000202f2f0c06774202000c14399956000b2d0200f00b3b3018000a5a51030a5e520302fd170000066d0c080af42500

is parsed as
```
029   : 0B dif (6 digit BCD Instantaneous value)
030   : 2D vif (Power 10² W)
031 C!: 0200F0 ("power_kw":220000.2)
```
however, this should be `-200W`. This value can be negative if volume flow is actually reversed or return temperature > supply temperature.

I've seen this for both `power_kw` as well as for `flow_m3h`.

demetz:
I believe this is linked to https://github.com/wmbusmeters/wmbusmeters/pull/480

weetmuts:
Yes, there is a simple fix for the values that can be negative, replace:

 VifScaling::Auto
with 
 VifScaling::AutoSigned

Do you want to do a pull request?

weetmuts:
Thanks! I discovered some error flags and status bits in your telegram so I added the status field to the ultraheat driver.  Code pushed.

Do you have a manual for your meter?  Currently your tpl status has a mfct specific bit set.
We can add better translation of the bit, which is currently reported as UNKNOWN_20

demetz:
I've requested them from the manufacturer...

pvagner:
Hello,
I have got Lansen [LAN-WMBUS-CX-TH](https://www.lansensystems.com/assortment/sensors/air-quality/lan-wmbus-cx-th/) temperature and humidity sensor.
The wmbus frames are encrypted with a key. When reading negative values these are reported as very high numbers.
See the analyze output below.
Can you please advice how to receive correctly formatted negative temperature values in the output please?
I have rebuilt the app with the recent patch.

```
./build/wmbusmeters --analyze=<key_redacted> 2e44333041000600091b7aa700202585a14580592a8ac1413733152d98c9257c74660f5831bc7771136765b81575c7 
Auto driver  : not found!
Best driver  : lansenth 09/11
Using driver : lansenth 09/11
000   : 2e length (46 bytes)
001   : 44 dll-c (from meter SND_NR)
002   : 3330 dll-mfct (LAS)
004   : 41000600 dll-id (00060041)
[...]

PR #966, opened 2023-05-20
consider negative values for flow and power

demetz:
flow can be negative

power can also be negative, for example if return temperature > supply temperature or flow is reversed


demetz:
related to https://github.com/wmbusmeters/wmbusmeters/issues/965

weetmuts:
Thanks!

PR #968, opened 2023-05-22
lansenth driver: parse temperature as AutoSigned in order to see negative values

pvagner:
As discussed in #965 .
For me it's working fine over here.


weetmuts:
Super! Can your provide a un-encrypted test telegram, for example using --logtelegrams ?

weetmuts:
Many thanks! I just fixed the test case and added the auto-detect for this new model.

PR #974, opened 2023-06-05
Stop installation script on error

travnick:
Just a little improvement to installation script
* replace `cp` by `install` with mode `755` in `install_binaries.sh`, so make sure that everyone is able to execute the `wmbusmeters`
* stop `./scripts/install_binaries.sh` on failure
* stop `./install.sh` on failure

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -44,19 +44,22 @@ do
         esac
 done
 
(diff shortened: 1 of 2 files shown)


weetmuts:
Thank you!

Issue #737, opened 2022-12-12
resetafter - not working


testuser7:
I have a problem with the rtl usb dongle (Astrometa DVB-T2). It stops working after some time.
The resetafter option does not work:
```
/wmbusmeters # ./wmbusmeters --resetafter=10s --donotprobe=all --debug "rtlwmbus:CMD(rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a)" MyWater izar '*' NOKEY
(wmbusmeters) version: 1.10.1
(config) using device: rtlwmbus:CMD(rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a) 
(config) number of meters: 1
(serial) expecting devices to work
(serial) expecting devices to work
(lookup) with cmd "rtlwmbus:CMD(rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a)"
(main) opening rtlwmbus:CMD(rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a)
Started config rtlwmbus listening on none using CMD(rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a)
(rtlwmbus) using command: rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a
(rtlwmbus) opening cmd_0
(bgshell) exec background "/bin/sh"
(bgshell) arg "-c"
(bgshell) arg "rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a"
(serial) EVENT thread interrupted
(serialcmd) opened /bin/sh pid 868 fd 3 (rtlwmbus)
(main) regular reset of rtlwmbus  using CMD(rtl_sdr -f 868.95M -s 1.6e6 - 2>/dev/null | rtl_wmbus -p s -a) will happen every 10 seconds
(wmbus) no alarm (expected activity) for rtlwmbus
(serial) registered regular callback HOT_PLUG_DETECTOR(0) every 2 seconds
(serial) waiting for stop
```
Output from rtl_sdr:
```
/wmbusmeters # rtl_sdr -f 868.95M -s 1.6e6 -
Found 1 device(s):
  0:  astrometadvbt2, dvbt2, SN: 

[...]

PR #991, opened 2023-07-07
Add rtl_reset to docker image

testuser7:
The rtl_reset application automatically resets rtl-sdr devices. rtl-sdr devices hang very often. We can use it like this:
`rtl433:CMD(rtl_433 -F csv -f 868.95M || rtl_reset)`

Issue:  #737

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,13 +2,16 @@ FROM multiarch/alpine:${TARGETARCH}${TARGETVARIANT}-latest-stable AS build
 RUN apk add --no-cache alpine-sdk gcc linux-headers librtlsdr-dev cmake libusb-dev bash
 RUN git clone https://github.com/wmbusmeters/wmbusmeters.git && \
     git clone https://github.com/weetmuts/rtl-wmbus.git && \
-    git clone https://github.com/merbanan/rtl_433.git
+    git clone https://github.com/merbanan/rtl_433.git && \
+    git clone https://github.com/ED6E0F17/rtl_reset.git
 WORKDIR /wmbusmeters
 RUN make
 WORKDIR /rtl-wmbus
 RUN make release && chmod 755 build/rtl_wmbus
 WORKDIR /rtl_433
 RUN mkdir build && cd build && cmake ../ && make
+WORKDIR /rtl_reset
+RUN make
 
 FROM multiarch/alpine:${TARGETARCH}${TARGETVARIANT}-latest-stable as scratch
 ENV QEMU_EXECVE=1
@@ -17,6 +20,7 @@ WORKDIR /wmbusmeters
 COPY --from=build /wmbusmeters/build/wmbusmeters /wmbusmeters/wmbusmeters
 COPY --from=build /rtl-wmbus/build/rtl_wmbus /usr/bin/rtl_wmbus
 COPY --from=build /rtl_433/build/src/rtl_433 /usr/bin/rtl_433
+COPY --from=build /rtl_reset/rtl_reset /usr/bin/rtl_reset
 COPY --from=build /wmbusmeters/docker/docker-entrypoint.sh /wmbusmeters/docker-entrypoint.sh
 VOLUME /wmbusmeters_data/
 CMD ["sh", "/wmbusmeters/docker-entrypoint.sh"]


BIBOLV:
Yes, looks good. I will need to replicate this to snap and ha-addon also.

weetmuts:
I could have wmbusmeters check if it can find the rtl_reset binary and use it when it performs the regular --resetafter=23h since it seems to be a somewhat recurring problem....
what do you think?

weetmuts:
Here is another report.
https://raspberrypiandstuff.wordpress.com/2017/08/04/rtl_433-on-a-raspberry-pi-made-bulletproof/

Now looking at the actual reset, perhaps we do not need to include a separate program. 
It could be done from within wmbusmeters since it is either:
rc = ioctl(fd, USBDEVFS_RESET, 0); (from the usbreset.c source)
or
libusb_reset_device( devh ); (from the rtl_reset source)

both which could be done within wmbusmeters without the external rtl_reset program which adds
convenience functions to find the rtl_sdr device.

Perhaps this usbreset is what I was looking for when dealing with the malfunctioning im871a/amb8465 dongles as well....

weetmuts:
"Edited: Or maybe you want to add it to main tree and get rid of dependency? Codebase is small so it would not make huge difference in maintenance?"

Yes, I agree! :-)

testuser7:
It would be great if such functionality was implemented directly in wmbusmeters.
There's only problem I mentioned in the issue [#737 ](https://github.com/wmbusmeters/wmbusmeters/issues/737#issuecomment-1402906157) that rtl_sdr doesn't detect dongle hangup. Only rtl_433 exits with error code 3 when dongle hangs.

weetmuts:
@testuser7 Yes, I have added a proposed flag -f to rtl_wmbus which will make it detect if the data flow stops.
https://github.com/weetmuts/rtl-wmbus/tree/AddFlowCheck

And I have added a fix to wmbusmeters to detect this error, and capture the stder from  stderr.
https://github.com/wmbusmeters/wmbusmeters/pull/1015

The new code will trigger a reset of the rtlsdr/rtlwmbus pipeline.
see the code:
           if (it != end)
            {
                debug("(rtlwmbus) detected rtl_wmbus exit due to stopped data flow. Resetting pipeline!\n");
                reset();
            }

This seems to be working. Now I only have add the usb device reset code as well.

But perhaps I can detect some stderr output from rtl_sdr to trigger the usb reset instead of a normal restart?
I do not want to reset the usb device infinitum unnecessary, or may it does not matter?
Is the detection of lack of flow enough to consider it to be time for a usb device reset?

testuser7:
> But perhaps I can detect some stderr output from rtl_sdr to trigger the usb reset instead of a normal restart?
> I do not want to reset the usb device infinitum unnecessary, or may it does not matter?
> Is the detection of lack of flow enough to consider it to be time for a usb device reset?

In rtl_433, hang is detected if there are no frames for 250ms: https://github.com/merbanan/rtl_433/blob/master/src/rtl_433.c#L1502

Is using rtl_sdr necessary for some reason? We can replace them with rtl_433 like this:
`rtl_433 -R 0 -f 868.95M -s 1600k -w - | rtl_wmbus -p s -a`
or with pre-decoding:
`rtl_433 -R 0 -X n=wmbus,m=FSK_PCM,s=10,l=10,r=500 -Y squelch -f 868.95M -s 1.6m -F null -w - | rtl_wmbus -p s -a`

weetmuts:
@testuser7 This is very clever! I have tried rtl_433 and your solution works fine! 

However rtl_433 uses 4% cpu where rtl_sdr uses 1% cpu on my laptop. Probably because rtl_433 
does some analysis anyway. This could probably be fixed. 

The problem with wmbusmeters not being able to capture stderr from rtl_sdr was due to a bug
in rtl_sdr that has now been fixed! 

https://github.com/osmocom/rtl-sdr/commit/142325a93c6ad70f851f43434acfdf75e12dfe03

So, as soon as rtl_sdr deb package gets this fix, then we can rely on it for wmbusmeters.

testuser7:
> However rtl_433 uses 4% cpu where rtl_sdr uses 1% cpu on my laptop. Probably because rtl_433
> does some analysis anyway. This could probably be fixed.

My cpu usage tests (inside docker):
1. `rtl_sdr -f 868.95M -s 1.6m - 2>/dev/null | rtl_wmbus -p s -a`
   - rtl_sdr: 0.1%
   - rtl_wmbus: 2.5%
2. `rtl_433 -R 0 -Y squelch -f 868.95M -s 1.6m -w - | rtl_wmbus -p s -a`
   - rtl_433: 0.7%
   - rtl_wmbus: 2.5%
3. `rtl_433 -R 0 -X n=wmbus,m=FSK_PCM,s=10,l=10,r=500 -Y squelch -f 868.95M -s 1.6m -F null -w - | rtl_wmbus -p s -a`
   - rtl_433: 1.3%
   - rtl_wmbus: 2.5%
4. `rtl_433 -R 104 -Y squelch -f 868.95M -s 1.6m -F csv`
   - rtl_433: 1.3%

It's best to use standalone rtl_433.

> The problem with wmbusmeters not being able to capture stderr from rtl_sdr was due to a bug in rtl_sdr that has now been fixed!
> 
> [osmocom/rtl-sdr@142325a](https://github.com/osmocom/rtl-sdr/commit/142325a93c6ad70f851f43434acfdf75e12dfe03)
> 
> So, as soon as rtl_sdr deb package gets this fix, then we can rely on it for wmbusmeters.

I don't think that will ever happen. The alpine package uses a different repository: https://git.alpinelinux.org/aports/tree/main/librtlsdr/APKBUILD

weetmuts:
@testuser7 rtl_433 is a brilliant piece of software designed to receive multiple protocols and frame formats and encodings at once. But because of this it cannot be as sensitive as rtl_wmbus which is tailored for wmbus only.
The rtl_wmbus software is significantly more sensitive than rtl_433, at least in my tests.

It is good that both options exists though.

Ah, yes, for docker, we could build rtl_sdr from source, just as we build rtl_433, rtl_wmbus and wmbusmeters.
@BIBOLV

BIBOLV:
Yes, sure, I can rewrite docker to build rtl_sdr from source, but looking at last answers in issue - are you sure it would help? If I could replicate issue, I could look deeper and try to find issue, but how useful info from someone remotely is a challenge. 
Is there a way to set device parameter that would allow rtl_sdr to write output to file? That way we could check if rtl still provides some output and check if wmbus frames are there.
Or I could ask someone with issue to clone his drive, upload somewhere and I could check at my end if with that I can replicate issue. 
Those are my only ideas currently.

weetmuts:
You are right Bibo, building the rtl_sdr from the latest src does not immediately help with the problem of a hung rtlsdr dongle. However it enables us to track the error message from rtl_sdr. To properly detect if the flow of data has stopped we can do:

1) Use rtl_433 to generate the stream that is piped into rtl_wmbus. rtl_433 already detects lack of flow of data and exits. But it rtl433 uses 4% cpu whereas rtlsdr uses 1% cpu (though this can probably be fixed).

2) Use rtl_433 for wmbus decoding (no need for rtl_wmbus) but rtl_wmbus is significantly better at detection/decoding the radio signals than rtl_433, ie higher sensitivity, longer distance.

3) Use rtl_sdr (with or without workaround for stderr)+ rtl_wmbus(with flow detection) to exit when flow of data 
Will exit, but we might miss out on debug information from rtl_sdr that indicates wheter we should do an usb-reset. Or may the assumption is that if the flow has stopped (and the dongle is still plugged in) then we should usb-reset. 

4) Use rtl_sdr(with stderr fix) + rtl_wmbus(with flow detection)
Same as 3, but nicer command line to get the stderr data from rtl_sdr.

 I have pinged the deb maintainer for rtl_sdr that it would be nice with a new package update with the bug fix. But that might delay for some time. Also for docker it is a different maintainer for the alpine packages, which I have not yet emailed.

I would prefer one of these two solutions:

A) do workaround for  rtl_sdr with the /tmp/file storage for the stderr, until the bug fix is stably in all repos (deb alpone rpm etc) which might take a year. 
Then use the -f option introduced in the weetmuts/rtl_wmbus fork to abort if the flow of data stops.
Then check if two continous restarts of rtl_wmbus has failed with stop of flow of data, then do an usb_reset.

B) use rtl_433 to pipe the flow of data to rtl_wmbus. fix the cpu problem in rtl_433 when just streaming data.
rtl_433 will abort by itself if the flow of data stops. No need to use the -f weetmuts/rtl_wmbus option.
Then check if two continous restarts of rtl_wmbus has failed with stop of flow of data, then do an usb_reset.

testuser7:
There are at least 3 repositories where rtl-sdr is being developed:
- https://github.com/steve-m/librtlsdr (same as osmocom)
- https://github.com/librtlsdr/librtlsdr
- https://github.com/old-dab/rtlsdr

Merging them all would be the best solution. There is even a request for help with this: https://github.com/librtlsdr/librtlsdr/issues/120

PR #1045, opened 2023-10-07
Update README.md with paragraph about SDR naming (swradio0...255)

geraldhuber:


weetmuts:
Thanks! I also clarified that wmbusmeters uses librtlsdr to probe the the SDR devices.

PR #1090, opened 2023-11-11
Added cold water meter type to itron driver

Andi252:
The pull requests adds a new type to the itron driver. Type 0x16 is a cold water meter.


weetmuts:
Many thanks! You do not happen to know what the two manufacturer specific values are? 
They are translated to woot:s right now. :-)

Andi252:
I don't know what the meanings of the additional fields are. It might be some fields which are filled beginning of next year, because the billing period ends on the 31th of December and all meter were installed in October this year.

I will monitor the fields and let you know if I see any changes in the future.

I checked on all the water meter I have access to and all of them have the same values there.

|3A4497269220362300077AAD0020A52F2F_0413A8000000066D131814**EC2B007413FEFEFEFE426C1F01047F1400060C027F9A2A0E79187103002300**|

The part starting form EC2B00 is always the same.

Issue #1054, opened 2023-10-14
Device Support Itron EquaScan h-Koax RF I


huhndomi:
### Type of request

New device support

### Meter description

I would like to read four Itron EquaScan h-Coax RF I heat meters using wmbusmeters. These are not recognized in the current version 1.14.0. These heat meters are also sold in identical form by Engelmann in the SensoStar series. I assume that Engelmann is the manufacturer. I have attached the manufacturer's data sheets in the appendix.

I read the data using a Raspberry Pi3 and a nano CUL 868 MBUS with the firmware nanoCUL_r568_mbus_c1t1_bufsize300.hex

Could it be that the CUL buffer is too small for the telegram, since the following message appears before it?
(dvparser) warning: unexpected end of data
(dvparser) warning: unexpected end of data (vif expected)

According to the Engelmann instructions, the following information should be transmitted:
o Energy (heat/cold energy,
Pulse input 1, pulse input 2,
Pulse input 3)
o Total volume
o Flow
o Performance
o Hint code
o Return temperature
o Temperature difference

[P0413 EquaScan hKoaxRF_D_0720_II.pdf](https://github.com/wmbusmeters/wmbusmeters/files/12908106/P0413.EquaScan.hKoaxRF_D_0720_II.pdf)
[1080619044_2020_07_03_BA_S3_Itron_de.pdf](https://github.com/wmbusmeters/wmbusmeters/files/12908107/1080619044_2020_07_03_BA_S3_Itron_de.pdf)

### Logged telegrams from meter in decrypted format

```shell
(wmbusmeters) logging started 2023-10-14 23:13:26 1.14.0-32-g6e3bac9
(wmbusmeters) version: 1.14.0-32-g6e3bac9
(config) store meter files in: "/var/lib/wmbusmeters/meter_readings"
(config) using device: /dev/ttyUSB0:cul:c1 
(config) number of meters: 0
(lookup) with file/hex "/dev/ttyUSB0" cul
(serial) check if /dev/ttyUSB0 can be accessed
(serial) tty /dev/ttyUSB0 can be accessed
(serial) before config: /dev/ttyUSB0 speed(B38400) input() output() control(CLOCAL|CREAD|CSIZE) local() special_chars() modem(DTR|RTS)
(serial) after config:  /dev/ttyUSB0 speed(B38400) input() output() control(CLOCAL|CREAD|CSIZE) local() special_chars() modem(DTR|RTS)
[...]

PR #1091, opened 2023-11-11
Add new driver for itron heat meter

Andi252:
The pull request adds a new driver for an itron heat meter. See discussion #1081.

The meter is called [UltraMaXX](https://www.itron.com/emea/solutions/product-catalog/cf-ultramaxx-mk)
and the RF module is called [EquaScan hMIU RF](https://www.itron.com/emea/solutions/product-catalog/equascan-hmiu-rf)

was extended to #1054


Andi252:
How are the defines for dd-mfct fields created? My meter produces the id 9726 (ITW) and the meter from #1054 creates the id 97a6. I could add this second meter into the driver, but I don't know how to add the second id.

Andi252:
The Itron-Meter from #1054 produces a dd-mfct of 97a6 which decodes into MANFCODE('i','T','W'), when I add this MANFCODE('i','T','W') to my driver the frame from #1054 is decoded and detected correctly.

@weetmuts do you know of other meters where the MANFCODE is not produced from upper case charaters?
Should I add a funktion with "normalizes" the dd-mfct before matching?

Andi252:
I tried to find a definition of the mfct in the offical OMS specification. But the only definition I could find was the "3 times upper case letter" which I would interpret as 3 * 5 bit => 15 bit of total length. In the frame the space available is 16 bit. I couldn'f find a special purpose of this additional bit. Therefore any mfct value should be between 'AAA' = 1057 or 0x0421 and 'ZZZ' = 31710 or 0x7bde.

Unfortunately I could find several test frames in the code which have values > 31710.

@weetmuts What is the best way to handle these invalid frames?

Andi252:
@weetmuts what do you need to accept the pull request?

weetmuts:
Many thanks @Andi252 ! Great work in writing the c++ driver and many thanks for the fix for the weird iTW and aPT manufacturer fields. 

Nowadays, the drivers are text files stored in drivers/src 
I rewrote your code into a new driver: https://github.com/wmbusmeters/wmbusmeters/blob/master/drivers/src/itronheat.xmq

I commited the iTW workaround separately as well.
Many thanks! This pull request can be closed without a merge.

Andi252:
@weetmuts thanks for converting the driver into the xmq format.

PR #1109, opened 2023-11-25
Update driver_kamheat.cc

Jacman777:
I would like to add support to Kamstrup Multical 402  heat meter. 
[simulate_KAM402.txt](https://github.com/wmbusmeters/wmbusmeters/files/13462723/simulate_KAM402.txt)
Kam402
Adding Kamstrup Multical 402 Heat Meter


weetmuts:
Thanks!  I added the telegram to the test cases as well, to prevent any future regressions!
Do you happen to know anything about the error flags?

PR #927, opened 2023-03-24
Add stand-alone script to generate and publish Home Assistant discovery messages

pbasista:
Add a stand-alone Python script to generate and publish Home Assistant discovery messages to MQTT broker based on the files with the current meter readings and on the driver templates.

*Motivation:*

The current approach to provide Home Assistant discovery is implemented via a Home Assistant add-on. However, only [some](https://www.home-assistant.io/addons/) Home Assistant installation methods allow the users to install add-ons. For example, when using Home Assistant Core or Home Assistant Container, the option of installing and using add-ons is unavailable.

In addition, if I understand correctly, the current way of performing the Home Assistant discovery relies on [explicit](https://github.com/wmbusmeters/wmbusmeters/blob/3177c44c82c2bf08522c69b3c18a19f830937943/ha-addon/mqtt_discovery.sh#L36) definition of every meter in a configuration file. The wmbusmeters application can, however, use [wildcards](https://github.com/wmbusmeters/wmbusmeters/blob/3177c44c82c2bf08522c69b3c18a19f830937943/src/util.cc#L646) for specifying meter IDs. This allows defining multiple similar meters in a single file. I believe that such a feature significantly improves user experience and that it should be utilized by the Home Assistant discovery procedure as well.

If you find this script useful, feel free to merge it.

Issue #610, opened 2022-09-11
Support MQTT discovery outside the HA addon


patrislav1:
I'm using wmbusmeters with HA - but it runs outside of HassOS, due to limitations of Proxmox USB passthrough (the SDR stick cannot be passed through to the HassOS VM, so I have to run wmbusmeters in a separate LXC container, where USB passthrough is no problem).

So now I want to populate the sensors in HA using MQTT discovery. But in my standalone setup, I can't run the [ha-addon](https://github.com/weetmuts/wmbusmeters/tree/master/ha-addon) discovery script, as it depends on `bashio` which is not available on a regular system outside HassOS.

Looking at [mqtt_discovery.sh](https://github.com/weetmuts/wmbusmeters/blob/master/ha-addon/mqtt_discovery.sh) it seems like the "heavy lifting" for the actual discovery can be neatly separated from the interfacing to the HassOS/add-on environment. As a "proof of concept" I copy/pasted the lines [48](https://github.com/weetmuts/wmbusmeters/blob/master/ha-addon/mqtt_discovery.sh#L48)-78 from the original script and feed it with keys/values retrieved from `/etc/wmbusmeters.d` files from a standalone wmbusmeters installation.

I think the "heavy lifting" part that I copy/pasted can be refactored into a dedicated script (like `discover_meter.sh` or similar), and this script can then be called from `mqtt_discovery.sh` as well as from `mqtt_discovery_standalone.sh` (or so) to avoid code duplication.

If something like this has any chance to be upstreamed, I'd be willing to make a PR. But due to my setup I can't test if my changes break anything in the HassOS add-on.

WDYT?

Related to #575

```bash
#!/bin/bash

MQTT_ARGS="-h homeassistant.fritz.box -u mqtt-username -P mqtt-password"
SENS_CONF_DIR=/etc/wmbusmeters.d
MQTT_DISCOVERY_PREFIX="homeassistant"

for sen_conf in $SENS_CONF_DIR/*; do
    declare -A aryKV
    for line in $(cat $sen_conf); do
        IFS='=' read -r -a kv <<< "$line"
        aryKV[${kv[0]}]=${kv[1]}
    done
    echo name: ${aryKV['name']} driver: ${aryKV['driver']}

    if [ "${aryKV['id']+true}" ] && [ "${aryKV['driver']+true}" ] ; then
        file="mqtt_discovery/${aryKV['driver']}.json"
        if test -f "$file"; then
[...]

PR #1113, opened 2023-11-29
feat: add support to run shell script when meter is first seen

pimlie:
Hope you don't mind this PR 'out of the blue', feel free to close it if you don't like it :)

## What

I would like to be able to run a shell script whenever a meter is seen for the first time. This feature would also resolve the concerns expressed in this stale PR: https://github.com/wmbusmeters/wmbusmeters/pull/927 about a home-assistant addon vs integration. Also it would probably provide an easier implementation then https://github.com/wmbusmeters/wmbusmeters/issues/610 once the home-assistant wmbus integration is added.

## Why
The reason I think it would be nice to have this feature added is that users would be able to better split the device metadata en sensor data into separate topics when publishing to mqtt. And with that we could then create a similar integration as [tasmota](https://www.home-assistant.io/integrations/tasmota/) for wmbus.
F.e. a tasmota device publishes to both `tasmota/discovery/$MAC_ADDR/config` and `tasmota/discovery/$MAC_ADDR/sensors`. Within home-assistant the config topic is to listen for _device changes_, i.e. the tasmota integration listens to that topic and whenever something changes it will create/update a device within home assistant. The sensors topic is then used to populate the sensor data for that (home-assistant) device.

That translates to the following screenshot, only one mqtt device publishes to the `homeassistant` topic handled by the mqtt integration, but 2 tasmota devices publish to `tasmota/discovery` handled by the tasmota integration. The idea is that we/I would also add a wmbus integration to home assistant so that we'd have a similar card that immediately lists all wmbus devices
image

The limitation of the current implementation is that it'd require two calls to mosquitto_pub to publish both to `config` and `sensors` topic for each device. Doing that on every telegram received for a device is unneeded and not very efficient, not for mqtt and not for home-assistant to update the internal device. [This solution](https://github.com/wmbusmeters/wmbusmeters/issues/610) (when extended to the above) could work but it's not super nice as it requires external logic. It would be more user friendly if this was baked into wmbusmeters

## How
Internally the above is basically already supported in wmbusmeters through the [metermanager->addMeter](https://github.com/pimlie/wmbusmeters/blob/feat-exec-shell-on-meter-add/src/metermanager.cc#L63) method. So basically we'd only need to add a [configuration option](https://github.com/pimlie/wmbusmeters/blob/feat-exec-shell-on-meter-add/README.md?plain=1#L438) for another shell command and [execute that](https://github.com/pimlie/wmbusmeters/blob/feat-exec-shell-on-meter-add/src/main.cc#L556-L586).

An example meter config would then look like:
```
name=Water meter
driver=q400
id=12345678
metershell=mosquitto_pub -t "wmbus/discovery/12345678/config" -m "$(jq -n --arg id "$METER_ID" --arg type "$METER_TYPE" --arg name "$METER_NAME" '{id: $id, type: $type, name: $name}')"
shell=mosquitto_pub t "wmbus/discovery/12345678/sensors" -m "$METER_JSON"
```
## Remarks
- We could probably improve/split the METER_JSON variables so we don't need the `jq` command, but didn't want to make the PR too big. 
- Please also note the 3 tests I've added, f.e. [this one](https://github.com/pimlie/wmbusmeters/blob/feat-exec-shell-on-meter-add/tests/test_metershell.sh) shows that `metershell` runs only once but `shell` runs on every telegram


weetmuts:
Great work! Many thanks for this useful feature! I really appreciate that you contributed both code and test cases!

kzajac83:
@pimlie before start checking my code executed by `metershell` then want to know in details how should it works. 

1. When add new meter file to /etc/wmbusmeters.d and start/ restart wmbusmeters daemon then understood should once launch, yes?
2. On every start or restart wmbusmeters daemon should also executed once for all existed meters file in /etc/wmbusmeters.d, yes?
3. There are other conditions when is executed once ?

PR #1114, opened 2023-11-29
fix: date format should use semicolon as time separator

pimlie:
Small bug fix to replace a `.` with a `:` in the time component when printing a telegram in the human readable format


weetmuts:
Yep, you are right. Should be safe to change this, since human readable output should not be depended upon by automatic scripts. Thanks!

PR #1129, opened 2024-01-01
Added output for text_t1_meters.sh, when test failed

Andi252:
the telegram for the mkradio3 does not contain a year in there date encoding, therefore the tests have to be adapted each year.


weetmuts:
Yes! A bit silly that we have to fix the tests every year.... Something should be done before 2025. :-)
Thanks for the fix!

PR #1185, opened 2024-02-22
Update README.md

jsreynaud:
Fix an inconstancy between code that use  `--driverslib` and doc that say `--driverlib`...


weetmuts:
Thanks! The dynamic drivers are not fully functional yet, but soooon. The first two can be seen here: https://github.com/wmbusmeters/wmbusmeters-drivers/tree/main/wmbusmeters.drivers.d

PR #1196, opened 2024-03-01
topaseskr: Fix wording for date values

avandorp:
Those are period end dates, not start dates. That was my bad, sorry.


weetmuts:
Thanks!

PR #1227, opened 2024-04-01
c5isf driver extension

alalons:
Extend driver to support Heat and cold values.
Heat values are tariff 0 and cold values are tariff 1


weetmuts:
Do you have an example telegram?

weetmuts:
Without an example telegram we cannot merge this. Sorry.

PR #1288, opened 2024-05-23
Include link to Debian Experimental into README

Tom1000:
I found that wmbusmeters is also available in Debian Experimental since 2023 and propose to update README accordingly.


weetmuts:
Thanks!

PR #1297, opened 2024-06-10
Add support for W and Wh units

ottmi:
Adding support for Watts (W) and Watt hours (Wh) units on top of kW and kWh.

### DIFF
--- a/src/units.cc
+++ b/src/units.cc
@@ -41,12 +41,14 @@ using namespace std;
     X(Day,  Hour, {vto=vfrom*24.0;}) \
     X(Day,  Year, {vto=vfrom/365.2425;}) \
     X(Year,  Day, {vto=vfrom*365.2425;}) \
+    X(WH,  KWH, {vto=vfrom/1000.0;})     \
     X(KWH, GJ, {vto=vfrom*0.0036;})     \
     X(KWH, MJ, {vto=vfrom*0.0036*1000.0;})     \
     X(GJ,  KWH,{vto=vfrom/0.0036;}) \
     X(MJ,  GJ,  {vto=vfrom/1000.0;}) \
     X(MJ,  KWH,{vto=vfrom/1000.0/0.0036;}) \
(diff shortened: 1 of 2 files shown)


weetmuts:
Thanks! But why? Do you have a meter reporting Watts?

weetmuts:
Or rather, do you have a meter which has a normal reporting that is in the Watt range? :-)
What kind of meter is that?

weetmuts:
Have you run `make test` ? I think there is some internal testing (testinternals.c) that checks all the unit conversions.

ottmi:
My meters actually report kW and kWh, but I'm a big fan of storing all measurements in SI units without prefixes in the database. So if wmbusmeters reports them in the right unit already, I don't need to add an additional pre-processing step before I insert them into the database.

I've just looked into `make test` and found the section with the corresponding tests. If you feelk like merging my PR, I can certainly add the missing tests.

weetmuts:
wmbusmeters takes units very very seriously! :-) since the mbus protocol contains tons of different units for the same thing, it  is valuable to the user that wmbusmeters standardizes energy to kwh and power to kw.

However, now with your fix, you can add a conversion: calculate_total_wh=calculate_total_kwh
(Perhaps that is what was on your mind originally. :-) )
I added the missing tests for you.

Try this now:
```
wmbusmeters --format=json --calculate_total_wh=total_kwh --calculate_active_power_overall_w=active_power_overall_kw 689292680800723612452394150102cd0000008c1004688902008c1104688902008c2004000000008c21040000000002fdc9ff01ee0002fddBff01100002acff0120008240acff010a0002fdc9ff02ef0002fddBff02110002acff0224008240acff02070002fdc9ff03ee0002fddBff030e0002acff031c008240acff03060002ff68010002acff0062008240acff00190001ff1300f416 MyMeter eltako 23451236 NOKEY | jq .
```

ottmi:
Thank you! That's exactly how I'm already using my patched version :-)

Issue #1257, opened 2024-04-23
Izar driver problem with short PRIOS payload (L=0x13)


lmmodesto:
### Type of request

Bug

### OS version

Ubuntu 22.04.4 LTS - 

### wmbusmeters version

1.16.1-19-gff72e1d

### Your message goes here

Hello,

Firstly, thank you for this great project!

I'm decoding this short PRIOS telegram: 1344304c35c48c008800a0510000134ad243164a

When I run the non-debug version the telegram is decoded with some incorrect measurements.

When I run the debug version the program crashes.

Please, see additional information in the attached file.

Could you verify?

Thanks!

Marcelo Modesto

[izar wmbusmeters error.txt](https://github.com/wmbusmeters/wmbusmeters/files/15077554/izar.wmbusmeters.error.txt)

PR #1301, opened 2024-06-13
fix for issue #1257

lmmodesto:
When the telegram size is 0x13 (DLL L field) the decoded content size will be 5.

There're invalid address access in some lines like this one:

<code>double last_month_total_water_consumption_l_ = uint32FromBytes(decoded_content, 5, `true);</code>

More details at #1257


weetmuts:
Many thanks!

PR #1325, opened 2024-07-18
Update the BMeters hydrodigit driver

sajicek:
Update for the hydrodigit driver

## Bit of backstory
I've had access to BMetering software and their RFM-RX2 adapter for a few years now, but I've recently acquired the MBWUSB2 868, which comes with a lot of handy features, one of them being the ability to send (emulate) a meter response.

I've been able to decode most of the hydrodigit packet, by capturing the data from a real meter by the MBWUSB, storing them and then retransmitting those data outside of the range of the meter (to ensure the software gets the data we have and can analyze), and capturing those with the RFM and the software.

## Two of my packets
These packets were chosen intentionally, as they show how frame ids work, how a backflow is encoded, contain monthly data and the second one even has a water leak.

Before this change, the driver was able to understand only the `0C` and `04` DIFs (total_m3 and meter_datetime)

(module ids modified for the sake of privacy)
1. `4644B4092143658713077A9C000000 0C1364390400 046D212F1635 0F152A0F000000440F00C00F00511000D51000B20B00180C007C0C00E60C00560D00D10D00400E00C60E0000` - this is a nice one to showcase the backflow encoding
2. `4944B4092243658713077A7F000000 0C1363020400 046D242C1236 0F950A24042507000000A405006E0700850900CA0B004A0E00FFFFFFFFFFFF020000020000250000B3010095030000` - this one adds a leak date to the definition, plus shows how the monthly data looks before module installation

## The decoding
Sending it to the software, the first packet fills in 17 columns, the second one adds leak date. Two of these columns are from the `0C` and `04` DIFs and the rest is / must be in the 0F data.

In the following part, I will strip the bytes from the beginning and when the value has a corresponding value in the software, I will write it in the ( ) brackets at the end of the line.
&nbsp;

#### Dissecting the first packet's `0F` part:
`0F 152A0F000000440F00C00F00511000D51000B20B00180C007C0C00E60C00560D00D10D00400E00C60E0000`
`0F` - DIF - manufacturer specific data - obvious
&nbsp;

`15 2A0F000000440F00C00F00511000D51000B20B00180C007C0C00E60C00560D00D10D00400E00C60E0000`
`15` - defines the content of the data that follows, each bit might define a part of the message, or it could be just an index to a map of responses, naming it frame identifier similar to the hydroclima driver https://github.com/wmbusmeters/wmbusmeters/blob/901674d700f324b9faef80bc13680ae98b03c9a0/src/driver_hydroclima.cc#L119-L122 (has no equivalent in the software as it defines what the packet contains and which columns will be filled)
&nbsp;

`2A 0F000000440F00C00F00511000D51000B20B00180C007C0C00E60C00560D00D10D00400E00C60E0000`
`2A` - the purpose of the top half of the byte is unknown, the bottom half (`A` in this case) contains the voltage of the battery inside the module - the values are mapped as they don't increment at steady rate - `0`, `E` and `F` all map to 3.7 V, the rest from `1` to `D` is 1.9, 2.1, 2.2, 2.3, 2.4, 2.5, 2.65, 2.8, 2.9, 3.05, 3.2, 3.35, 3.5 (in software column named voltage = 3.05)
&nbsp;

`0F000000 440F00C00F00511000D51000B20B00180C007C0C00E60C00560D00D10D00400E00C60E0000`
`0F000000` - backflow - int32 - divide by 1000 to get the value in m3 (backflow shown as 0.015 m3)
&nbsp;

`440F00 C00F00 511000 D51000 B20B00 180C00 7C0C00 E60C00 560D00 D10D00 400E00 C60E00 00`
Next 12 x 3 bytes are the monthly values - be careful that they are ordered as months, as in January, February... and not (current month) -1, -2 etc.

The values represent the "total value" recorded at the end of the specified month (e.g. the value in the April column should be the same, as the value you could read as the "total_m3" at the midnight between April 30th and May 1st).

When reading from module before the "end of the month value" gets written, the value for that month is actually previous year's value, meaning the April value can be higher than the May value, when you're reading the meter in May.

`440F00` - int24 - January data - divide by 100 to get the value in m3 (39.08)
`C00F00` - int24 - February data - divide by 100 to get the value in m3 (40.32)
`511000` - int24 - March data - divide by 100 to get the value in m3 (41.77)
`D51000` - int24 - April data - divide by 100 to get the value in m3 (43.09)
`B20B00` - int24 - May data - divide by 100 to get the value in m3 (29.94)
`180C00` - int24 - June data - divide by 100 to get the value in m3 (30.96)
`7C0C00` - int24 - July data - divide by 100 to get the value in m3 (31.96)
`E60C00` - int24 - August data - divide by 100 to get the value in m3 (33.02)
`560D00` - int24 - September data - divide by 100 to get the value in m3 (34.14)
`D10D00` - int24 - October data - divide by 100 to get the value in m3 (35.37)
`400E00` - int24 - November data - divide by 100 to get the value in m3 (36.48)
`C60E00` - int24 - December data - divide by 100 to get the value in m3 (37.82)

and the remaining `00` with unknown purpose
&nbsp;

#### Few highlights from the second packet's `0F` part:
`0F 95 0A 240425 07000000 A40500 6E0700 850900 CA0B00 4A0E00 FFFFFF FFFFFF 020000 020000 250000 B30100 950300 00`
The frame identifier is `95` (only a single bit difference from `15`, and contains only one more info - might be a coincidence, might not) - adds leak date as the first 3 bytes after voltage and before backflow.

Speaking of voltage, the byte is now `0A` and it's still parsed only as voltage: 3.05 V

`240425` - leak date in YYMMDD format, all in BCD (in software: leak date: 25.04.2024)

Another obvious difference are the `FFFFFF` months, these are before the module was active and get translated to 0. The next two months have a small amount flowed through them during the product verification. 
&nbsp;

Anyone who can count might have noticed that the number of columns decoded from the packet is 16 (+1 for leak date in the second packet) - 2 from before the `0F`, 1 voltage, 1 backflow and 12 monthly data, but in the beginning, I said it fills 17 (+1). That last one is "number of alarms" field, which was always zero for the tested packets. I've tried changing the `00` with unknown purpose at the end to see if it was somehow encoded in there, but none of the tested hex values seemed to change anything. The number of alarms might also be derived from the frame identifier, which might specify that there will be an "errors" field in the packet and if it's not there, then just print 0. But that's all just a speculation on how it works.

## Conclusion
Enjoy the updated driver and if you have a packet with different frame identifier than the ones identified (if you try to run it through, it will tell you), please open an issue with the packet included so we can all learn more.

Before someone says it, I know there is a packet with a different frame id in the driver file (the warm water one), specifically frame id `03`, but it doesn't contain much (the first `00` after the frame id has an unknown purpose, the rest - 4x `00` is mapped to backflow). Didn't include that in the driver as I've never seen a packet like that elsewhere (plus it would be slightly annoying to add, as it has a different structure). There is one interesting thing about that packet - it doesn't have the `00` with unknown purpose at the end, which might suggest that it's somehow related to some part of the other data.
&nbsp;

Also, if anyone wants to add / improve the decoding on any BMeters products I might be able to help with that (the least I can do is translate the data to columns and values for you, so you could figure out how that's all in there)

### DIFF
--- a/src/driver_hydrodigit.cc
+++ b/src/driver_hydrodigit.cc
@@ -13,55 +13,287 @@
 
  You should have received a copy of the GNU General Public License
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
-*/
+ */
 
 #include"meters_common_implementation.h"
 
-namespace
-{
-    struct Driver : public virtual MeterCommonImplementation
-    {
-        Driver(MeterInfo &mi, DriverInfo &di);
-    };
+namespace {
+	struct Driver: public virtual MeterCommonImplementation {
+			Driver(MeterInfo &mi, DriverInfo &di);
+			void processContent(Telegram *t);
+	};
 
-    static bool ok = registerDriver([](DriverInfo&di)
-    {
-        di.setName("hydrodigit");
-        di.setDefaultFields("name,id,total_m3,meter_datetime,timestamp");
-        di.setMeterType(MeterType::WaterMeter);
-        di.addLinkMode(LinkMode::T1);
-        di.addDetection(MANUFACTURER_BMT,  0x06,  0x13);
-        di.addDetection(MANUFACTURER_BMT,  0x06,  0x17);
-        di.addDetection(MANUFACTURER_BMT,  0x07,  0x13);
-        di.addDetection(MANUFACTURER_BMT,  0x07,  0x15);
-        di.setConstructor([](MeterInfo& mi, DriverInfo& di){ return shared_ptr<Meter>(new Driver(mi, di)); });
-    });
+	static bool ok = registerDriver([](DriverInfo &di) {
+		di.setName("hydrodigit");
+		di.setDefaultFields("name,id,total_m3,meter_datetime,timestamp");
+		di.setMeterType(MeterType::WaterMeter);
+		di.addLinkMode(LinkMode::T1);
+		di.addDetection(MANUFACTURER_BMT, 0x06, 0x13);
+		di.addDetection(MANUFACTURER_BMT, 0x06, 0x17);
+		di.addDetection(MANUFACTURER_BMT, 0x07, 0x13);
+		di.addDetection(MANUFACTURER_BMT, 0x07, 0x15);
+		di.usesProcessContent();
+		di.setConstructor([](MeterInfo &mi, DriverInfo &di) {
+			return shared_ptr<Meter>(new Driver(mi, di));
+		});
+	});
 
-    Driver::Driver(MeterInfo &mi, DriverInfo &di) : MeterCommonImplementation(mi, di)
-    {
-        addNumericFieldWithExtractor(
-            "total",
-            "The total water consumption recorded by this meter.",
-            DEFAULT_PRINT_PROPERTIES,
-            Quantity::Volume,
-            VifScaling::Auto, DifSignedness::Signed,
-            FieldMatcher::build()
-            .set(MeasurementType::Instantaneous)
-            .set(VIFRange::Volume)
-            );
+	Driver::Driver(MeterInfo &mi, DriverInfo &di) : MeterCommonImplementation(
+			mi, di) {
+		addNumericFieldWithExtractor("total",
+				"The total water consumption recorded by this meter.",
+				DEFAULT_PRINT_PROPERTIES, Quantity::Volume, VifScaling::Auto,
+				DifSignedness::Signed,
+				FieldMatcher::build().set(MeasurementType::Instantaneous).set(
+						VIFRange::Volume));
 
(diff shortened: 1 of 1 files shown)


weetmuts:
Great work! Many thanks! I added the test cases to the source file and replaced the tabs with 4 spaces. ;-)

jukkapirinen:
@sajicek nice work. Have you had any more data fields to the hot water meter (0x17)?  Or received telegrams with other status than ok?
I have been working with the cold version of 0x17 meter. Here are my findings. 
https://github.com/wmbusmeters/wmbusmeters/issues/1342
In the issue there is a link to the protocol specification,  but seems MS code bits are different than specified at least for 0x13 meters.

Issue #1361, opened 2024-09-13
1.17.1 driver autodetection is broken


kitor:
### Type of request

Question

### OS version

Debian 12.5

### wmbusmeters version

wmbusmeters: 1.17.1-4-g3eee107

### Your message goes here

I was going to update `amiplus` driver with new IDs and test cases for Opus 1 and 3 (as described by me and Leo-PL in #1296)

I built wmbusmeters as usual:
```
apt install librtlsdr-dev libxml2-dev rtl-sdr
./configure
make clean
make -j6
make install
systemctl daemon-reload
systemctl enable wmbusmeters
mkdir -p /var/log/wmbusmeters/meter_readings
chown wmbusmeters:wmbusmeters /var/log/wmbusmeters/meter_readings
systemctl start wmbusmeters
```

What I'm surprised is that now every driver complains to me about auto detection failure, like my very own `apatoreitn`:

```
[2024-09-13_21:48:11] (meter) heat: meter detection did not match the selected driver apatoreitn! correct driver is: unknown!
(meter) Not printing this warning again for id: 35365047 mfct: (APA) Apator, Poland (0x8601) type: Heat Cost Allocator (0x08) ver: 0x04
[2024-09-13_21:48:11] (meter) please consider opening an issue at https://github.com/wmbusmeters/wmbusmeters/
[2024-09-13_21:48:11] (meter) to add support for this unknown mfct,media,version combination
telegram=|19440186475036350408A0_A100003200000000002D310815B018|+375
```
 
while [src/driver_apatoretin.cc](https://github.com/wmbusmeters/wmbusmeters/blob/3eee107dd8f50ca9b5cbfab46a50f3103991240c/src/driver_apatoreitn.cc#L40) clearly matches:
`        di.addDetection(0x8614 /* APT? */, 0x08,  0x04);`

Similar for `apator162`:

```
[2024-09-13_21:42:53] (meter) woda: meter detection did not match the selected driver apator162! correct driver is: unknown!
(meter) Not printing this warning again for id: 00058812 mfct: (APT) Apator, Poland (0x8614) type: Water meter (0x07) ver: 0x05
[2024-09-13_21:42:53] (meter) please consider opening an issue at https://github.com/wmbusmeters/wmbusmeters/
[2024-09-13_21:42:53] (meter) to add support for this unknown mfct,media,version combination
```

[...]

Issue #1296, opened 2024-06-10
TAURON Amiplus update


Mice30:
### Type of request

Existing device improvement

### Meter description

Hi,

I just added my Tauron energy meter to Home Assistant using the HA add-on. After entering the decryption key, I received a log message indicating that the Amiplus driver is not suitable for my meter. Some data is present, but not all the data that the telegram contains.

Can you help, please? :)

### Logged telegrams from meter in decrypted format

```shell
telegram=|8E4401061764695601027AE01380052F2F_066DC04B0E0A36800C78176469560AFDC9FC0143020AFDC9FC0243020AFDC9FC0343028E30833C0000000000008E20833C0000000000008E10833C3202000000000BABC8FC100000008E10030648510200008E20030000000000008E30030000000000000B2B9502000BAB3C0000002F2F2F2F2F2F2F2F2F2F2F2F2F2F2F|
```

wrobelda:
https://github.com/wmbusmeters/wmbusmeters/pull/1231

wrobelda:
You probably have the new generation of AMIPlus meters installed. I previously added a [PR](https://github.com/wmbusmeters/wmbusmeters/pull/1231) to have wmbusmeters actually recognize it as amiplus meter, although as of recent version this is no longer needed for whatever reason. 

Still, it would be nice to have the extra information decrypted as well, many of the basic parameters are currently missing.

pswiatki:
But wait! I have NES 73331 - no problems with extra parameters.

wrobelda:
> But wait! I have NES 73331 - no problems with extra parameters.

I can't recall, actually, it's been a few months since I last checked this, but my voltage is reported as Phase 3 and I think some tariff information was missing, too. But frankly, I may be wrong here, I didn't investigate this too much as I was after the energy consumption only.

Mice30:
It's great to see some feedback! I'll post the type of meter I have later on.

wrobelda:
See also this bug report:

https://github.com/wmbusmeters/wmbusmeters/issues/1230

abht234:
I can't get my NES 73331 meter to work. I am using the decryption key provided by the meter supplier. Any help would be greatly appreciated. 

```
(meter) myenergy: meter detection did not match the selected driver amiplus! correct driver is: unknown!
[...]

PR #1362, opened 2024-09-14
Fix driver autodetection, add new meter to amiplus, cleanup in apatoreitn code.

kitor:
Fixes driver autodetection issue from #1361 introduced in fccdea9345ee40450b93ccb58b4042efba61906a

Updates apatoreitn driver - I replaced hardcoded manufacturer id codes with global constants

Updates amiplus driver - support for Otus 1/3 as mentioned in #1296

What I'm not sure in how to proceed, as mentioned in fcbfce99b599eaff49d7bf22c4ba846b776068f2:

```
        // Apator Otus 1/3 seems to use both, depending on a frame.
        // Frames with APA are successfully decoded by this driver
        // Frames with APT are not - and their content is unknown - perhaps it broadcasts two data formats?
        di.addDetection(MANUFACTURER_APA,  0x02,  0x01);
        //di.addDetection(MANUFACTURER_APT,  0x02,  0x01);
```

I left the unimplemented telegram variant commented out. It doesn't decode via apatoreitn driver (maybe it contains new data format as mentioned in #1296? I remember similar stuff with some apator water meter).
This has the unfortunate outcome of wmbusmeters still trying to decode both frames and raising "meter detection did not match" warning in log. Perhaps I should uncomment it?

I did not run `make testd` as it doesn't work for me. After making a debug build and attempting to run `make testd` my console is indefinitely spammed with  `AddressSanitizer:DEADLYSIGNAL`.

### DIFF
--- a/src/driver_amiplus.cc
+++ b/src/driver_amiplus.cc
@@ -33,6 +33,11 @@ namespace
         di.addDetection(MANUFACTURER_APA,  0x02,  0x02);
         di.addDetection(MANUFACTURER_DEV,  0x37,  0x02);
         di.addDetection(MANUFACTURER_DEV,  0x02,  0x00);
+        // Apator Otus 1/3 seems to use both, depending on a frame.
+        // Frames with APA are successfully decoded by this driver
+        // Frames with APT are not - and their content is unknown - perhaps it broadcasts two data formats?
+        di.addDetection(MANUFACTURER_APA,  0x02,  0x01);
+        //di.addDetection(MANUFACTURER_APT,  0x02,  0x01);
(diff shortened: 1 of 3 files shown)


kitor:
Huh, those tests failed on something non-related...

weetmuts:
Thanks! I do not think you need (dd.mfct & 0x7fff)
because the driver mfct should be correct, it was broken with 0x8601 /* APA? */
but you fixed this. But it does not hurt either. :-)

weetmuts:
The debug version runs with address sanitizer to find memory bugs. However in the latest Ubuntu and perhaps others,
the address space randomization conflicts with the address sanitizer. You have to temporarily disable the randomization:
echo 0 > /proc/sys/kernel/randomize_va_space

Issue #1306, opened 2024-06-19
Hydrocal M4 cooling


Webbeh:
BMeters Hydrocal M4
Using driver hydrocalm3

The data for heating is decoded properly, but cooling isn't. Same for the current volujme and power.

The device is showing 0.139 kWh of cooling energy and 0.043 m^3 of cooling volume.

Device is setup to send the "combined" data (volume and energy for heating and cooling)
https://wmbusmeters.org/analyze/3a44B409238709051a0d8c00187ac0000000046d26B113360c03060100000c13470000008c1003390100008c1013430000000f6401000000000000 

This is in "instantaneous" mode (sends the heating kWh+m3, volume, power and temperatures)
https://wmbusmeters.org/analyze/3E44B409238709051A0D8C00187ACC000000046D29B113360C03060100000C13470000000B3B0000000B280000000A5998250A5D74250F6402000000000000

and another one with some volume flow > 0
https://wmbusmeters.org/analyze/3E44B409238709051A0D8C00187AE6000000046D2FB113360C03060100000C13470000000B3B6939000B280000000A5966260A5D97250F6402000000000000
(Value could be quite high, I was blowing in the device to get it spinning)

Again, this is a hydrocal M4 and not a M3. Mfg date : 2023.

Thank you!

kzajac83:
I was wondering what is difference between below frame types in Hydrocal M4 and now understood, both no coding frames

1. Standard frame, 45 bits, only heating info
https://wmbusmeters.org/analyze/2c44B409418012051a0d8c20837a2f120020046d24B401390c08000000000c13000000000f6400100000000000


3. Extend frane, 59 bits, cool+heating
https://wmbusmeters.org/analyze/3A44B409418012051A0D8C20CD7A79120020046D32A002390C08000000000C13000000008C1008000000008C1013000000000F6401100000000000


This data come from new not used heat meter hence zeros energy and flows.  

Just need upgrade drivers for extend frame in your cause and also you need switch Hydrocal M4 from standard frame to extend frame.

Webbeh:
I am 100% confident the cooling frames did work at some point, as I saw that data from wmbusmeter software on a linux machine a few months ago.

However, I am unsure whether bmeters updated the firmware of the devices in the meantime and changed the data format in a new firmware...

kzajac83:
[...]

PR #1379, opened 2024-10-02
Add specific driver hydrocalm4

sunflowerenergias:
I have generated a new specific driver for the BMeters Hydrocal M4 meters, the one from the Hydrocal M3 model was currently being used and was not 100% compatible with the cooling values.

More information can be seen in the issue #1306


weetmuts:
Great work! But all new drivers should be written as text config files.
See if you can create drivers/src/hydrocalm4.xmq 

wmbusmeters --analyze=drivers/src/hydrocalm4.xmq 3A44B409381317051A0D8C00B07A7D000000_046D05AA023A0C03000000000C13000000008C1003050000008C1013040000000F6401000000000000

sunflowerenergias:
I have modified the driver generation to use the text configuration format

Thanks!

weetmuts:
Great work! Many thanks!

PR #1387, opened 2024-10-09
Allow daemon to be started with custom config

ivivanov-bg:


weetmuts:
Can you add a comment in the README in this pull request on how this works from the command line.

ivivanov-bg:
The entry is already there:
```
You can also start the daemon with another set of config files: wmbusmetersd --useconfig=/home/me/.config/wmbusmeters /tmp/thepidfile
```

It's just that it wasn't working

weetmuts:
Ouch, thank you very much for the fix!

PR #1396, opened 2024-10-18
Add support for Kamstrup MC303 cooling energy

makischu:
The heat meter Kamstrup Multical 303 is available as pure heat meter or as heat+_cooling_ meter. 
wmbusmeters currently offers great support for the pure heat meter configuration, but fails to decode the cooling energy register, which the meter only provides in the heat+cooling configuration, which seems to be less common.
I implemented it, added a test case, and propose to pull it, as others might also benefit from.

--analzye before:
```
025   : 04 dif (32 Bit Integer/Binary Instantaneous value)
026   : 86 vif (Energy kWh)
027   : 3C combinable vif (BackwardFlow)
028 C?: 00000000
```
--analyze after:
```
025   : 04 dif (32 Bit Integer/Binary Instantaneous value)
026   : 86 vif (Energy kWh)
027   : 3C combinable vif (BackwardFlow)
028 C!: 00000000 ("total_energy_backward_kwh":0)
```

### DIFF
--- a/src/driver_kamheat.cc
+++ b/src/driver_kamheat.cc
@@ -127,6 +127,18 @@ namespace
             );
 
         addNumericFieldWithExtractor(
+            "total_energy_backward",
+            "The total backwards energy consumption recorded by this meter.",
+            DEFAULT_PRINT_PROPERTIES,
+            Quantity::Energy,
+            VifScaling::Auto, DifSignedness::Signed,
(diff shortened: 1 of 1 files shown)


weetmuts:
Great work! I will merge this since it is a patch on an existing driver. 

All new drivers should be written as text files! Check drivers/src for the new ones!

PR #1449, opened 2024-12-30
Extended support for Maddalena radio evo (evo868)

Stubenh0cker:
Hi there,

since I had issues with reading the telegrams correctly I added support for cold and warm water meters (0x16, 0x6), since the driver_evo868.cc only seem to support water meters (0x7)


weetmuts:
Thanks!

Issue #1519, opened 2025-04-29
Make docker-entypoint.sh "configurable"


kubax:
### Type of request

Enhancement

### OS version

Linux

### wmbusmeters version

latest

### Your message goes here

Hi,

i just installes wmbusmeters on my kubernets cluster, and wanted to have the config file in a configmap instead of a file on a longhorn volume, so i modified the docker-entrypoint.sh a bit to make it a bit more configurable from where the config file is loaded (and other directories) and though you may include it to the container.

```
#!/bin/sh
set -e

: ${WMBUSMETERS_DATA:=/wmbusmeters_data}
: ${LOGS_DIR:=${WMBUSMETERS_DATA}/logs}
: ${READINGS_DIR:=${WMBUSMETERS_DATA}/readings}
: ${ETC_DIR:=${WMBUSMETERS_DATA}/etc}

[ ! -d "$LOGS_DIR" ] && mkdir -p "$LOGS_DIR"
[ ! -d "$READINGS_DIR" ] && mkdir -p "$READINGS_DIR"
[ ! -d "$ETC_DIR/wmbusmeters.d" ] && mkdir -p "$ETC_DIR/wmbusmeters.d"
[ ! -d "$ETC_DIR/wmbusmeters.drivers.d" ] && mkdir -p "$ETC_DIR/wmbusmeters.drivers.d"

if [ ! -f "$ETC_DIR/wmbusmeters.conf" ]; then
  cat <<EOF > "$ETC_DIR/wmbusmeters.conf"
loglevel=normal
device=auto:t1
donotprobe=/dev/ttyAMA0
logtelegrams=false
format=json
meterfiles=${READINGS_DIR}
meterfilesaction=overwrite
logfile=${LOGS_DIR}/wmbusmeters.log
EOF
fi

exec /wmbusmeters/wmbusmeters --useconfig="$ETC_DIR"
```

this way one can add environment variables to set
- ETC_DIR where the userconfig will be stored (and even one created if none is available)
- WMBUSMETERS_DATA to define the root data directory
- LOGS_DIR to define another log directory when creating the userconfig
- READINGS_DIR to define a directory for the readings when creating the userconfig

as i did not build a complete dockerimage (i just overwritten the docker-entrypoint.sh with a one i have in a configmap in kubernetes) i did not create a pull request as it MIGHT make problems. on my system it looks like it's running fine.

weetmuts:
Thanks! Can you make a pull request?

[...]

PR #1524, opened 2025-05-13
Make directories for the Docker version more configurable through environment variables

kubax:
This is the PR to the Issue #1519

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -3,7 +3,7 @@ set -e
 
 : ${WMBUSMETERS_DATA:=/wmbusmeters_data}
 : ${LOGS_DIR:=${WMBUSMETERS_DATA}/logs}
-: ${READINGS_DIR:=${LOGS_DIR}/readings}
+: ${READINGS_DIR:=${LOGS_DIR}/meter_readings}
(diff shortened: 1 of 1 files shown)


weetmuts:
Just checking, this change is backwards compatible? I.e. other docker users will not experience any change?

kubax:
you might want to alter the config paths to the ones that were set previously, because in the PR i might have changed it to the ones i am using...

```
: ${WMBUSMETERS_DATA:=/wmbusmeters_data}
: ${LOGS_DIR:=${WMBUSMETERS_DATA}/logs}
: ${READINGS_DIR:=${WMBUSMETERS_DATA}/readings}
: ${ETC_DIR:=${WMBUSMETERS_DATA}/etc}
```

In my PR i have the readings directory set to `/wmbusmeters_data/readings` while the default was `/wmbusmeters_data/logs/meter_readings` but i don't know if that would affect someone, because i don't know if they are needed after processing...

I did just test it for my needs and have it running this way since, sorry :/

weetmuts:
@BIBOLV Can you check that this change is ok?

BIBOLV:
Changed readings dir to previous. Tnx for PR!

PR #1553, opened 2025-07-11
Update driver_hydrodigit.cc

blavka:
Replaced hardcoded frame identifier checks with bitmask logic for better flexibility and clarity.

Also implemented decoding of the fraud date field, including extraction of fraud type flags from the upper nibble of the month byte.


weetmuts:
Great work!

weetmuts:
I just made a small change to the contents field, instead of:

"Battery voltage, Backflow, Monthly data" 
it generates
"BATTERY_VOLTAGE BACKFLOW MONTHLY_DATA"

that way it consistent with all other boolean values in the majority of the drivers.

weetmuts:
Change leak_date/fraud_date to use standard format 2024-05-24 which is used by all other driver fields with suffix _date in wmbusmeters. :-)

PR #1557, opened 2025-07-22
axioma to use q400 driver

leifnel:
By changing 0x10 to 0x01 it works with my axioma water-meter.
di.addDetection(MANUFACTURER_AXI, 0x07,  0x01);

I don't know if it was a typo or both versions exist.

### DIFF
--- a/src/driver_q400.cc
+++ b/src/driver_q400.cc
@@ -30,7 +30,7 @@ namespace
         di.setDefaultFields("name,id,total_m3,timestamp");
         di.setMeterType(MeterType::WaterMeter);
         di.addLinkMode(LinkMode::T1);
-        di.addDetection(MANUFACTURER_AXI, 0x07,  0x10);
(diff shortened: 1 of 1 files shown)


weetmuts:
Great! Do you have an example telegram for the test suite?

leifnel:
614409074897280601077AD1000000046D2C173B370420580193040413FB94000004933BFB94000004933C01000000023B02000259E306446D000021374413C08A000044933BC08A000044933C0100000034FD17000000000424BA01930401FD745B

weetmuts:
>I don't know if it was a typo or both versions exist.

Thanks! The manuacturers often have several different version numbers.

weetmuts:
The existing test suite shows that there are telegrams with AXI, 0x07, 0x10
so I made sure both 0x10 and 0x01 is part of the driver and added your telegram
to the test suite as well. 

Thanks!

PR #1560, opened 2025-07-30
Fix issue 1559 - Correct calculation of expected length when AMB8465 is in CMD mode

colintd:
payload_len in CMD mode already includes length of RSSI if present.  Existing calculation added an extra byte, meaning messages were never completed, and instead discarded.  Fix is just to remove the + rssi_len.  Tested this morning on Debian.


weetmuts:
Aha! Many thanks!

PR #1601, opened 2025-10-08
Harden config parsing: CRLF support and numeric meter ID acceptance

LiLasini:
Improves config parsing robustness while keeping existing behavior for valid files.
Changes:
•	Line-based parsing with CRLF support and leading/trailing whitespace trimming.
•	Skips empty/comment lines earlier and warns on malformed lines without '='.
•	Accepts pure numeric meter IDs without warning (only warns if neither numeric nor valid expression).
•	Keeps key secrecy (key=<notprinted>).
No changes to device handling, link mode calculation, or shell execution logic.


weetmuts:
Nice! But could you fix the indentation so that else if, is on the same line? Makes it easier to compare.

LiLasini:
Identation fixt, so else if is on the same line.

Please check the modified meter id validation:
// Accept pure numeric non-empty IDs silently (common case).

If it is not necessary after all, I can remove it.

HaraldSchmitt:
On my Raspberry PI installation, I had the problem “no meterfiles configured,” even though I had some in wmbusmeters.d.
I compiled this pull request and then it worked.
Apparently, there is a problem when the meterfiles in wmbusmeters.d are numeric. With the fix, the problem disappears.

PR #1594, opened 2025-09-28
New driver for NZR EcoCount Compact electricity meters

spectrality-github:
New driver for NZR EcoCount Compact electricity meters as amiplus driver is not fully matching.

Some details was here:
https://github.com/wmbusmeters/wmbusmeters/issues/1040

Later will try to submit NZR EcoCount Compact electricity meter discovery template for wmbusmeters-ha-addon
and my implementation of Mbus TCP using socat for  wmbusmeters-ha-addon

Issue #1040, opened 2023-09-30
Wired Mbus TCP support


spectrality-github:
### Type of request

Enhancement

### OS version

HA addon

### wmbusmeters version

Latest

### Your message goes here

   I have my Mbus master USB adapter (from AliExpress) connected to OpenWrt device with `ser2net` running.
And  I'm interested in native Mbus TCP support in wmbusmeters, as it is implemented in libmbus.
`mbus-tcp-request-data`
http://www.rscada.se/libmbus/

 Currently trying to workaround lack of Mbus TCP support by adding and starting `socat` inside wmbusmeters-ha-addon container. Like described here: https://www.lucid-control.com/accessing-usb-io-module-network-linux-socat/ 
 
 Next line is added before starting wmbusmeters in run.sh:
`socat -d -d pty,group-late=tty,link=/root/ttyMBUS0,mode=660,rawer,echo=0,b2400,waitslave,ignoreeof tcp:192.168.x.x:2003 &`
And using `/root/ttyMBUS0` as device in wmbusmeters config.
ser2net config is:
`2003:raw:30:/dev/mbusUSB0:2400 8DATABITS EVEN 1STOPBIT -RTSCTS LOCAL -XONXOFF`

But socat process is stopping and not restarting when ser2net is closing TCP connection by timeout or connection problems and whole container need to be restarted to restore connectivity. Also in this case ser2net timeout should be alwai configured longer that pool interval. 

Also there is problem with MQTT discovery form wired Mbus devices
 As colons are required in driver name (amiplus:mbus) for wired Mbus devices, but should be removed to match discovery template files.
```
MQTT Discovery ...
 Adding meter: MyMeter ...
  File /config/wmbusmeters/etc/mqtt_discovery/amiplus:mbus.json not found.
MQTT Discovery cleanup...
```
 I fixed that in my private fork with additional line in mqtt_discovery.sh  https://github.com/wmbusmeters/wmbusmeters/issues/711
`aryKV["driver"]=${aryKV["driver"]%%:*}`
https://github.com/spectrality-github/wmbusmeters-ha-addon/blob/9351b966d42ccfe838aa140dc314b67b0fb45570/wmbusmeters-ha-addon-edge/mqtt_discovery.sh#L42
It word be good if this will be fixed in Main branch.

[...]

PR #1620, opened 2025-10-26
New xmq driver for NZR EcoCount Compact electricity meters

spectrality-github:
New driver for NZR EcoCount Compact electricity meters as amiplus driver is not fully matching.

Some details was here: https://github.com/wmbusmeters/wmbusmeters/issues/1040

https://www.nzr.de/leistungen/messtechnik/elektrizitaet/hutschiene-3/ecocount-compact
NZR M-Tool
https://www.nzr.de/fileadmin/user_upload/software/M-Tool_5.3.zip

PR for c++ driver https://github.com/wmbusmeters/wmbusmeters/pull/1594

### DIFF
--- a/drivers/src/nzr.xmq
+++ b/drivers/src/nzr.xmq
@@ -1,0 +1,197 @@
+// Copyright (C) 2024 Fredrik Öhrström (gpl-3.0-or-later)
+driver {
+    name           = nzr
+    default_fields = name,id,total_energy_consumption_kwh,current_power_consumption_kw,current_power_consumption_1_kw,current_power_consumption_2_kw,current_power_consumption_3_kw,voltage_at_phase_1_v,voltage_at_phase_2_v,voltage_at_phase_3_v,current_at_phase_1_a,current_at_phase_2_a,current_at_phase_3_a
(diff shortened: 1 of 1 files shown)


weetmuts:
Great work! I just added `attributes=STATUS,INCLUDE_TPL_STATUS` 
to the status field so that it automatically picks up any status bits from the tpl status.

Then I minimized the default fields. The reason is that the fields can always be selected 
with `--format=fields --selectfields=voltage_at_phase_1_v,current_power_consumption_1_kw`
the default fields are just there to see something interesting from the meter when just testing it.
Typically `name,id,status,total_energy_consumption_kwh,timestamp` or any total that is relevant.

It is sort of a remnant from the beginning, however the default fields are tricky, they cannot
be changed since this would cause a incompatible change. Unfortunately when cleaning up
old drivers this has had to happen. To avoid such problems, put as few fields into the default 
fields as possible.

weetmuts:
https://github.com/wmbusmeters/wmbusmeters/commit/321a30bbed216872b3ed492d0a8c78cf28b4a987

PR #1633, opened 2025-11-13
Added Lansen and Sharky Drivers, hybrid frame (starts with 68 and contains 5e or 4e later) support and improved stability.

seetherisingazeez:


weetmuts:
Oups. This did not work. I had to revert the merge. It think this requires a better way of jumping into the wmbus part of the mbus telegram that does not require ignoring warnings about lengths....

weetmuts:
Is the 
051   : 0D dif (variable length Instantaneous value)
052   : FD vif (Second extension FD of VIF-codes)
053   : 3B vife (Reserved)
054   : 5F varlen=95

a standard for embedding wmbus or is a manufacturer specific solution?

PR #1641, opened 2025-11-19
CSV/fields output shows placeholders like ?consumption_at_history_5_m3? for dynamic (template based) fields 

LiLasini:
Problem: CSV/fields output shows placeholders like ?consumption_at_history_5_m3? for dynamic (template based) fields 
(e.g. history_x_date, consumption_at_history_x_m3). 
These fields are correctly decoded and appear in JSON, but not in CSV when explicitly listed via selectfields/selectedfields.

Root cause: The CSV rendering logic (checkPrintableField in meters.cc) only matches static driver field template names (fi.vname()+"_"+unit). 
For template-expanded fields the internal storage uses the expanded name (e.g. consumption_at_history_5) while the loop compares 
against the unexpanded template string containing {storage_counter ...}. Therefore selection fails and a placeholder is emitted.

Fix: Add a minimal pre-check that attempts to interpret each requested CSV field as an already expanded numeric field:
1.	Strip unit suffix (e.g. _m3, _date).
2.	Look up the value directly from the meter’s numeric value cache (pair of expanded name + unit).
3.	If found, render it normally and skip the old template comparison.

The meters.cc patch (adding the tryExpandedTemplateField pre-check) fixes this generically for all drivers.

This is my second pullrequest for problems with meterfiles. Maybe meterfiles an csv are not often used, and for this reason you will have problems with it.


weetmuts:
Your proposed pull request strips away all comments? Can you fix the pull request so that it is minimal. It is a bit hard to see what is going on.

LiLasini:
I cannot make these changes. The pull request is a suggestion on how it can be solved.
However, I have described the problem in detail. Feel free to fix it with your own code. There is a discussion with explanations.

weetmuts:
ok, I will keep this open while I ponder how to fix this.

weetmuts:
Many thanks! I merged your fix separately and added a regression test case.

PR #1649, opened 2025-11-25
Add dBm UoM

kzajac83:


weetmuts:
Nice! But if 1 dBm is equivalent to 0.001 watts or one milliwatt, shouldn't dBm be of the same SI unit as Watts... hmmm.

But alas, if we create a dBm unit from the SI units:     X(W,      1.0,        SIExp().kg(1).m(2).s(-3))         
w        Watt (W)                         1kgm²s⁻³

Then wmbusmeters will apply the linear transform between W and dBm, which is not correct since the transform is logarithmic/exponential.

However we can create a new SI unit dbm, then define: X(DBM, 1.0, SIExp().dbm(1))
Then we can add the explicit conversion between DBM and W and KW in the LIST_OF_CONVERSIONS
where the conversion code uses log/exp.

Its not like I think there will be any common use of the ability to convert between dBm and W, but since the unit system is pretty advanced, it would be sad not to make a best effort when implementing dbm.

weetmuts:
dbm is not an SI unit, but the trick of having distinct units for temperature F,K,C to block any automatic conversions has been used before. The name SIExp is perhaps not the best name any more. SIExpAndSomeMore...

pswiatki:
> Nice! But if 1 dBm is equivalent to 0.001 watts or one milliwatt, shouldn't dBm be of the same SI unit as Watts... hmmm.

I posit: 0 dBm is 0.001 W (1 mW). :-)

weetmuts:
@pswiatki You are absolutely right! :-) 0 dBm is 1mW. 

Been thinking, I think the best solution is to add a `bool non_linear;` to the SIExpr class. If this is true, then wmbusmeters will not use the linear scaling factor, instead it will apply a custom transform.

Then dBm will be a power unit like W and the automatic conversion will work.

What do you think @pswiatki @kzajac83 ?

kzajac83:
dBm as a unit is dimensionless UoM, because dBm is defined in reference to 1 mW (dBµ to microV, dBV to 1V, etc.) . In other words, the numerator 𝑃[mW]  divided by the denominator 1 mW gives a dimensionless ratio.

image

Since the reference is 1 mW, the denominator can be omitted when the input is already in mW:

image


The same logic applies to other dB-referenced units:
dBV (referenced to 1 V), dBu (referenced to 0.775 Vrms), dBW (referenced to 1 W, meaning 0 dBW = 30 dBm), etc.

From an engineering point of view these formulas are standard. **The above is the conversion from mW to dBm**.

Frederic is correct that this is a logarithmic scale, so the reverse conversion (from dBm back to linear units in mW) is:

image

or to get power in watts:

image


So agree Frederic with your proposition :)

PS. 1 dBm is approximately 1.26 mW, not 1 mW ;)

BR, Krzysiek

pswiatki:
Is this whole dBm thing exclusively for test & verification infrastructure?

kzajac83:
No, dBm is not exclusive to test and verification environments.
dBm is a standard unit used everywhere in RF engineering, telecommunications, wireless systems, antenna measurements, radio modules, IoT devices, and even everyday equipment like Wi-Fi access points or Bluetooth chips. And of course WM-Bus :)

It’s simply a logarithmic expression of absolute power referenced to 1 mW, so it is used both in normal operation and in lab/testing setups.

pswiatki:
> No, dBm is not exclusive to test and verification environments. dBm is a standard unit used everywhere in RF engineering, telecommunications, wireless systems, antenna measurements, radio modules, IoT devices, and even everyday equipment like Wi-Fi access points or Bluetooth chips. And of course WM-Bus :)
> 
> It’s simply a logarithmic expression of absolute power referenced to 1 mW, so it is used both in normal operation and in lab/testing setups.

I am a software engineer who has worked closely with RF devices for the last 20 years or so. So, in general, I am very much aware of the above, Krzyśku. My question was local to this place, where it is used here in the `wmbusmeters` project. 
I suspect some devices do report RF levels in `dBm` directly, while the most popular dongle (`rtlsdr`) is problematic in this regard. Hence my question.

kzajac83:
In the wmbusmeters project, the naming convention is:

`somename_UoM `

(where UoM stands for unit of measurement).

Answering your question Piotrku, the main parameter for signal strength is `rssi_dbm`.
The inconsistency in wmbusmeters is that, although the unit after the underscore should be `dBm`, this information is missing when running:

`wmbusmeters --listunits`

An additional benefit is the (probably rarely used, but still) possibility to change the currently non-existent dBm unit.

weetmuts:
I do like the idea of adding the dBm unit (suffix _dbm in wmbusmeters lingo) and I think we should do it.

The reason its not there right now, is because the dBm value comes from those dongles that supply it,
outside the wmbus protocol. Its just a hardcoded print of  a double value created in for example wmbus_iu891a.cc
see rssi_dbm = (int8_t)payload[7]. :-)

There is no dBm unit in the wmbus protocol, yet. But who knows it might get added in the future, you could for example have a wmbus relay send status telegrams about the dBm when receiving from other meters.

kzajac83:
There’s one more inconsistency related to dBm that ties directly into what you wrote. As you mentioned yourself, you “attach” the dBm unit directly to the raw value provided by the dongle/hardware.

While that makes sense and is necessary for wireless wM-Bus, adding an `rssi_dbm `field with a hardcoded value of 0 (so 1mW) for wired M-Bus is incorrect. That was supposed to be the subject of my second change around dBm area, although maybe we can resolve it more easily right here ;)

weetmuts:
I created a new pull request: https://github.com/wmbusmeters/wmbusmeters/pull/1655

Lets continue the discussion there.

Issue #1653, opened 2025-11-27
Kamstrup Multical 303, status/info reporting


jktjkt:
### Type of request

Existing device improvement

### Meter description

We just bought a bunch of Kamstrup Multical 303, the type info as printed on the device is 303W40276. The supplier mounted one of them in a reversed direction, which is shown on the display as info code `0004000`. It appears that this status field is reported as DIF 0x02, VIF 0xFF, VIFE 0x22. The `kamheat` driver currently supports two status fields, one is a 32bit `difvifkey = 04FF22`, the other is `vif_range = ErrorFlags`, and this particualar field is not supported. All the other ones work fine, so thanks for an awesome SW.

I added a new .xmq file with a new field, `difvifkey = 02FF22`, `mask_bits = 0xffff`, and the first 16 bits of status fields. It appears to decode the packet correctly. I wonder how to integrate this with the existing `kamheat.xmq` file, preferably without introducing a new field name and without duplicating the flag definitions. I'm also waiting for a confirmation from our vendor about the expected packet structure, right now I only know that on a working meter, the field is set to `0x0000`, whereas on this one with the reverse flow it's set to `0x0002`, which matches the `WRONG_FLOW_DIRECTION` flag.

### Logged telegrams from meter in decrypted format

```shell
52442d2c1269078340048d2067e08d1620d2ca7804060000000004140000000004ff070000000004ff080000000002592618025d2816023B000002ff220002026c393B440600000000441400000000426c213B
```

weetmuts:
Great work! Hmm, I tried to just add a second match setting. I.e.
```
        field {
            name     = status                                                                                   
            quantity = Text                                                                                     
            info     = 'Status and error flags'
            attributes = STATUS,INCLUDE_TPL_STATUS                                                              
            match {
                difvifkey = 04FF22                                                                              
            }
            match {
                difvifkey = 02FF22                                                                              
            }
```

It actually seems to be working. I get: 
```
058   : 02 dif (16 Bit Integer/Binary Instantaneous value)
059   : FF vif (Manufacturer specific)
060   : 22 vife (?)
061 C!: 0002 ("status":"WRONG_FLOW_DIRECTION")
```

The reason is that the content of the loaded/found wmbus field is translated into a 
int64_t (ie 64 bits) integer, which is then anded with the bit mask, then the bit lookups are performed.
So the 2 byte (16 bit error value) happens to be the lower bits of the long 32 bit error value. So it just works.

[...]

PR #1654, opened 2025-11-27
kamheat: support 16bit status fields

jktjkt:
We just bought a bunch of Kamstrup Multical 303, the type info as printed on the device is 303W40276. The supplier mounted one of them in a reversed direction, which is shown on the display as info code 0004000. It appears that this status field is reported as DIF `0x02`, VIF `0xFF`, VIFE `0x22`, but the `kamheat` driver currently assumes that it's sent as a 32bit `difvifkey = 04FF22` (or as `ErrorFlags), and these are not being sent by our devices.

Even though the fields match (some of) the flags in the 32bit status word, apparently there's no other way to code this in wmbusmeters than to introduce another data field (and have it squashed/injected into the other one, because reporting both `status` and `status16` in case there's been no actual `status` field would be silly, IMHO).

Bug: https://github.com/wmbusmeters/wmbusmeters/issues/1653
Link: https://www.sge.cz/images/MC303/MULTICAL303-Technical-Description-English.pdf


weetmuts:
Thanks!

PR #1655, opened 2025-11-28
Add non-linear dBm unit.

weetmuts:


weetmuts:
This is the first step, to have support for dBm in the calculations and unit tracking of values.

The next step is to fix the rssi_dbm that is currently hardcoded and gives bogus values (0) for mbus telegrams. :-)

Any, the unit can be tested as such:
`./build/wmbusmeters --format=fields --selectfields=foo_dbm,current_power_consumption_kw --calculate_foo_dbm='current_power_consumption_kw' 2D442D2C5768663230028D20E4E2C81C20878C78_04041A03000004843C00000000042B0300000004AB3C00000000 FOO omnipower 32666857 NOKEY`

will print: `34.771213;0.003`
i.e. it successfully converted 0.003 kW into 34.78 dBm.

Scaling with 1000...

`./build/wmbusmeters --format=fields --selectfields=foo_dbm,current_power_consumption_kw --calculate_foo_dbm='1000 factor * current_power_consumption_kw' 2D442D2C5768663230028D20E4E2C81C20878C78_04041A03000004843C00000000042B0300000004AB3C00000000 FOO omnipower 32666857 NOKEY`

`64.771213;0.003`

Only test w to dbm:

`/build/wmbusmeters --format=fields --selectfields=foo_dbm --calculate_foo_dbm='0.0001 w' 2D442D2C5768663230028D20E4E2C81C20878C78_04041A03000004843C00000000042B0300000004AB3C00000000 FOO omnipower 32666857 NOKEY`

`-10`

Only test dbm to w:
`./build/wmbusmeters --format=fields --selectfields=foo_w --calculate_foo_w='0 dbm' 2D442D2C5768663230028D20E4E2C81C20878C78_04041A03000004843C00000000042B0300000004AB3C00000000 FOO omnipower 32666857 NOKEY`

`0.001`

Please test!

weetmuts:
Listing the units print:
```
Power ─────────────────────────────────────────────

w        Watt (W)                         1kgm²s⁻³
mjh      mega Joule per hour (MJ/h) 277.778kgm²s⁻³
m3ch     cubic meter celsius per hour (m³°C/h)3600m³cs⁻¹
kw       kilo Watt (kW)                1000kgm²s⁻³
jh       Joule per hour (J/h)   0.000277778kgm²s⁻³
dbm      decibel-milliwatt (dBm)        NL-kgm²s⁻³
```

Where NL means non-linear.

weetmuts:
@kzajac83  To reflect over Krzysiek's comment:

"dBm as a unit is dimensionless UoM, because dBm is defined in reference to 1 mW (dBµ to microV, dBV to 1V, etc.) . In other words, the numerator 𝑃[mW] divided by the denominator 1 mW gives a dimensionless ratio."

I understand that, strictly speaking, it is dimensionless, but since there is a simple (albeit non-linear) between dBm and W, my preference is that dBm belongs to the Power quantity. This means that it can be converted automatically
between any power value.

I have not implemented casts yet in the formula calculator. Casts could be used to convert any unit into another.
However that is also a risk of course. All the unit tracking in wmbusmeters is because I want  to minimize any bugs caused by bad units. We do not this to happen: https://en.wikipedia.org/wiki/Mars_Climate_Orbiter

weetmuts:
Ah, not all units inside Power can be automatically converted between each other.

There is a m3ch=cubic meter celsius per hour (m³°C/h) which cannot be converted
to W since we do not know the density or specific heat capacity. Even though we can guess
that it is water.

Where did this unit come from? A meter of course with a manufacturer specific field
where the manual said that it is m3c=cubic meter celsius (m³°C), so that unit 
was added, and for this energy m3c I also added power m3ch.

kzajac83:
Thank you Frederic for driving this topic forward. In general I agree with your approach and I will run some tests later when I have more time. I only disagree with the last part of your comment.

> Ah, not all units inside Power can be automatically converted between each other.
> 
> There is a m3ch=cubic meter celsius per hour (m³°C/h) which cannot be converted
> to W since we do not know the density or specific heat capacity. Even though we can guess
> that it is water.
> 
> Where did this unit come from? A meter of course with a manufacturer specific field
> where the manual said that it is m3c=cubic meter celsius (m³°C), so that unit
> was added, and for this energy m3c I also added power m3ch. 


We discussed m3ch about a year ago in the context of the kamheat driver.
I will repeat it here: from a physics point of view, m³°C and m³°C/h are not energy or power units.
They are manufacturer-specific Kamstrup units without real physical meaning, and this is real reason why cannot be converted to standard SI units.

But I don’t think it’s worth going into a long discussion or arguing about it.
The manufacturer (Kamstrup) simply provides this unit, and as a compromise it ended up being supported in wmbusmeters as well :)

weetmuts:
"The manufacturer (Kamstrup) simply provides this unit, and as a compromise it ended up being supported in wmbusmeters as well :)" 

Haha, indeed, we have to put it somehwere. :-) 
 
Still, they are a measurements.... just odd ones, take a cubic meter of water and measure the temperature in Celcsius multiply and you get m³°C. Why not send two separate values m3 and C... who knows.

The SIUnit structure can handle any strange combination of SI units, this is necessary for in-meter calculations.
Wmbusmeters tracks the units through the calculation so for example in driver_em24.cc
There is:

        addNumericFieldWithCalculator(
            "total_apparent_energy_consumption",
            "Calculated: the total apparent energy consumption.",
            DEFAULT_PRINT_PROPERTIES,
            Quantity::Apparent_Energy,
            "sqrt("
            "      (total_energy_consumption_kwh * total_energy_consumption_kwh) "
            "      +"
            "      (total_reactive_energy_consumption_kvarh * total_reactive_energy_consumption_kvarh)"
            "    )");

Before you do the sqrt the SIUnit contains a measurement unit that is sort of useless, it has no mapping to any named unit. However after the sqrt the unit exponents are divided by 2 and out comes an energy compatible unit.

All calculations >must< currently terminate in a named unit (m3, c, datetime, w). If the SIUnit does not map towards a known unit, then the calculation will fail and return NaN or if you are lucky you might get a warning at startup
that the formula cannot be calculated.

Perhaps this hard requirement >must< should be removed? If someone wants to report an oddboll unit in the telegram, why should they not be allowed to? :-) Yay, lets calculate m³cs⁻³kg², lets not restrict the output the named units.
If we can do that m3c is just one of those.

The named units are chosen to be sort of human readable and rememberable, ie we write w and kw
instead of 1kgm²s⁻³ and 1000kgm²s⁻³.

But if the telegram printer detects that the unit does not have a name (ie it is oddboll) then it generates a new ascii unit
after >two< undescores. Like this:

"total_m3": 123  
"total_kw": 123  
"total__m3c": 123  
"bar__m3cs-3kg2": 444   
This is the really oddboll: m³cs⁻³kg² above.

And no, I do not want to put "foo_m³°C" in the key field in json.  Sure it works for json, but that is an ugly artifact of json. 
Sometimes you see that people even store values in the key part in json. The keys should be ascii, since I want to transform the telegram to XMQ and XML. (https://libxmq.org). But since we only use letters, minus and numbers this should work...

Then we have to invent a quantity called ODDBOLL is something... :-)

weetmuts:
SIUnit might not be the best name for the struct, rather UnitBuiltFromFundamentalSIUnits or something shorter but in the same spirit.

kzajac83:
I didn’t realize there was a requirement that every calculation must end with a named unit. At first glance it seems logical, but after thinking about it more, I agree it might be better to drop that restriction. It should be perfectly valid to produce some exotic or even dimensionless indicator if the author of the calculation knows what they’re doing. In that case, why impose limits?

And the double-underscore naming trick is an interesting solution. As for the structure name, as you are owner up to you :)

My test using my telegrams

Power -> dBm

```
krzysiek@wyse:~/wmbusmeters$ ./build/wmbusmeters --format=json --ppjson --calculate_out_dbm='power_kw' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "out_dbm":75.773884,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-11-28T20:03:33Z"
}
```

dBm -> Power

Unfortunelty no [mW] then use [W] UoM

```
krzysiek@wyse:~/wmbusmeters$ ./build/wmbusmeters --format=json --ppjson --calculate___foo_w='90 dbm'  5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "__foo_w":1000000,
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-11-28T20:16:05Z"
}
```

But minus seems not work (or wrong execute)

```
krzysiek@wyse:~/wmbusmeters$ ./build/wmbusmeters --format=json --ppjson --calculate___foo_w='-90 dbm'  5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
wmbusmeters: src/formula.cc:1196: NumericFormula* FormulaImplementation::top2Op(): Assertion `op_stack_.size() > 1' failed.
Przerwane
```

I can calculate "Jh --> W"
```

krzysiek@wyse:~/wmbusmeters$ wmbusmeters --format=json --ppjson --calculate___out_jh='1 w' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "__out_jh":3600,
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-11-28T20:40:28Z"
}
```


But cannot calculate "dBm --> W"

```
krzysiek@wyse:~/wmbusmeters$ wmbusmeters --format=json --ppjson --calculate___out_dbm='1 w' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
Could not extract a valid unit from calculated field name __out_dbm
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-11-28T20:40:37Z"
}
```

pswiatki:
> I understand that, strictly speaking, it is dimensionless, but since there is a simple (albeit non-linear) between dBm and W, my preference is that dBm belongs to the Power quantity. This means that it can be converted automatically
> between any power value.

I reflect on this as well. I posit, dBm is the unit of power, why not. It would be a multiplier (X) of 1 mW. This multiplier can be larger that one, or smaller. Sooo.....   dB is dimensionless, yes. But we have this "m" in addition. I think this gives the unit (of dBm) the dimension.

weetmuts:
@kzajac83 You accidentally put some extra underscores:
`--calculate___out_dbm`
should be
`--calculate_out_dbm`

The conversions seem to work, however the formula parser cannot handle unary negative values yet. 
I am going to fix that, but that is independent from the unit conversion.

weetmuts:
@kzajac83 Forget what I said about the underscores, I just tested with the latest build of the branch and your example works fine... did you perhaps run another version of wmbusmeters?

```
./build/wmbusmeters --format=json --ppjson --calculate___out_dbm='1 w' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "__out_dbm":30,
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-11-30T10:24:40Z"
}
```

weetmuts:
> In that case, why impose limits?

The (w)mbus protocol tracks the units, so my intent is that wmbusmeters should track and enforce the units.
Unit confusion can happen elsewhere, but not on my watch. :-)

Now, I did not expect a meter to send m3c.... :-O so our discussion regards this...thing of a unit... is very valuable.
Nor do I expect anyone to calculate half-baked values (the SI-units do not match any known named unit) for output.

The double underscore unit solution (present SI-Unit combos that has no known name, W,kW etc after double underscores) would solve both problems.

Ie we can remove m3c as a named unit but it can still be presented to the end user using the SI-Unit combo with no known nam. It would no longer be listed in the --listunits list.

weetmuts:
Ok, managed to find the Kamstrup 303 meter manual, the m3c fields are explained like this:

7.1.2 Energy registers E8 and E9
E8 and E9 are used as a basis for calculating the volume-based average temperatures of inlet and outlet,
respectively. With every volume count (every 0,01 m 3 for q p 1,5 m 3
/h with CCC=419), the registers are accumulated
by the product of m 3 x °C, which makes E8 and E9 a suitable basis for calculating the volume-based average
temperatures.
E8 and E9 can be used for average calculation during any period of time as long as the volume register is read at the
same time as E8 and E9.

So, yeah they call them Energy registers, but they are used to calculate volume based average temperatures.

weetmuts:
After some more researching, as for the dimensionless of dBm and a lot of other logarithmic units, they all seem to be strictly dimensionless, but they all carry embedded reference values 

1 mW for dBm, 1 mol/L for pH,  0.775V RMS for dBu, etc etc)

Thus the current implementation seems to do the right thing, it encodes the reference value unit in the SI exponents and scale_ then it applies the non-linear transform back and forth when necessary to get to/from the reference unit.

kzajac83:
You have right, partially I use old (my working prod) wmbusmeters (path without ./build. prefix) :(


Later end of day I will try make new test....

kzajac83:
> @kzajac83 Forget what I said about the underscores, I just tested with the latest build of the branch and your example works fine... did you perhaps run another version of wmbusmeters?

Yes, I make mistake :( due to I downloaded last PR and yes "Power-> dBm" working fine.

```
krzysiek@wyse:~/wmbusmeters$ ./build/wmbusmeters --format=json --ppjson --calculate___out_dbm='0.1 w' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "__out_dbm":20,
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-12-02T00:44:49Z"
}
```


```
krzysiek@wyse:~/wmbusmeters$ ./build/wmbusmeters --format=json --ppjson --calculate___out_dbm='0.0001 w' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "__out_dbm":-10,
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-12-02T00:45:20Z"
}
```

```
krzysiek@wyse:~/wmbusmeters$ ./build/wmbusmeters --format=json --ppjson --calculate___out_dbm='0.0036 jh' 5e4424237442224653047af000500558463B5890c95e687730039B7a601f49e69e539a8B88aa81B13f172353fBB67e4ac0c2346cff3ddaf7B390543B75e218cdd507df5e764c8756d115a1e0e9418cdf4a74dBf544fB96c91aa0d6BB5ae8eB FOO sharky 46224274 NOKEY
{
    "_":"telegram",
    "media":"heat",
    "meter":"sharky",
    "name":"FOO",
    "id":"46224274",
    "__out_dbm":-30,
    "flow_temperature_c":82.6,
    "operating_time_h":15480,
    "power_kw":37.791,
    "return_temperature_c":56.4,
    "target_date":"2024-12-31",
    "target_energy_consumption_kwh":189341.388889,
    "target_volume_m3":7319.1,
    "total_energy_consumption_kwh":191005.277778,
    "total_volume_m3":7370.3,
    "volume_flow_m3h":1.27,
    "timestamp":"2025-12-02T00:49:03Z"
}
```

kzajac83:
>Unit confusion can happen elsewhere, but not on my watch. :-)

hehe :)

Regarding the “m3c” UoM I only wanted to point out that it’s not a physical power unit, and “m3ch” is not really an energy unit either (power over time becomes energy, hence the “h”). If this exotic UoM works as intended, I see two options:

- if it works reliably, just leave it as is
- or move “m3c” out of the Power dimension, and “m3ch” out of the Energy dimension into something more appropriate dimension
 

> Thus the current implementation seems to do the right thing, it encodes the reference value unit in the SI exponents and scale_ then it applies the non-linear transform back and forth when necessary to get to/from the reference unit.
 
Agree, however I can confirm only dBm ;)

PR #1657, opened 2025-12-04
Non standard baud rate support for rc1180

karol57:
I needed to increase baud rate of rc1180 for troubleshooting.
Didn't helped much, but I would prefer to leave mine at 230400.
Also, maybe will be useful for someone else :)


weetmuts:
Great work! I cannot test it though, my rc1180 is in a box somewhere. :-)

PR #1666, opened 2025-12-13
read libxml2-include from pkg-config

prauscher:
Using pkg-config is more flexible and required when building for nix.


weetmuts:
Thanks!

weetmuts:
Alas! The build immediately failed on macos and some dockers. I had to restore it temporarily
and will then move the location to configure.ac. A stub is there.

PR #1678, opened 2026-01-06
fix(amb8465): prevent false protocol errors on RF noise/rx timeout

### LINKED ISSUE #1677: amb8465: Device hangs requiring physical reboot after repeated RX timeouts
### Type of request

Bug

### OS version

Debian GNU/Linux 12 (bookworm)

### wmbusmeters version

1.19.0

### Your message goes here

**Issue description**
I am operating an AMB8665-M (in wmbusmeters identified as amb8465) in a busy/noisy environment (city with lots of devices around sending data).  After a few days, wmbusmeters stops working until I do a complete reset of the OS. In the log I see frequent entries like this:
`too many protocol errors(20) resetting /dev/ttyUSB0 amb8465`

Just a restart of wmbusmeters does not recover, I see
`Expected amb8465 on /dev/ttyUSB0 but did not find it! Ignoring tty!`

In verbose mode I see a lot of RX timeouts (`rx long delay`), leading to repeated attempts to reset the serial device. After a sequence of these errors and resets, the USB dongle eventually stops responding completely. 

**Expected behavior**
The daemon should handle RX timeouts and partial telegrams gracefully, resetting the device if necessary without causing a permanent hardware/driver lockup.

**Proposed solution**
I have written a small fix to prevent the frequent resets and tested it. Will link the PR here. However, this does not address why the stick becomes unresponsive after the frequent resets.

Tom1000:
**Problem**
High RF noise causes frequent `rx long delay` events. The current implementation interprets 20 of such subsequent noise events as an error which trigger a reset of the device.

**Solution**
Improved the AMB8465 error handling to avoid aggressive resets on partial frames.

This partially fixes #1677

### DIFF
--- a/src/wmbus_amb8465.cc
+++ b/src/wmbus_amb8465.cc
@@ -752,8 +752,15 @@ void WMBusAmber::processSerialData()
         if (chunk_time.tv_sec >= 2)
         {
             verbose("(amb8465) rx long delay (%lds), drop incomplete telegram\n", chunk_time.tv_sec);
-            read_buffer_.clear();
-            protocolErrorDetected();
+
+            // Only trigger a protocol error if we were receiving a specific command response 
+            // from the stick (starts with AMBER_SERIAL_SOF).
+            if (read_buffer_.size() > 0 && read_buffer_[0] == AMBER_SERIAL_SOF)
+            {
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks!

PR #1679, opened 2026-01-07
Add stdin:json input format for per-telegram decryption keys

blavka:
Allows passing telegram data as JSON with individual decryption keys, enabling dynamic key management without pre-configured meters.

Input format: {"telegram": "HEX", "key": "KEY", "driver": "auto", "format": "wmbus|mbus"}

Supports both WMBUS and MBUS telegrams with auto-detection.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,6 +189,7 @@ PROG_OBJS:=\
 	$(BUILD)/wmbus_rtl433.o \
 	$(BUILD)/wmbus_simulator.o \
 	$(BUILD)/wmbus_rawtty.o \
+	$(BUILD)/wmbus_jsontty.o \
 	$(BUILD)/wmbus_rc1180.o \
 	$(BUILD)/wmbus_utils.o \
 	$(BUILD)/xmq.o \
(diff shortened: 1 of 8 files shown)


weetmuts:
Great work! I will take a look!

weetmuts:
Very nice indeed! Great work at reverse engineering the software to add the meters etc! 

I change the name of the format from tty:json to tty:jsonl since it is actually reading json lines.
Otherwise users might push a pretty printed json into it and it will fail.

I added a type name for the request json object so it looks like this:
{"_":"decode", "telegram":"....", "key":"..."}
We might want to have other requests in the future.

I also used the existing json parser inside wmbusmeters (part of the xmq.h xmq.c from https//libxmq.org)
The fields are then extracted using `hex_s = xmqGetString(doc, "/decode/telegram");`
no need for hardcoded json scanning. Better if we want to add more features.

As a side effect it also accepts xml and xmq. Ie you can do:
`<decode><telegram>...</telegram><key>...</key></decode>`
or 
`decode{telegram=... key=...}`

Not that I expect anyone to use this in the near future! :-)
stdin:xmll and stdin:xmql are there for symmetry, but it is actually auto-detecting the input format.

blavka:
Thank you,
yes, that's right, it's jsonl.
I'd love to see that improvement.

weetmuts:
@blavka Yep, your code and the improvement is in place in the new release 1.20 version of wmbusmeers.
Please test!

PR #1683, opened 2026-01-10
* add backwardflow_at_set_date to waterstarm

ulda:
I bought a branded waterstarm from brummerhoop.de.
it has the backward flow field in the packet, so I copied an apropriat entry to it.


weetmuts:
Many thanks! I just change the field name to total_backwards_at_set_date_m3 to be more consistent.

PR #1697, opened 2026-02-01
Add --decodingserver=PORT for TCP-based telegram decoding

amirfuhrmann:
## Summary

- Extract decode logic from `wmbus_xmqtty.c` into a reusable `decode.h`/`decode.cc` module shared by both `stdin:jsonl` and the new TCP server
- Add `--decodingserver=PORT` flag that starts a single-threaded `select()`-based TCP server accepting JSON, XMQ, or XML decode requests (one per line) and returning compact JSON Lines responses
- Each client gets its own `DecoderSession` with independent meter caching, non-blocking I/O, IPv6 dual-stack socket, and graceful shutdown on SIGTERM/SIGINT
- Existing `stdin:jsonl` behavior is unchanged

## New files

| File | Description |
|------|-------------|
| `src/decode.h` / `src/decode.cc` | Shared decode module (extracted from `wmbus_xmqtty.c`) |
| `src/decoding_server.h` / `src/decoding_server.cc` | TCP server implementation |
| `examples/decoding_client.py` | Sample Python client (one-shot and persistent session) |
| `Dockerfile.dev` | Alpine-based build/test environment |

## Usage

```bash
wmbusmeters --decodingserver=12345
```

```bash
echo '{"_":"decode","telegram":"1844AE4C4455223368077A55000000041389E20100023B0000","key":"NOKEY","driver":"iperl"}' | nc localhost 12345
# → {"_":"telegram","media":"water","meter":"iperl","name":"","id":"33225544","max_flow_m3h":0,"total_m3":123.529,...}
```

## Test plan

- [x] Docker build passes (`make && make test` — all existing tests pass)
- [x] Manual server test: single telegram decode returns correct JSON
- [x] Error handling: malformed JSON, missing fields return JSON error responses
- [x] Persistent connections: multiple requests over one TCP session work correctly
- [x] Python client: both one-shot and session modes verified against server

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,33 @@
+FROM alpine AS build
+RUN apk add --no-cache alpine-sdk gcc linux-headers libxml2-dev libxslt-dev cmake libusb-dev bash samurai jq netcat-openbsd python3
+
+# Build librtlsdr from source (required dependency)
+RUN git clone --depth 1 https://github.com/steve-m/librtlsdr.git /librtlsdr
+WORKDIR /librtlsdr
+RUN cmake -B build -G Ninja \
+        -DCMAKE_BUILD_TYPE=MinSizeRel \
+        -DCMAKE_INSTALL_PREFIX:PATH=/usr \
+        -DDETACH_KERNEL_DRIVER=ON \
+        -Wno-dev && \
+    cmake --build build -j `nproc` && \
+    cmake --install build
+
+# Build xmq tool (needed for tests)
+RUN git clone --depth 1 https://github.com/libxmq/xmq.git /xmq
+WORKDIR /xmq
+RUN ./configure && make -j`nproc`
+
+# Copy local source
+COPY . /wmbusmeters
+WORKDIR /wmbusmeters
+
+# Clean any host build artifacts then prepare build dir
+RUN rm -rf build build_debug build_arm && \
+    mkdir -p build && \
+    cp /xmq/build/default/release/xmq build/xmq
+
+# Build wmbusmeters
+RUN make -j`nproc`
+
+# Run tests
+RUN make test
--- a/Makefile
+++ b/Makefile
@@ -164,6 +164,8 @@ PROG_OBJS:=\
 	$(BUILD)/bus.o \
 	$(BUILD)/cmdline.o \
 	$(BUILD)/config.o \
+	$(BUILD)/decode.o \
+	$(BUILD)/decoding_server.o \
 	$(BUILD)/drivers.o \
 	$(BUILD)/dvparser.o \
 	$(BUILD)/formula.o \
--- a/examples/decoding_client.py
+++ b/examples/decoding_client.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+"""
+Example client for the wmbusmeters decoding server.
+
+Start the server:
+    wmbusmeters --decodingserver=12345
+
+Then run this script:
+    python3 decoding_client.py
+
+Protocol: JSON Lines over TCP — one JSON request per line, one JSON response per line.
+"""
+
+import json
+import socket
+
+
+def decode_telegram(host, port, telegram_hex, key="NOKEY", driver="auto", fmt=None):
+    """Send a single decode request and return the parsed JSON response."""
+    request = {
+        "_": "decode",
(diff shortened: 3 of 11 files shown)


weetmuts:
Nice, but I am a bit hesitant to open up a TCP server without encryption/protocol/safety etc. 

What is your use case? I would assume that everyone wants to use curl to interact with it, which  is not possible with a simple TCP server.

amirfuhrmann:
> Nice, but I am a bit hesitant to open up a TCP server without encryption/protocol/safety etc.
> 
> What is your use case? I would assume that everyone wants to use curl to interact with it, which is not possible with a simple TCP server.

To describe the use case, i have a very big application written in python, which receives telegrams, and needs to decode them, it has all the keys needed as well. So i need to have a service to decode them, locally, i can do a more tight integration python <-> C, but it's quite a bit of work. and not very good idea architecturally anyway. 

So my use case is entirely local, and encryption isn't required, just a waste of CPU time, we can limit the scope to listening on localhost only (127.0.0.1) not 0.0.0.0. I just need decoding, not any meter reading. Write now i am using wbusmeters through a tiny fastapi that is spawning an--analyze ew process for every read, but it's not very efficient CPU usage.

If you have an idea for a better approach, or modified approach, or any modifications to this appoach. i am very interested. 

Thanks for you consideration

weetmuts:
Yes, I see. First do not use analyze! This tries all drivers on all telegrams! It wastes a lot of cpu. That is why it is called analyze and the purpose is to understand the telegram content to develop and find the correct driver. Not to use in production. :-)

You can reduce the CPU usage by using the standard invocation:
`wmbusmeters --format=json <hex> <name> <driver> <id> <key>`
for example
`wmbusmeters --format=json <hex> MyName auto ANYID <key>`
or replace auto with the proper driver. This will only load a single driver per telegram.

But I do see the value of having a wmbusmeters daemon running and then passing requests to it
where the meters are dynamically added, like the jsonll batch input, and the output is not sent
to the standard output/printer but instead back to the requester.

The device language for specifying device is pretty advanced. We can say:
`auto` (to scan for serial devices)
`im871a` (to scan serial devices but only probe for im871a)
`/dev/ttyUSB1:im871a` (to probe for the im871a only on /dev/ttyUSB1)
`rtlwmbus[123123]`   (use librtlsdr to find this particular rtlsdr dongle with the id 123123)
`rtlwmbus[123123](ppm=17)` (use the choosen dongle and pass the ppm parameter to the device)
`rtlwmbus:CMD(...)` feed the rtlwmbus driver from a command typicall the rtl_sdr|rtl_wmbus combo.
But CMD is single direction, it sends telegrams to wmbusmeters but you cannot pass the output to the cmd.
The first device /dev/ttyUSB0) is assumed to be a serial device, if present. 

To differentiate a serial device from a unix socket, we could add: 
`SOCKET(/var/run/wmbusmeters/wmbusmeters.sock)`
and for symmetry:
`SERIAL(/dev/ttyUSB0)` which just happens to be the default to be backwards compatible.

This would trigger wmbusmeters to create this unix socket and connect it to the existing select in main.
This device would also notify the printer that it should send the output back to the socket.

You could start this with:
`wmbusmeters SOCKET(/var/run/wmbusmeters/wmbusmeters.sock):jsonll`
or from the config file:
`device=SOCKET(/var/run/wmbusmeters/wmbusmeters.sock):jsonll`

Then you can open the unix socket directly from python, or you can open a tcp port using socat:
`socat TCP-LISTEN:9000,fork UNIX-CONNECT:/var/run/wmbusmeters/wmbusmeters.sock`

Ping @BIBOLV @kzajac83 @pswiatki

weetmuts:
With such a socket we should also be able to decode using a single shell command:
`echo '{"_":"decode","telegram":"..."}' | socat - UNIX-CONNECT:/var/run/wmbusmeters/wmbusmeters.sock`

This should be a test case.

amirfuhrmann:
thanks for these suggestions, let me try and work it out. I'll try to update the PR tonight or tomorrow

amirfuhrmann:
i reworked to comply with your approach, i can eliminate the other PR, this one works well for me

amirfuhrmann:
precursor  to #1704

Issue #1533, opened 2025-05-30
NEW meter: id: 87218402 mfct: (DME) DIEHL Metering, Germany (0x11a5) type: Water meter (0x07) ver: 0x63


brenkem:
### Type of request

New device support

### Meter description

The meter Type 173 from "Diehl Meetering" seems not to be supported for now. Please advise how to enable support.

```
# wmbusmeters --logtelegrams --format=json auto:c1 watermeter unknown 87218402 NOKEY
Started auto cul on /dev/ttyUSB0 listening on c1
(meter) watermeter: meter detection did not match the selected driver unknown! correct driver is: unknown!
(meter) Not printing this warning again for id: 87218402 mfct: (DME) DIEHL Metering, Germany (0x11a5) type: Water meter (0x07) ver: 0x63
(meter) please consider opening an issue at https://github.com/wmbusmeters/wmbusmeters/
(meter) to add support for this unknown mfct,media,version combination
(meter) newly created meter (watermeter 87218402 unknown) did not handle telegram!

```

```
# wmbusmeters auto:c1
Started auto cul on /dev/ttyUSB0 listening on c1
No meters configured. Printing id:s of all telegrams heard!
Received telegram from: 87218402
          manufacturer: (DME) DIEHL Metering, Germany (0x11a5)
                  type: Water meter (0x07) encrypted
                   ver: 0x63
                device: cul
                  rssi: -47 dBm
                driver: unknown!
```

### Logged telegrams from meter in decrypted format

```shell
(serial) received binary "6236333434413531313032383432313837363330373335433638433030393939"
(cul) checkCULFrame "b6344A51102842187630735C68C00999"
(cul) no eol found yet, partial frame
(serial) received binary "3030463030324332353341314530313030334435334345343832433735393132"
(cul) checkCULFrame "b6344A51102842187630735C68C0099900F002C253A1E01003D53CE482C75912"
(cul) no eol found yet, partial frame
(serial) received binary "3846313237374143323030343130373130374131333434343342303837423633"
(cul) checkCULFrame "b6344A51102842187630735C68C0099900F002C253A1E01003D53CE482C759128F1277AC2004107107A134443B087B63"
(cul) no eol found yet, partial frame
(serial) received binary "3241373239313842443045323745413436323930313737303030394545393235"
[...]

PR #1708, opened 2026-02-04
Add support for Diehl Type 173 variant

gladk:
This change adds the support for (DME) DIEHL Metering, Germany (0x11a5) type: Water meter (0x07) ver: 0x63. Issue #1533 

I tested it and it works just fine:

```
{
  "_": "telegram",
  "media": "water",
  "meter": "evo868",
  "name": "watermeter",
  "id": "11111118",
  "consumption_at_history_1_m3": 4.674,
  "consumption_at_set_date_m3": 0.2,
  "history_1_date": "1970-01-01",
  "total_m3": 6.592,
  "current_status": "OK",
  "set_date": "2025-12-31",
  "timestamp": "2026-02-04T19:36:59Z",
  "device": "rtlwmbus[00000001]",
  "rssi_dbm": 141
}
```

### DIFF
--- a/src/driver_evo868.cc
+++ b/src/driver_evo868.cc
@@ -34,6 +34,7 @@ namespace
         di.addMVT(MANUFACTURER_MAD,  0x06,  0x50);
         di.addMVT(MANUFACTURER_MAD,  0x07,  0x50);
         di.addMVT(MANUFACTURER_MAD,  0x16,  0x50);
+        di.addMVT(MANUFACTURER_DME,  0x07,  0x63); // Diehl Type 173 variant reusing evo868 payload
 
     });
 


weetmuts:
Nice! Do you have a telegram that we can add to the regression tests? You can get a decrypted telegram
with --logtelegrams

eddyjay:
What is missing for this to be merged onto main?

weetmuts:
I got the telegram from @gladk thanks! It turns out that we should create a new driver for this meter.
The evo868 is for meters from another manufacturer (MAD) and tries to decode a lot of other fields that the Diehl meter does not have, and there are fields in the Diehl meter that does not exist in the old evo868 meter.

So, i did just that, check it out here:
https://wmbusmeters.org/drivers/dme173.xmq.html

If you are running version 2.0 of wmbusmeters you can use this new driver directly, without recompiling.
Just use the name @dme173 (the @ sign will trigger a download of the new xmq file automatically
and store it locally.)
It is now also part of the main source and compiled into wmbusmeters if you build from source as normal.

Now, the interesting part is that there are many recurring fields in wmbus water meters, 
the most common is Volume+Unit+Data encoding wmbusmeters has library for matching those
just write `use = total_m3` and all meters that send some sort of volume will be normalized into m3
despite the meter sending litres or cubic feet (yeah it exists in the standard) or if the meter uses BCD or
binary or float etc..

So almost all water meter drivers can be used to extract this. Just  iperl and you will see the total_m3.

Now, Volume in a storage slot, typically 1, means the recorded meter volume at the end of the last billing period, 
typically the end of last month. This way the water company can drive around and pick up the previous month consumption easily, it does not matter when they do it during the month.

This is such a recurring concect, these values are called target_m3 or target_kwh etc.
The library rule assumes the value is in storage nr 1.

But in this meter, it looks like they are using storage nr 8... so I could not re-use the target_m3 library rule.
Then in storage nr 1 they have something that could be the end of last year? Or the second last month?

Can you check this for me?

Then we lack understanding of the error flags. If a bit gets set, it will be printedin the status like ERROR_01
but it is nicer to know what it means and you do that by mapping the bit to a text like BURST.

Then there is an unknown field that needs to be understood.

Please check the decoding here:
https://wmbusmeters.org/analyze/5344A5113322332263078C0085900F002C256C1601004BB24778D6AC091E7ADF003107102F2F0C138386000004FD17000000000A5A550002FD74EE114C1300020000426C3F3C84046D3B375F318C041374460000

weetmuts:
@eddyjay wrote in an email: > It's funny that I have the same device, made in January 2026, and just recently installed, but the encoding in this one is slightly different. Is it something worth investigating?

Yep, your telegram has backwards flow in it, this is typical for modern wmbus meters, the are highly configurable.
Depending on the manufacturer/purchaser settings, the meter can send all sorts of different data.
This is why the wmbusmeters drivers can be so complicated.

In your case the new field was in the library, so I just added `use = total_backward_m3` to the dme173.xmq file

Why not try all the wmbusmeters library settings everytime for each telegram? Because, it slows things down, and you are not sure that the manufacturer uses the fields like all other normal manufacturers do....

weetmuts:
@eddyjay And your telegram lacks the flow temperature, so it just disappeared from the json for your meter.

gladk:
Thanks @weetmuts for analysis! I have just compiled the master branch it it just works out-of-box, adding flow_temperature_c in my case.

What I am missing from the master branch is the auto-detection of the driver. With "auto" in the command line, i dont get the flow-values. Only id, timestamp etc.

Should I add some more parameters into the command line, only auto is not enough?

Basically, this MR can be closed if the new driver is already in master branch.

PR #1717, opened 2026-02-11
Add native kaden driver for Kaden C10/D10 HCA meters

Hymbajs:
## Summary

- Add native `kaden` driver for Kaden C10/D10 heat cost allocators (VIP manufacturer, type 0x08, version 0x1E)
- Full field extraction: status flags, fabrication_no, current/historical HCA consumption, room/radiator temperatures, actuality duration, set_date (DD-MM format), and tampering duration
- Add `strdaymonth()` utility function for DD-MM date formatting (used by set_date field, which encodes only day-month for annual billing date)
- Includes tests for both D10 and C10 models

## Fields extracted

| Field | Description |
|-------|-------------|
| `status` | Error flags (OK, VOLTAGE_INTERRUPTED, SENSOR_T1/T2/T3_OUTSIDE_MEASURING_RANGE) |
| `fabrication_no` | Meter serial number |
| `current_consumption_hca` | Current HCA value |
| `consumption_at_set_date_hca` | HCA at last billing date |
| `room_temperature_c` | Corrected room temperature (external temp VIF) |
| `radiator_temperature_c` | Radiator surface temperature (flow temp VIF) |
| `actuality_duration_min` | Time without measurement in current period |
| `actuality_duration_at_set_date_min` | Time without measurement at billing date |
| `set_date` | Annual billing date in DD-MM format (e.g. `01-01`) |
| `meter_datetime` | Current meter date/time |
| `tampering_duration_s` | Total time device removed from radiator |

## Test plan

- [x] `testinternals` passes with embedded D10 and C10 test telegrams
- [x] Verified against 10 random real-world C10 meters — all values physically consistent
- [x] `test.sh build/wmbusmeters` — all tests pass ("All tests ok\!")

### DIFF
--- a/src/driver_kaden.cc
+++ b/src/driver_kaden.cc
@@ -1,0 +1,200 @@
+/*
+ Copyright (C) 2024 Fredrik Öhrström (gpl-3.0-or-later)
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include"meters_common_implementation.h"
+
+namespace
+{
+    struct Driver : public virtual MeterCommonImplementation
+    {
+        Driver(MeterInfo &mi, DriverInfo &di);
+    private:
+        void processContent(Telegram *t);
+    };
+
+    static bool ok = staticRegisterDriver([](DriverInfo&di)
+    {
+        di.setName("kaden");
+        di.setDefaultFields("name,id,current_consumption_hca,set_date,consumption_at_set_date_hca,status,timestamp");
+        di.setMeterType(MeterType::HeatCostAllocationMeter);
+        di.addLinkMode(LinkMode::T1);
+        di.addMVT(MANUFACTURER_VIP, 0x08, 0x1E);
+        di.usesProcessContent();
+        di.setConstructor([](MeterInfo& mi, DriverInfo& di){ return shared_ptr<Meter>(new Driver(mi, di)); });
+    });
+
+    Driver::Driver(MeterInfo &mi, DriverInfo &di) : MeterCommonImplementation(mi, di)
+    {
+        addStringFieldWithExtractorAndLookup(
+            "status",
+            "Meter status from error flags and tpl status field.",
+            DEFAULT_PRINT_PROPERTIES |
+            PrintProperty::STATUS | PrintProperty::INCLUDE_TPL_STATUS,
+            FieldMatcher::build()
+            .set(MeasurementType::Instantaneous)
+            .set(VIFRange::ErrorFlags),
+            Translate::Lookup(
+            {
+                {
+                    {
+                        "ERROR_FLAGS",
+                        Translate::MapType::BitToString,
+                        AlwaysTrigger, MaskBits(0xffff),
+                        "OK",
+                        {
+                            { 0x0001, "VOLTAGE_INTERRUPTED" },
+                            { 0x0004, "SENSOR_T2_OUTSIDE_MEASURING_RANGE" },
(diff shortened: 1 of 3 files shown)


Hymbajs:
## Test telegrams

```bash
# Kaden D10 (ID: 23800604)
echo '4e443059040680231e087ac40040050e6aa476257c0c3adae8277edc999b39b38222fcb387a91e94cb6ed47ceec6470f5f686f89a8574415fa262bd43c88f7f153ce3c66e8e44da338a06c62ab21b1' | ./build/wmbusmeters --format=json stdin:hex KadenD10 kaden 23800604 82B0551191F51D66EFCDAB8967452301

# Kaden C10 (ID: 23701267)
echo '4e443059671270231e087a6f0040051183e178afb952391c01f78104bc3fd33a5232ba7e70f514a062dac99059a7c74a55227dfae9d9590145f685f4ae6a62288bbba6eaf92d86797254644a2cdf46' | ./build/wmbusmeters --format=json stdin:hex KadenC10 kaden 23701267 82B0551191F51D66EFCDAB8967452301
```

weetmuts:
Nice, but did you use Claude to generate this pull request? The pull request summary looks like other Claude generated summaries. If it was the AI, please tell it that drivers should no longer be written as native drivers, they should be written as xmq text files. :-)

I wrote the xmq for you. You can see it here: https://wmbusmeters.org/drivers/kaden.xmq.html

A few questions, are you sure the status bits are in that order?

The meter use actuality duration in a creative way. I tried to improve the naming of the fields and the descriptions.
Can you double check that I understood you correctly?

weetmuts:
Oh, and the year removal from the target_date field is odd. Does the meter always send a bogus year?

The ability to hard-code a string into a text field that should have a proper unit, is an old artifact that will go away.

wmbusmeters takes the units (suffixes such as _date and _h and _c) very seriously. 
The modified string without the year is not a valid _date. 

If I understand why the year should be ignored, then we can figure out the proper solution. :-)

Anyway thanks for the work!

Hymbajs:
Yes, Claude was used — thanks for writing the XMQ driver, it's much cleaner!

**Status bits:** After checking the Kaden C10 user manual (PDF), I found that **bit 3 (0x0008) is actually TAMPERING** (device removed from radiator), not `SENSOR_T1_OUTSIDE_MEASURING_RANGE`. The manual states: *"v registru F.F je zaznamenána chyba tampering (kód 8)"*. This matches the D10 test data — status `0x0008` with radiator temp ≈ room temp (19.7°C vs 19.6°C), meaning the device is not mounted on an active radiator.

The other bit definitions (0x0001 VOLTAGE_INTERRUPTED, 0x0004 SENSOR_T2, 0x0020 SENSOR_T3) are not documented in the manual — only bit 3 is confirmed. I will contact Kaden directly to ask for a complete specification of the F.F error register bits.

**Actuality duration:** Your renaming to `no_measurement_duration` / `no_measurement_duration_last_year` / `tampering_duration` with hours display unit is better — confirmed correct. The Kaden manual describes these as "Aktuální/Minulá doba bez náměru v topném období" (time without measurement during heating season).

I'll close this PR since your XMQ version supersedes it. However, would it be possible to merge the basic kaden XMQ driver soon? We have a number of Kaden devices deployed and would appreciate having at least basic decoding available in a release.

Hymbajs:
The `target_date` field represents the **billing date** ("zúčtovací den") — an annual recurring date when HCA values are settled and reset. Since it repeats every year, **the year has no meaning**. The Kaden manual defines register C.51 as just "month + day" (2-byte CP16 format). However, the wM-Bus Date type G encoding forces a year to be included, so Kaden fills the year bits with all 1s, resulting in 2127.

We can keep the full date including the year for standard compliance, but it is rather confusing for the end user. In my native driver I stripped the year to DD-MM for this reason — but I understand that breaks the `_date` suffix semantics in wmbusmeters.

I will contact the manufacturer directly to ask for a more detailed specification of the wM-Bus data frame.

weetmuts:
Thanks!

Just curious, the meaning of "Aktuální/Minulá doba bez náměru v topném období" (time without measurement during heating season).

Is this independent of tampering? I.e. can the meter detect no-measurements without the tampering bit being activated? 

There are other units such as _txt for plain text, we could place 01-01 there using the calculate function.
But the calculate function in xmq does not have support for string manipulations nor the conversion of
dates into strings. When I add that, we could create the target_txt field.
Or we could split it into target_month and target_d which are existing units.

I have other more pressing things to do, so this is in the future.
Lets have target_date now with a info text that says:
`Yearly billing starts and ends on the beginning of this month/day. Ignore the year. This field will be replaced with target_txt in the future.`

PR #1739, opened 2026-03-03
Added support for IMST iU893A-XL

icho911:
This change adds support for newly released USB-Stick IMST iU893A-XL. Only added the new mode_type.

I tested it and can now receive telegrams. The stick is recognized as iU891A.


weetmuts:
Thanks!

PR #1741, opened 2026-03-04
Feature/add json tty support

hollie:
# Summary 
Adds support for a wMBus receiver that outputs received telegrams in json format.

The receiver is connected over a USB virtual COM port and outputs data in the format:

```
{ 
"timestamp" : 784647039,  
"wMBus" : { 
  "L" : 47, 
  "C" : "0x44",
  "M" : "KAW",
  "ID" : 12756012 ,
  "version" : "0x3C",
  "type" : "0x16" ,
  "rssi": -85,
  "raw" :  "2F44372C126075123C168D20E4025B6C216A35FF2DFA193287FF280EEA356FEF99D45A8FA487B66BD3612FDA9A2287CD" } 
}
```

The required configuration entry looks like this:

`device=/dev/serial/by-id/<your-device>-if00:jsontty` 

# Changes
Adds `wmbus_jsontty.cc` and updated the code to include it as an option when configuring a wMBus receiver.

# Test plan
- [x] make runs without warnings
- [x] I have been testing this successfully for over one month on an raspberry pi


weetmuts:
Ah, no that is not the way to do it. There is already jsonl support for decoding and parsing with a separate key.

What you want is just feeding telegrams into wmbusmeters and use the normal decoding with meters defined in meter files. There are several formats that are suitable. Raw, hex, rtlwmbus. 

Also, why not search for jsonl and how it is implemented already using the xmq parsing library to avoid reimplementing json parsing. Talk to claude about this....

Also, those test plans added by claude are truly worthless. How about actually adding a test?

This is how you should do it, without modifying wmbusmeters:

device=hex:CMD(socat -u /dev/serial/yourdevice,b115200,raw,echo=0 STDOUT | jq -r .wMBus.raw)

If your serial port is already setup correctly, just "cat /dev/serial/yourdevice" instead of socat might also work.

hollie:
Hey @weetmuts,

thanks for the heads-up and the pointer. I'll update my test setup to implement the solution you propose.

Best regards,
 Hollie.

weetmuts:
Good luck! To be able to use the wmbusmeters meter setup and pass telegrams using json is on the TODO list.
But i will modify the current stdin:jsonl input for that.

hollie:
Hey @weetmuts 

I have it up and running with an intermediate helper script to fix the raw wmbus output. My receiver reports the CRC of the raw packet and I need to strip that from the hex string and update the field length. Other than that all is working as expected with the stock wmbusmeters code :+1:

I am noting down my config below as reference for others:

Relevant device config line in wmbusmeters.conf:

`device=hex:CMD(/usr/local/bin/wmbusreader.sh)`

The script that is called looks like this:

```
#!/bin/bash

# Explicitly set the PATH so the background service can find jq, grep, and stty
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

# Define the serial port
SERIAL="/dev/ttyACM0"

# Configure the serial port
stty -F $SERIAL 115200 raw -echo

# Wrap everything in an infinite loop so the daemon never exits
while true; do
    
    # Read, filter, extract, and mathematically correct the hex string
    cat "$SERIAL" | grep --line-buffered "wMBus" | jq --unbuffered -r '.wMBus.raw' | while read -r rawhex; do
        
        # Check if we got a valid string length before processing
        if [[ ${#rawhex} -gt 6 ]]; then
            # 1. Extract the first 2 characters (L-field) and convert hex to decimal
            L_hex="${rawhex:0:2}"
            L_dec=$((16#$L_hex))

            # 2. Subtract 2 to account for the dropped PHY CRC
            L_new_hex=$(printf "%02X" $((L_dec - 2)))

            # 3. Extract the payload (skip first 2 chars, drop last 4 chars)
            payload_len=$((${#rawhex} - 6))
            payload="${rawhex:2:$payload_len}"

            # 4. Output the corrected telegram
            echo "${L_new_hex}${payload}"
        fi
        
    done
    
    # If cat crashes or hits EOF, wait 1 second and reopen the port
    sleep 1
    
done
```

weetmuts:
Many thanks for the example script! 

Ah, those pesky length bytes. The professional dongles adjust the length byte after the crc:s have been removed.
I manage to change rtl_wmbus to do the same. 

What is the name of your commercial device that produces slightly wonky telegrams with the wrong length?
wmbusmeters usually accepts the wrong length byte when whole telegrams are parsed, ie the real length is know.
But it prints a warning, I suppose that is what you avoid here?

However the length byte must be right when reading from raw byte streams since there is no way to know the length.

weetmuts:
Ah, yes, the hex input is in fact a raw byte stream, so you must adjust the length byte.....

jktjkt:
In the ESPHome implementation (which bundles a fork of wmbusmeters), the code deals with the raw output of the FSK decoder, so it [calls wmbusmeters' functions](https://github.com/SzczepanLeon/esphome-components/commit/407e2aadef074730165c8f0801ddedfe03e47608) to strip the CRC bytes and to adjust the length accordingly. The way how I understand @hollie's comment, he has to do the exact same thing, so maybe wmbusmeters could accept the "truly raw packet", including the CRCs and the original length bytes (as a different input format, or with a new flag in the input JSON saying "hey, this is totally unprocessed").

weetmuts:
Good idea, we could add stdin:hexcrc and hexcrc:CMD(....) and rawttycrc.

We can reuse the existing wmbus_rawtty.c code and just pass bool crc = true when
we expect the input to have DLL crc bytes.
  
Do you you want to have a go at this fix? :-) As you said all the crc testing and removal code already exists
and is used in the wmbus_cul.cc driver already.

weetmuts:
But it wont help @hollie current json content, which has the crc removed but the length not adjusted.

jktjkt:
> Do you you want to have a go at this fix? :-) As you said all the crc testing and removal code already exists and is used in the wmbus_cul.cc driver already.

I'm afraid I won't have much time for this for a few weeks, I have to finish the readout and visualization first, and only then I'll be able to refactor stuff. Sorry for this armchair suggestion :).

hollie:
> But it wont help @hollie current json content, which has the crc removed but the length not adjusted.

Hey @weetmuts to be clear: my device does send the CRC and the length field is matching the full packet including the CRC. I strip the CRC and adapt the length in the script.

And having the script as intermediate is not a problem for me. It just took some time to figure out why the packet decoded correctly on wmbusmeters.org and not in the application :smile:.

The device is a custom-built wmbus receiver based on a Silabs controller with built-in sub 1GHz radio.

weetmuts:
@hollie Oh, are you sure you handle both A and B frames and all the crcs (there can be more than one in a single telegram)?

hollie:
@weetmuts I can set it to whatever I want, the wMBus lib of Silicon Labs supports multiple modes (S, T, C, F, N) and the various frame types (A/B). It is just a compile option.

PR #1367, opened 2024-09-18
Driver detection for (DEV) Develco Products Electricity Meter Version 0x01

zapster:
The driver for the Develco Products Electricity Meter Version 0x01, as use by my local grid operator linznetz.at, is not detected automatically:

```
./build/wmbusmeters auto:t1
Started auto cul on /dev/ttyUSB0 listening on t1
No meters configured. Printing id:s of all telegrams heard!
Received telegram from: 00320787
          manufacturer: (DEV) Develco Products, Denmark (0x10b6)
                  type: Electricity meter (0x02) encrypted
                   ver: 0x01
                device: cul
                  rssi: -52 dBm
                driver: unknown!
```

It seems, however, that the amiplus driver works, as identified by `--analyze`:

```
./build/wmbusmeters --analyze=<key> auto:t1
Started auto cul on /dev/ttyUSB0 listening on t1
Auto driver    : not found!
Similar driver : amiplus 24/28
Using driver   : amiplus 24/28
000   : 3e length (62 bytes)
001   : 44 dll-c (from meter SND_NR)
002   : b610 dll-mfct (DEV)
004   : 87073200 dll-id (00320787)
008   : 01 dll-version
009   : 02 dll-type (Electricity meter)
010   : 7a tpl-ci-field (EN 13757-3 Application Layer (short tplh))
011   : 38 tpl-acc-field
012   : 00 tpl-sts-field (OK)
013   : 3005 tpl-cfg 0530 (AES_CBC_IV nb=3 cntn=0 ra=0 hc=0 )
015   : 2f2f decrypt check bytes (OK)
017   : 0C dif (8 digit BCD Instantaneous value)
018   : 78 vif (Fabrication no)
019 C?: 30253390
023   : 06 dif (48 Bit Integer/Binary Instantaneous value)
024   : 6D vif (Date and time type)
025 C!: 687214123940 ("device_date_time":"2024-09-18 20:50:40")
031   : 0E dif (12 digit BCD Instantaneous value)
032   : 03 vif (Energy Wh)
033 C!: 189169000000 ("total_energy_consumption_kwh":699.118)
039   : 0E dif (12 digit BCD Instantaneous value)
040   : 83 vif (Energy Wh)
041   : 3C combinable vif (BackwardFlow)
042 C!: 926501000000 ("total_energy_production_kwh":16.592)
048   : 0B dif (6 digit BCD Instantaneous value)
049   : 2B vif (Power W)
050 C!: 260200 ("current_power_consumption_kw":0.226)
[...]

PR #1752, opened 2026-03-07
amiplus: decode total energy consumption from DIF/VIF 0E03 to avoid bogus totals

zapster:
## Summary

This PR makes `amiplus` decode `total_energy_consumption_kwh` from a strict key match (`DifVifKey("0E03")`) instead of a broad `AnyEnergyVIF` matcher.

The change fixes intermittent mis-decoding of total energy consumption (e.g. millions/billions kWh) seen on real Apator/AMI+ telegrams.

## Problem

For my meter `id=00320787` (`amiplus` driver, support added https://github.com/wmbusmeters/wmbusmeters/pull/1367), decoded totals alternated between correct values (~`7436.xxx` kWh) and clearly wrong values such as:

- `393153731.563008`
- `12138361.944444`
- `8665487929.722221`

Example from captured log:

Correct:

```text
telegram=|3E44B6108707320001027A020008052F2F_0C7830253390204B2968336015660E036669430700000E833C9265010000000B2B5702000BAB3C0000002F2F2F2F|+132
{"_":"telegram",...,"total_energy_consumption_kwh":7436.965,...}
```

Wrong:

```text
telegram=|3E44B6108707320001027A020008052F2F_0C783025339022493D6A316217640E036769430700000E833C9265010000000B2B5702000BAB3C0000002F2F2F2F|+141
{"_":"telegram",...,"total_energy_consumption_kwh":393153731.563008,...}
```

## Root cause

`src/driver_amiplus.cc` previously used:

```cpp
.set(MeasurementType::Instantaneous)
.set(VIFRange::AnyEnergyVIF)
```

`AnyEnergyVIF` can match multiple energy-coded fields (`EnergyWh`, `EnergyMJ`, `EnergyMWh`, `EnergyGJ`) in mixed frame layouts. In bad frames this bound `total_energy_consumption` to the wrong field.

## Why `0E03` is correct

The protocol description from my grid operator Linz Netz states the following regarding wireless M-Bus spec (unfortunately only available in German):

https://www.linznetz.at/media/linz_netz_website/netz_dokumente/Beschreibung-Wireless_M-Bus-Schnittstelle.pdf

- Page 6 (`MBUS-Daten`):
  - `1.8.0 0x0E 03 ... Zählerstand Bezug Wirkenergie`
  - `2.8.0 0x0E 0x83 0x3C ... Zählerstand Lieferung Wirkenergie`
- Page 9 (`Beispiel (6/6)`):
  - `0x0E 0x03 ... 18561 Wh 1.8.0`
  - `0x0E 0x83 0x3C ... 16604 Wh 2.8.0`

Parser-level corroboration in wmbusmeters core:

- `src/dvparser.cc`: `0xE` data field => `12 digit BCD`
- `src/dvparser.h`: `VIFRange::EnergyWh` covers `0x00..0x07` (incl. `0x03`)
- `src/wmbus.cc`: DIF function bits `0x00` => `Instantaneous`

## Change

`src/driver_amiplus.cc`

For `total_energy_consumption`:

```cpp
FieldMatcher::build()
  .set(MeasurementType::Instantaneous)
  .set(DifVifKey("0E03"))
```

## Validation

I captures a few telegrams and decoded them.

### Replay before fix

- `51` telegram/json pairs with total
- `33` in correct range (`7000..8000`)
- `18` wrong (`>1e6`)

### Replay after fix

- no wrong decodings
- analyzed totals are only sane `7436.xxx`
- formerly wrong frames no longer emit bogus `total_energy_consumption_kwh`

### Build/tests

- local rebuild succeeded (`make`)
- `make test` passed

## Behavior impact

- prevents absurd total-energy-consumption spikes from wrong energy-field binding
- keeps correct `0E03` frames decoding as expected
- for variants without `0E03`, field may be absent instead of wrong (preferred)

That said, I've only validated with my meter and the tests. Not sure if there could be a negative impact on other meters using the `amiplus` driver. If that is a concern, should we make a (sub-)config for this particular meter?

### DIFF
--- a/src/driver_amiplus.cc
+++ b/src/driver_amiplus.cc
@@ -53,7 +53,7 @@ namespace
             VifScaling::Auto, DifSignedness::Signed,
             FieldMatcher::build()
             .set(MeasurementType::Instantaneous)
-            .set(VIFRange::AnyEnergyVIF)
+            .set(DifVifKey("0E03"))
             );
 
         addNumericFieldWithExtractor(


weetmuts:
Oupsie, I need an example telegram that demonstrates the wrong decoding from you.
AnyEnergyVIF should always be scale safe.

weetmuts:
Sorry, your claude assisted fix and analys seems to be completely wrong.

Your "broken" telegrams sends exactly that value:
https://wmbusmeters.org/analyze/3E44B6108707320001027A020008052F2F0C783025339022493D6A316217640E036769430700000E833C9265010000000B2B5702000BAB3C0000002F2F2F2F

What could be wrong is the scale factor, but according to vifScale in wmbus.cc the value should be divided
with 1,000,000 ie one million. which Makes sence 1000 to get from mWh to Wh and 1000 again to get to kWh.

Can you check with the meter manufacturer what the expect this value to be really....

weetmuts:
```
(gdb) p 0x000165923C830E00
$1 = 393153731563008
(gdb) p 393153731563008.0/1000000.0
$2 = 393153731.56300801
(gdb) p 393153731563008.0/1000.0
$3 = 393153731563.008
(gdb) p 393153731563008.0/1000000.0
$4 = 393153731.56300801
```

zapster:
@weetmuts thanks for double checking. good to know that humans still do proper reviews. :) I tried to find enough evidence that the solution that opencode found for my real-world problem is correct, but the only information I found was the (german) doc from my grid operator which only mentions the 0E03 key. That was enough the convince me.

But of course what you say makes very much sense.

> Can you check with the meter manufacturer what the expect this value to be really....

I'll try to get more information about meter.

> Oupsie, I need an example telegram that demonstrates the wrong decoding from you.
AnyEnergyVIF should always be scale safe.

Would that still be helpful to you?

I see you reverted this PR in https://github.com/wmbusmeters/wmbusmeters/commit/f6068e681dba9fa7aa273f4ff0a92feaca0ac162. Thanks! I assume there where no other changes in this area, e.g., attempts to fix the scaling, correct?

zapster:
[wmmeter-defunc01.log](https://github.com/user-attachments/files/26372193/wmmeter-defunc01.log)
fwiw, here is the log that I captured to analyze the issue.

PR #1721, opened 2026-02-12
Fix hydroclima driver for firmware version 0x85

Hymbajs:
## Summary

- Add auto-detection for BMeters HydroClima firmware version **0x85** (`addMVT(MANUFACTURER_BMP, 0x08, 0x85)`)
- Complete decoding of frame 0x11 (v0x85) manufacturer-specific data per **PAPP-HARF2** specification (KA1 structure, 47 bytes)
- Add DIF/VIF extractors for `set_date` and `consumption_at_set_date` (storage nr 1)
- Add two encrypted test cases for v0x85

## Problem

Version 0x85 HydroClima 2 ITN devices (manufacturer BMP, type 0x08) were not auto-detected because only version 0x53 was registered. When forced via `driver=hydroclima`, the driver misinterpreted the frame 0x11 data layout — frames 0x10 and 0x11 diverge at **byte 7** (after DAT), not byte 15 as originally coded.

### `src/driver_hydroclima.cc`

Frame 0x10 (v0x53, 24B mfct) and frame 0x11 (v0x85, 47B mfct) now have separate parsing paths diverging at byte 7.

**Frame 0x11 decodes the full KA1 structure:**

| Offset | Field | Description |
|--------|-------|-------------|
| 0 | Frame ID | 0x11 |
| 1–2 | STS | Status word |
| 3–4 | TIM | Current time |
| 5–6 | DAT | Current date |
| 7–8 | DOP | Housing open event date |
| 9–10 | TKA | Average heater temperature (current period) |
| 11–12 | TOA | Average ambient temperature (current period) |
| 13–14 | TMH1 | Max temperature (previous period) |
| 15–16 | TMHD1 | Date of max temperature |
| 17–18 | TKA1 | Average heater temperature (previous period) |
| 19–20 | TOA1 | Average ambient temperature (previous period) |
| 21–28 | CNI1–4 | Consumption history (4 × 2B, HCA) |
| 29–34 | TONI1–3 | Ambient temperature history (3 × 2B, 0x8000 = no data) |
| 35–43 | TK counts | Temperature bin counters (3 × 3B) |
| 44–46 | U | Total consumption all periods (3B, /10 = HCA) |

## Verification

Tested on **52 real-world v0x85 meters** — all decode correctly with plausible values. Backward compatibility with v0x53 confirmed.

## Test plan

- [x] `make` compiles without warnings
- [x] `testinternals` passes (exit 0)
- [x] `test_json_stdin.sh` passes
[...]

PR #1756, opened 2026-03-12
Complete hydroclimav2 IXML grammar with null_value support

Hymbajs:
## Summary

- Complete IXML annotation for `hydroclimav2.xmq` driver — all 14 DV keys covering frame_v10 and frame_v11
- Add 10 new field definitions: previous period temperatures, CNI1-4 consumption history, TONI1-3 ambient temperature history, total_consumption
- Add `null_value` attribute support in XMQ field definitions — when extracted value matches `null_value`, it renders as JSON `null` (implemented in `driver_dynamic.cc`, `meters.cc`, `meters.h`)
- TONI sentinel `0x8000` (-327.68°C) correctly renders as `null` instead of a misleading temperature value

### Fields added
| Field | DV key | Description |
|-------|--------|-------------|
| average_ambient_temperature_last_month | 4265 | Previous period avg ambient temp |
| average_heater_temperature_last_month | 4259 | Previous period avg heater temp |
| consumption_at_set_date_1..4 | 82016E..C2026E | LIFO consumption history (HCA) |
| ambient_temperature_at_set_date_1..3 | 820165..820265 | LIFO ambient temp history |
| total_consumption | 83056E | Total HCA across all periods (24-bit integer) |

### Changes
- `drivers/src/hydroclimav2.xmq` — IXML grammar, fields, tests
- `src/driver_dynamic.cc` — `null_value` extraction and NaN replacement
- `src/meters.h` / `src/meters.cc` — `null_value` member in FieldInfo
- `src/generated_database.cc` — regenerated
- `tests/generated_tests.xmq` — 3 test cases (v0x53 NOKEY, v0x85 encrypted, v0x85 TONI null)

## Test plan

- [x] testit.sh: v0x53 (68036198, NOKEY) — frame_v10 OK
- [x] testit.sh: v0x85 (93000952, encrypted) — frame_v11 OK
- [x] testit.sh: v0x85 (93001021, TONI null) — `ambient_temperature_at_set_date_3_c: null` OK
- [x] Validated against 52 real HydroClima devices (Praha Na Míčánkách + Liberec)
- [ ] `build/xmq` required for `test.sh` generated tests (needs `libxslt1-dev`)

Supersedes the C++ fix in #1721.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/drivers/src/hydroclimav2.xmq
+++ b/drivers/src/hydroclimav2.xmq
@@ -239,5 +239,12 @@ driver {
             json = '{"_":"telegram","media":"heat cost allocation","meter":"hydroclimav2","name":"HC1","id":"93000952","ambient_temperature_at_set_date_1_c":22.66,"ambient_temperature_at_set_date_2_c":23.12,"ambient_temperature_at_set_date_3_c":22.66,"average_ambient_temperature_c":22.71,"average_ambient_temperature_last_month_c":21.78,"average_heater_temperature_c":49.19,"average_heater_temperature_last_month_c":37.34,"consumption_at_set_date_1_hca":468,"consumption_at_set_date_2_hca":2265,"consumption_at_set_date_3_hca":1913,"consumption_at_set_date_4_hca":1632,"current_hca":596,"max_ambient_temperature_c":58.2,"target_date":"2025-12-31","target_hca":2265,"total_consumption_hca":2435,"status":"OK","timestamp":"1111-11-11T11:11:11Z"}'
             fields = 'HC1;93000952;OK;596;22.71;1111-11-11 11:11.11'
         }
+        test {
+            comment = 'Version 0x85 frame_v11 with TONI null sentinel 0x8000'
+            args = 'HC3 hydroclimav2 93001021 NOKEY'
+            telegram = 5144b0092110009385088c20fb7afb4040252F2F_036E000000436E000000426C3F350F110001B9604734CE2E2708FD071B088832940766070000000000000000C407260800802DD400000000000000220000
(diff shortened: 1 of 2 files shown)


weetmuts:
Now we are talking! This looks excellent! In particular I like that you added the null_value setting to the dynamic driver.

How did you create the DV keys to generate from the ixml? Like 83056E.
Even though I have now read wmbus telegrams since 2017, I find it somewhat tedious to reverse engineer the DV key.

I was thinking that we could also support a more explicit text that is then recreated into the DV key,
that is then decoded again when the field matcher runs.

So something like `+'int32 Instantaneous Volume_l Storage_1'` would generate 4413

No need to update this pull request with that work, it could be separate.

Hymbajs:
Thanks! The DV keys were created by combining two approaches:

1. **`--analyze` mode** — running the telegram through `wmbusmeters --analyze=:hydroclimav2:KEY` gives the byte-by-byte breakdown with DIF/DIFE/VIF/VIFE already parsed. So I could see e.g. `83 05 6E` in the output and just concatenate them into `83056E`.

2. **PDF cross-reference** — the BMeters PAPP-HARF2 spec defines the KA1 frame structure with exact byte positions and data types. For multi-storage fields (CNI1-4, TONI1-3) the DIFE storage numbers follow a predictable pattern, so once you decode the first one, the rest are just incrementing the storage bits.

The readable text idea sounds great — constructing DV keys manually from the spec (especially with DIFE extension bits for higher storage numbers) is error-prone. Something like `'int24 Instantaneous HeatCostAllocation Storage_3'` → `82026E` would be much nicer to write and review.

weetmuts:
I made a small follow up change, renamed current_hca to consumption_hca, both in the driver and in the library.

It is part of my work to try to standardize the field names, and since we are creating a new driver, we are allowed
to change the names.

I like total_m3 for the latest reading of water volume, and target_m3 for the most reasonable billing value (typically end of last month).
Likewise for energy total_kwh and target_kwh.

The hca counter is for some meters reset once per year, so it is not strictly increasing, therefore total_hca
might not be perfect name, so I used current_hca. But since current also means amperage (electricity)
it feels off in my mind. So I renamed the library value to consumption_hca.

weetmuts:
Have we ported all test cases from the old driver to the new, so that we can switch over the MVT:s? Ie auto lookup of which driver to use.

weetmuts:
You can now see the IXML decode as part of the analyze output:
https://wmbusmeters.org/analyze/5144b0092110009385088c20fb7afb4040252F2F036E000000436E000000426C3F350F110001B9604734CE2E2708FD071B088832940766070000000000000000C407260800802DD400000000000000220000:hydroclimav2

weetmuts:
Oh, the tpl status field has bit 0x40 set. Do you know what this means?

PR #1762, opened 2026-03-15
Cleanup config location and include path handling

pepe82sh:
1. Removed legacy behaviour from config path
2. Made configuration directories configurable at build time (default behaviour matches current behaviour)
3. Utilise existing autotools to get correct include paths and libraries in a platform independent fashion

This is tested on FreeBSD and MacOS, but not on a Linux distribution. It might be worth it to build on a linux distribution before merging (I didn't have one readily available). From a users point of view, nothing should have changed. The installation steps match the previous ones.


pepe82sh:
It might make sense to check changes in Makefile in the individual commits. dfd10fc9d789adedc73cfd2d8bffb4a197360c9d only moves it to autoconf/Makefile.in, while 7a1c9771898c0267d7752ef5a6d5e78f7f28252e introduces the changes. The combined diff gives no indication of what's changed and what hasn't, since it doesn't seem to pick up on the move itself.

weetmuts:
Thanks! But why did you create a Makefile.in ? Is it really necessary?

pepe82sh:
> Thanks! But why did you create a Makefile.in ? Is it really necessary?

One of the main advantages of using autoconf is that is that it helps you getting the correct CFLAGS and LDFLAGS. The usual way would be to inject them via the Makefile.in. I tried to follow best practices as much as possible, taking advice from [here](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Makefile-Substitutions.html). Another approach could be to create another makefile to inject it into, and include that in the "normal" Makefile. Wouldn't have the benefit of enforcing .configure to be executed before make, but should be possible I think.

Edit: Maybe it would "feel" better to have the Makefile.in in the root directory, so it's obvious what strategy is expected with this codebase?

weetmuts:
Yes, I see. I think however that I will eventually move the build to the process that I use here: https://github.com/libxmq/xmq 

Try building libxmq and check the contents of the build output. Not sure if it builds on FreeBSD though. :-)

I.e. the configure script creates a spec.mk file that is included by the Makefile.
The spec.mk can be read and understood, ie all the variables are easily checked because they are written in a file.

So, same problem different solution. 

I used the trick of generating a spec.gmk file when I designed the improved build system for OpenJDK many years ago.

pepe82sh:
@weetmuts I can come up with a solution that utilises a spec.mk, all I care about is getting rid of hard coded paths. I'll not attempt to copy what's in the repo you pointed at though, but I can use this PR to get one step closer to your preferred solution.

pepe82sh:
@weetmuts I've updated the PR in accordance with  your feedback

pepe82sh:
Old tests config locations assumed "old" behaviour with 'etc' injection. Moved those out of etc for tests to not fail anymore.

weetmuts:
When I try it, my Ubuntu las version 0.6.0 of rtlsdr but now we require 2.0.2 ?
Which is nice, but this might break a lot of builds out there.

Have I accidentally forced an older version of rtlsdr or is the debian deb
so very much out of date?

pepe82sh:
The min versions were pretty much a best guess, sticking with the major versions I saw on my system. If it worked for you before, setting the min to 0.6.0 should be fine and will not break it for anyone I suppose. Would that be the suggestion to set min to 0.6? That would be the one that shipped with Debian bullseye in 2024.

weetmuts:
Thanks! I merged your changes separatey, github created a lot of nonsense in the diff.

weetmuts:
Check if it works for you. You had removed some brew specific config, you sure it will work with mac/brew still? :-)

weetmuts:
Ok, it does not build the docker stuff.... do you know why?

[4](https://github.com/wmbusmeters/wmbusmeters/actions/runs/23246410917/job/67575918297#step:5:1015)
:: checking for pkg-config... no
:: checking for LIBXML... no
:: configure: error: in `/data/parts/wmbusmeters/build':
:: configure: error: The pkg-config script could not be found or is too old.  Make sure it
:: is in your PATH or set the PKG_CONFIG environment variable to the full
:: path to pkg-config.

pepe82sh:
From looking at it the Dockerfile needs pkgconfig in the app add command. I can try and fix it tonight.

I tested the build with brew and it works fine. Not having to worry about the specific host system is why I suggested this strategy.

Looking at the Dockerfile: is there a reason you get all dependencies from the head of their git repo and build them? I would have expected it to be pulled as packages as a more stable approach.

pepe82sh:
@weetmuts I just saw the actions and the commit. Is it safe to assume that this has been fixed?

weetmuts:
Yes! It seems to be working now.

weetmuts:
Thanks for the improvement! I had postponed this work for way too long.

pepe82sh:
@weetmuts  I also have some issues with the call to external programs via shell. The call to pidof I can get rid of just patching it out for the freebsd package (I think that only exists to potentially emit a warning, so not having it should work), the checks for rtlsdr and rtlwmbus I don't really understand, since they don't seem to be used from the shell after that, and the call to curl seems easy to fix by using libcurl instead of a call to the actual executable. Would you be open for me providing improvements in that direction?

weetmuts:
wmbusmeters starts the pipeline "rtl_sdr | rtl_wmbus" in a shell script when it detects an rtl_sdr dongle.
But I had tons of error reports where people had not installed these programs corrrectly. So therefore
it checks before. You cant really get rid of those external calls with a major reworking....

pidof is just for convenience. to warning if the daemon is running. Also a lot of people had trouble because
of this, so the warning is important.

I do not quite see the problem with curl being an external program.... surely curl is in every freebsd?

weetmuts:
In short, I do not think it is worthwhile to try to avoid the external calls. But we can avoid pidof in freebsd.

weetmuts:
You can of course ignore the checks for rtl_sdr and rtl_wmbus in freebsd, but that would be a worse user experience since wmbusmeters will start the pipeline anyway if you use an rtlsdr, and you will get much worse information if something is wrong.

pepe82sh:
pidof is there, but not necessarily in /bin/pidof. In FreeBSD it would be /usr/local/bin/pidof (if the package for it is installed). I can patch that just for freebsd, introduce $PREFIX for the build process, or not call external programs unless necessary. Haven't decided, but just removing it is the simplest without patching too much stuff after fetching this repo. Second easiest would be to just patch the path to be '/usr/local/bin/pidof', but I don't think that covers all corner cases. The most correct thing given the concept might be to support $PREFIX at buildtime, which isn't too big of a change either. I'm just not sure if that is supposed to have an impact on runtime, or if it's meant to only to be considered at build time.

As for curl, I pretty much think it's not only very uncommon to call external programs when they offer libraries, it also introduces additional complexity. So in the current implementation, you put together some strings, create a call to curl with those strings as arguments, curl then parses those arguments, uses libcurl to execute accordingly and write the received data into a file, takes the return code and prints it to stdout, which is then parsed by you and interpreted.
If you just use libcurl directly, you safe a ton of indirection, string writing and parsing, and computational resources.

weetmuts:
Yes, I understand, but I also balance the fact that I do not want wmbusmeters to contain too much powerful external software. Another person wanted to open a tcp socket, but I made that into a unix socket. They can use socat or some other tool to map a public tcp to the unix socket. Otherwise I will eventually get requests for encryption of the tcp socket and then... and then... and then..

wmbusmeters seems to work fine with a lot of external tools, mosquitto to push mqtt telegrams.
Curl to fetch remote REST api calls, not just fetching drivers. 
invoke sql to store into database. etc etc.
You can start a server with netcat check the README.

 I would like to keep it self-contained as much as possible and have wmbusmeters work in pipelines.
The unix philosophy. :-)

pepe82sh:
It's your project after all, that's why I'm asking before making an effort to write a PR. Just let me share my thoughts on your reply:

Using a shared library isn't introducing more or less complexity to the program itself. It's still calling external software, just using an API that is built for machines instead of a command line call that is designed for humans. It also doesn't match the unix philosophy, since no matter how you turn it, your software will call curl to do it's job, one way or the other.
What DOES match the unix philosophy is that you just pipe out the values in some text format, which is a beautiful thing, because the users can do with it what they want, how they want. Also the Alarms I get, since it's allowing users to pipe to whatever they need to be done.

weetmuts:
@pepe82sh I changed from pidof to pgrep which should be a core binary under FreeBSD.

The pgrep test is only invoked when wmbusmeters has failed to find any dongle.
Which is where the warning makes sense, ie you could not find any dongle perhaps
because another wmbusmeters has taken the serial device. :-)

PR #1763, opened 2026-03-16
rfmamb: convert to XMQ with mfct_tpl_status_bits support

Hymbajs:
## Summary
- Convert BMeters RFM-AMB driver from native C++ to XMQ format
- Add XMQ engine support for `mfct_tpl_status_bits` block — allows XMQ drivers to override standard TPL status byte interpretation
- BMeters maps TPL bit 3 as `MODULE_REMOVED` (tamper) instead of standard `PERMANENT_ERROR`; bits 5-6 as `TEMP_OUT_OF_RANGE` / `RH_OUT_OF_RANGE`

## How it works
New `mfct_tpl_status_bits` XMQ block parsed in `driver_dynamic.cc`, stored in `DriverInfo::mfct_tpl_status_bits_`. When the lookup mask covers standard bits (0-4), `decodeTPLStatusByteWithMfct()` skips standard decoding and uses the manufacturer mapping for the entire byte.

Existing C++ drivers using `setMfctTPLStatusBits()` (lansenth, ei6500, etc.) are unaffected — the new code path only activates when `coversFullByte()` returns true.

## Files changed
- **New**: `drivers/src/rfmamb.xmq` — XMQ driver (16 fields + 7 status bit mappings)
- **Removed**: `src/driver_rfmamb.cc` — replaced by XMQ
- **Modified**: `src/driver_dynamic.cc/.h` — parse `mfct_tpl_status_bits` block
- **Modified**: `src/translatebits.h` — `Lookup::coversFullByte()` method
- **Modified**: `src/wmbus.cc` — extended `decodeTPLStatusByteWithMfct()`
- **Modified**: tests/simulations — `PERMANENT_ERROR` → `MODULE_REMOVED`

## Test plan
- [x] `test_drivers.sh rfmamb` — OK json + OK fields (2 tests: NOKEY + encrypted)
- [x] `test_stdin_and_file.sh` — OK (6 subtests)
- [x] `test_serial_bads.sh` — OK (2 subtests)
- [x] `test_drivers.sh lansenth` — OK (existing C++ mfct_tpl_status_bits driver still works)
- [x] Bulk tested 155 real RFM-AMB devices from 7 locations — 155/155 OK, all `MODULE_REMOVED`


weetmuts:
>This PR depends on https://github.com/wmbusmeters/wmbusmeters/pull/1763 (rfmamb XMQ conversion) for the mfct_tpl_status_bits XMQ feature which enables proper TPL status decoding

Well, I understand the need, but it is not PROPER decoding. It is YET ANOTHER WORKAROUND for stonking meters that violate the standard.... :-/

weetmuts:
But this is a good implementation, and I had yeat to add support for the mfct specific bits high in the tpl status byte anyway. So this request solves both problems. Thank you very much!

weetmuts:
But on the other hand, if the module removed perhaps it is a PERMANENT_ERROR..... 
again, did we violate the consistency of the json output from drivers?
Perhaps this is acceptable.

PR #1766, opened 2026-03-16
c5isf: convert Zenner Zelsius C5 ISF to XMQ driver

Hymbajs:
## Summary
- Convert Zenner Zelsius C5 ISF heat/cooling meter driver from native C++ to XMQ format
- 55 fields total across 3 telegram types (T1A1, T1A2, T1B)
- First XMQ driver to use `DecimalsToString` map_type for status decoding (18 error conditions)
- 14-month energy/volume history (storage_nr 32-45)

## Files changed
- **New**: `drivers/src/c5isf.xmq` — XMQ driver (55 fields, 3 MVTs, 3 tests)
- **Removed**: `src/driver_c5isf.cc` — 294 lines native C++ replaced
- **Modified**: `src/generated_database.cc` — compact builtin entry + 3 MVT mappings
- **Modified**: `tests/generated_tests.xmq` — 3 test blocks (T1A1 heat, T1B billing, T1B heat+cool)

## Test plan
- [x] `test_drivers.sh c5isf` — OK json + OK fields (all 3 tests pass)
- [x] Bulk test on 276 real Zenner devices from device_list (pending)


Hymbajs:
## Bulk test results

Tested on 38 real Zenner Zelsius C5 ISF devices (all that have captured telegrams in our device lists):

- **38/38 OK**,
- **0 partial decodes** — all fields match correctly
- Status: 38× OK
- Energy range: 0–12,870 kWh (avg 2,622)
- Volume range: 0.007–2,035 m³ (avg 431)
- Flow temperature: 20.6–53.9 °C
- Return temperature: 16.7–51.1 °C

DecimalsToString status lookup works correctly — first XMQ driver to use this map_type.

weetmuts:
Great work, there is an improvement you can try, I implemented looping for repeated fields in incremented storage positions. Try this:

```
field { 
    name = 'prev_{storage_nr-31]_month'  
    quantity = PointInTime  
    info = 'Historical month date.'   
    match { 
        measurement_type = Instantaneous  
        vif_range = Date  
        storage_nr = 32,45 
    }
}
```

and likewise for the other repeating fields.

Hymbajs:
Good catch! Added the missing T1A2 (volume history, media=water) test case in fe71fae4.

weetmuts:
Thanks!

weetmuts:
Did some cleanups for example due_date_date should be due_date. The field name should always lack the unit.
If you trie to add `name = something_m3` wmbusmeters will complain because it wants 100% control of the units.
So you add `name = something` and the `quantity = Volume` will generate the default unit.

However, there is an exception for this check for: _date _month _year 
This is for historical reasons there are some old meter fields that use these endings.

So, the prev_1_month was changed to prev_1_month_datetime because the default unit for PointInTime is datetime.
Alas, pretty often the meter only sends the date part. So we override the default unit with just date
`display_unit = Date` so now it is prev_1_month_date

I just pushed this change. 

Alas the original field name was prev_1_month so we violated the rules of wmbusmeters as specified in the README: `If the driver name is the same we are only allowed to add new fields!`

 :-/ Oh, well, my bad.

Fortunately the current reading has the same name, it is "only" the historical data.

So, next time, if we cannot replicate the exact json from the old c++ driver because it uses 
old style formatting of the fields that are no longer permitted, then we create  new driver,
in this case it should have been c5isfv2.

This has happened a few times already, hydroclimav2 qwaterv2 qheatv2
The auto lookup can change to the new driver.
The idea is that users of wmbusmeters can upgrade at any time, as long as they are not using auto,
it will just work, always backwards compatible. Somehow we should publish a warning inside the
generated json that the old driver is going to be removed at a certain date, usually 6 months after 
the warning starts being printed.

@Hymbajs makes sense?

PR #1768, opened 2026-03-16
chore: migrate elf driver to xmq

Uzlopak:
I will be honest, I used claude.ai to migrate.


weetmuts:
Thanks for the disclosure of claude! It is indeed a chore to migrate, so the work in doing it, with or without claude is appreciated. 

But I think this:
  name     = meter_date
  quantity = Text

should actually be:
  name     = meter
  quantity = PointInTime

or even simpler just add
library {
    use = meter_date
}

before the fields { }

weetmuts:
name = meter
quantity = PointInTime
display_unit = date

The PointInTime deaults to full date+hour, but here we only want the date part.

weetmuts:
I merged your elf.xmq separately and did a few simplifications where we can use the library. Thanks!

Uzlopak:
I will take a look, maybe i can do some other migrations.

But i am currently working on a cool PR, which probably will blow you away...

PR #1769, opened 2026-03-17
Add short packet signatures for some Kamstrup meters

jktjkt:
The code change feels like a lot of repetition, but it's the way how it's currently done, and I have these meters installed, so here we go. Would it make sense to move this signature -> format mapping to the XMQ definition of meter types?


weetmuts:
Yes, it would be much better to add to the driver.xmq something like:
```
difvif_signatures {
   // We use these telegrams to extract the signature of the telegram already when 
   // the driver is loaded, so we can decode the first compressed telegram. No need
   // to wait 8 telegrams for the first long telegram. The telegram has to be decrypted already.
    extract =  <hex>
   }
}
```

PR #1771, opened 2026-03-17
ci: add linting for xmq, initial format of xmq files

Uzlopak:


weetmuts:
Nice! But I will definitely forget this when I commit/push, and I will get a lot of annoying lint errors. Perhaps we can add a hook in the git somewhere to check this earlier?

weetmuts:
In addition to this check. :-)

weetmuts:
Just for fun, have you tested this?
`xmq drivers/src/iperl.xmq pa`
and
`xmq drivers/src/iperl.xmq br`

You can also do:
`xmq something.json br`
`xmq something.json pa`

and
`xmq something.xml br`

and 
`xmq index.html delete //script delete //style br`

weetmuts:
Or this to extract all img from an html page and then render a new page with only the images:
`xmq dagens.html select //img add-root body add-root html to-html br`

Uzlopak:
No, but this looks cool. I was just trying all the options in the drivers make file and it was strange that the xmq files are not prettiffied. I think it should be fixed once and then be taken care of.

Do you squash merge or do you merge normally? 
If you dont squash i would recreate the commit history.

What you can do is also something like this:
https://github.com/metro-digital/sops-decrypt/blob/master/.github/workflows/build-action.yaml
I could rewrite it to create a commit prettified files, but then it would not retrigger the other workflows. I close and open the PR then to retrigger them.

Also client-side pre-commit hooks could be potentially annoying, because if you just want to change something in the readme.md suddenly it says that you should build and install xmq. 

Maybe it makes sense to put the pretty call into the driver compilation phase. This is very specialized project, so we can assume that only "specialists" will run some of the make commands of the driver folder. 

So I did this...

weetmuts:
Perhaps we can invoke pretty when we do "make install" from the drivers directory?

Uzlopak:
@weetmuts 
I added it to the database command because you should run the database command if you modify an xmq file. As database is triggered by install, i think it is "good". But If you like I can make it explicitly be called only in install.

You can see that it works as a workflow here:
https://github.com/Uzlopak/wmbusmeters/pull/1

weetmuts:
Nice! If you cleanup the commit history, its ready for merging. :-)

Uzlopak:
@weetmuts ready :)

weetmuts:
Thanks!

PR #1779, opened 2026-03-20
poc: wmbusmeters running in chromium-browser family, emscripten wasm build for serial devices

Uzlopak:
This is a proof of concept for creating a wmbusmeters wasm and use it in the browser. Currently it is very buggy.

And the obligatory disclaimer that I used claudeAI and chatGPT to write alot of the code. 

You can test it here: 
https://uzlopak.github.io/wmbus/wasm/wmbusmeters.html

I commited the wasm just so that you can play with it without being forced to compile it.

I used emscripten. Maybe clang is better. I have to  investigate this further. 

The generated wasm file gets very big, right now 9,4 MB. I used wasm_opt to reduce the size to about 5,4 MB. So dont wonder why it is committed as 5,4 MB but your generated wasm is much bigger. ;). wasm_opt with -Os flag took about 5 minutes to slim it down. 

The code itself has to be slimmed down A LOT. Probably some functions are missing and have to be exposed. 

The primary goal was to get a proof of concept to compile it to a wasm and the injection of the serialport of the browser via the web serial api, and get the device detection to work. 

I obviously dont expect this code to be merged at this stage.


weetmuts:
Amazing work! Lets work together on merging this!

weetmuts:
The original wmbusmeters binary was less than 64 KiB in size! But it could only talk to the im871a dongle and only listen to the multical21 meter. Now we have tons of debug strings in the binary, descriptions of all fields, the analyze functionality, all 118 drivers, etc etc. I am sure we can slim it down for the wasm binary, in particular if we go for downloadable drivers.

Uzlopak:
Still buggy but it feels awesome!

image

Uzlopak:
Tested in chrome under windows and works too. I made a screenshot.. but the internet was soooo slow because windows wanted to download updates... because i use it only every 6-12 months... but well, it works.

PR #1800, opened 2026-04-07
chore: convert fhkvdataiii to xmq

Uzlopak:
I dont understand this message in the terminal:

```
OK: db test Rooom fhkvdataiii
(wmbus) telegram length byte (the first) 0x31 (49) is probably wrong. Expected 0x34 (52) based on the length of the telegram.
(wmbus) telegram length byte (the first) 0x31 (49) is probably wrong. Expected 0x34 (52) based on the length of the telegram.
```


weetmuts:
The length warning is easily fixed by correcting the first byte to be the expected length. The length is obvious from the number of bytes so the length byte is redundant here, just fix it in the text cases. 

This PR contains a lot of interesting stuff. Can you explain what you added and why?

PR #1804, opened 2026-04-07
chore: header cleanup

Uzlopak:
- use uint8_t instead of unsigned char
- reduction of nested header dependencies.


Uzlopak:
Some header cleanup is good to reduce the potential need to compile again source files because some irrelevant file changed. Right now wmbus.h is heavily bound to everything. I think this PR is a little bit excessive too. 

I think it is better to split this PR in potential three parts:
- replace uchar with uint8_t
- header cleanup
- and if you like the forward declaration instructions

weetmuts:
Yes, please split into separate PRs.

weetmuts:
I prefer to keep uchar as it is.

Uzlopak:
If we remove uchar, then we remove the strong "binding" to util.h.

weetmuts:
Please make one PR with only cleanup of existing includes.

Then another PR where you split util.h into always.h (contains uchar typedef and the verbose/debug/trace macros and interna function.) and keep the remaining in util.h.

(If you change always.h, then yes everything will be recompiled.)

Then another PR where you split the remaining util.h into different parts that make sense.... :-)

weetmuts:
You can check: https://github.com/libxmq/xmq/blob/master/src/main/c/parts/always.h

util.h in wmbusmeters really needs to be cleaned up. I just stuffed things in there over the years. :-)

PR #1820, opened 2026-04-09
chore: simplify isCloseEnough, remove genericifyMedia

Uzlopak:
I think this is better.


weetmuts:
Hmm, no I think it is more confusing.

PR #1821, opened 2026-04-09
chore: optimize initializer

Uzlopak:
Instead of filling the vector with strings on initialization, we use the preprocessor to prefill the vector at compile time.

### DIFF
--- a/src/wmbus.cc
+++ b/src/wmbus.cc
@@ -217,19 +217,12 @@ struct Manufacturer {
     }
 };
 
-vector<Manufacturer> manufacturers_;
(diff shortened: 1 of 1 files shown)


weetmuts:
Nice, it will not matter much in the future when most of the drivers are xmq. But still.
Did claude find this for you?

Uzlopak:
@weetmuts 

No, Claude did not this. I was examining the generated code with bloaty and it showed up. I have various other optimizations git stashed, which make heavier use of the preprocessor so that it reduces unecessary calculations at runtime.

PR #1822, opened 2026-04-10
chore: remove redundant isDebugEnabled calls

Uzlopak:
debug() is a macro, which itself calls isDebugEnabled in an if condition. Calling explicitly isDebugEnabled() is only useful on a bundle of debug calls.


weetmuts:
Thanks!

PR #1826, opened 2026-04-10
perf: improve valueToString

Uzlopak:
This avoids the popping and pushing back of characters.


weetmuts:
Ah, no too scary to change this method and I doubt it runs that much faster.

PR #1829, opened 2026-04-10
Migrate engelmann-faw driver from C++ to XMQ format

Uzlopak:


weetmuts:
Thanks! Converted it to engelmann_faw and added a driver alias engelamann-faw.

PR #1832, opened 2026-04-10
Migrate qsmoke driver from C++ to XMQ format

Uzlopak:


weetmuts:
Fixes were needed in wmbus.cc so the migration was not that simple.

PR #1835, opened 2026-04-11
chore: remove unused frameTypeKamstrupC1 function

Uzlopak:


weetmuts:
Thanks! That was an old artifact from the old days when I did not know about the wmbus protocol and thought it was a Kamstrup invention. :-)

PR #1836, opened 2026-04-11
fix for #1796 apator162

Uzlopak:
Fixes #1796

With help of claude ai...

* Added woot_7C grammar rule — was the root cause of the real-world regression (all production Apator 162 meters use this tag for monthly history)
* Fixed checksum to word, word? — some telegrams (like has80) have a 4-byte trailing checksum instead of 2-byte
* Added 14 missing test cases from the original CC driver: Wasser5, MyTapWaterb/c/d/e/f/g, NewAndOld ×2, has80, has79, and 3× MyTapWatere (26202020)


weetmuts:
Super! Nice that claude understands invisible xml! :-) Good work anyway. This is the typical problem, the protocol keeps generating new tokens for which the length is unknown. 

This is why the mbus protocol is superior, you can safely skip parts that you do not understand. Here we need
to maintain a reverse engineered list of tokens and the following data length.

Unless of course there are some bits inside the token boken that tells us how long the value is, but I haven't seen any such pattern so far....

If someone knows a bit more about this protocol, it should be possible to extract more information from the telegrams.

PR #1837, opened 2026-04-11
fix: reduce segfaults

Uzlopak:


jktjkt:
First of all, please explain *why* you made these changes (and put that into the commit message).

Second, the ASAN's trace in #1788 looks like the XMQ library reusing something between different calls. It is not obvious to me how this is realated to wrapping a driver instance with an RAII pointer in a try/catch block.

PR #1840, opened 2026-04-12
wip

Uzlopak:
This is here just for demonstration purposes.


drivers/src/abbb23.xmq
```diff
@@ -4,7 +4,7 @@ driver {
     meter_type     = ElectricityMeter
     default_fields = name,id,total_energy_consumption_kwh,timestamp
     detect {
-        mvt = ABB,02,20
+        mvt = ABB,20,02
```
Uzlopak, line 7: !

drivers/src/itron.xmq
```diff
@@ -4,17 +4,14 @@ driver {
...
     default_fields = name,id,total_m3,target_m3,timestamp
     detect {
-        mvt = ITW,07,00
-        mvt = ITW,07,03
-        mvt = ITW,07,33
-        mvt = ITW,16,00
+        mvt = ITW,00,07
+        mvt = ITW,03,07
+        mvt = ITW,33,07
+        mvt = ITW,00,16
```
Uzlopak, line 10: !

weetmuts:
Ah! Of course! I make that mistake myself!

weetmuts:
But the new xmq drivers look correct. There should be a swap from the c++ driver to the xmq driver. The C++ driver function call has the wrong name it says MVT but it is actually MTV....

There is even a test in test_listen_for_all.sh that would fail if the MVT was messed up for some of the converted drivers, and the test passes.

PR #1841, opened 2026-04-12
fix: chore wrong detection

Uzlopak:


weetmuts:
Yes, these were indeed wrong. Thanks for the fix!

PR #1842, opened 2026-04-13
feat: wmbusmeters running in chromium-browser family, emscripten wasm build for serial devices

Uzlopak:
follow up from #1779


weetmuts:
Awesome work!

PR #1851, opened 2026-04-15
Migrate sontex868 driver from C++ to XMQ format

Uzlopak:


weetmuts:
Thanks! Actually this is the default: `dif_signedness = Signed`
so you do not need to specify it.

PR #1867, opened 2026-04-24
chore: void std::function

Uzlopak:


weetmuts:
Nope. I do not see the value of this PR.

PR #1879, opened 2026-04-25
Migrate aventieshca driver from C++ to XMQ format

Uzlopak:


drivers/src/aventieshca.xmq
```diff
@@ -0,0 +1,146 @@
...
+            quantity   = Text
+            info       = Deprecated.
+            attributes = DEPRECATED
+            match {
+                measurement_type = Instantaneous
+                vif_range        = ErrorFlags
+            }
+            lookup {
+                name            = ERROR_FLAGS
+                map_type        = BitToString
```
Uzlopak, line 67: Can this deprecated field be removed?

weetmuts: Yes!

weetmuts:
Thanks!

PR #1882, opened 2026-04-25
Migrate ei6500 driver from C++ to XMQ format

Uzlopak:


drivers/src/ei6500.xmq
```diff
@@ -0,0 +1,422 @@
...
+                map {
+                    name  = EOL_WITHIN_12_MONTH
+                    value = 0x00008000
+                    test  = Set
+                }
+                // map {
+                //    name  = SEODS_NOT_YET_COMPLETED
+                //    value = 0x00010000
+                //    test  = NotSet
+                // }
```
Uzlopak, line 367: If I uncomment this, the tests will fail. But this is done in the c++ driver. Why?

drivers/src/ei6500.xmq
```diff
@@ -0,0 +1,422 @@
...
+                vif_range        = Date
+                storage_nr       = 1
+            }
+        }
+        field {
+            name     = dust_level
+            quantity = Text
+            info     = 'Dust level 0 (best) to 15 (worst).'
+            match {
+                difvifkey = 8440FF2C
```
Uzlopak, line 155: is here the measurement_type missing? or is it neglectable?

drivers/src/ei6500.xmq
```diff
@@ -0,0 +1,422 @@
...
+                mask_bits = 0x1f
+            }
+        }
+        field {
+            name     = battery_level
+            quantity = Text
+            info     = 'Battery voltage level.'
+            match {
+                difvifkey = 8440FF2C
+            }
```
Uzlopak, line 169: is here the measurement_type missing? or is it neglectable?

drivers/src/ei6500.xmq
```diff
@@ -0,0 +1,422 @@
...
+                    test  = Set
+                }
+            }
+        }
+        field {
+            name     = obstacle_distance
+            quantity = Text
+            info     = 'The distance to a detected obstacle.'
+            match {
+                difvifkey = 8440FF2C
```
Uzlopak, line 262: is here the measurement_type missing? or is it neglectable?

drivers/src/ei6500.xmq
```diff
@@ -0,0 +1,422 @@
...
+                }
+            }
+        }
+        field {
+            name       = head_status
+            quantity   = Text
+            info       = 'Status of smoke detector sensors, merged into the status field.'
+            attributes = INJECT_INTO_STATUS,HIDE
+            match {
+                difvifkey = 8440FF2C
```
Uzlopak, line 317: is here the measurement_type missing? or is it neglectable?

weetmuts:
I made some changes: https://github.com/wmbusmeters/wmbusmeters/commit/9c1a59ff7ef75e0ea8a4700a19ae2a6fb9ae04e2

Replace quantity=Text with PointInTime for date fields.
Removed the defaults for dif signed=true and vif scaling=auto.
Re-added the comments from the c++ file.

Not sure why you had a problem with the SEODS, it seems to work now.
The bit is 0 when the installation procedure has NOT been performed.
So when it is 1, the warning should not be printed, thus NotSet.

The difvifkey replaces all measurement_type, vif_range, add_combinable settings.
It hardcodes the exact bytes that the field matcher uses, best for manufacturer specific fields like in this case.
I.e. the FF2D, otherwise I have to add many silly MFCT_VIFE_2D MFCT_VIFE_2E etc 256 of them or so...

weetmuts:
Many thanks for taking on this huge meter driver! This smokedetector is pretty impressive!

PR #1883, opened 2026-04-26
Iem3000

Uzlopak:
Please look at the diffs in the commits.

I took the IEM 3000 telegrams from https://github.com/rscada/libmbus/issues/125

Some changes are breaking.

There is a bug it seems.

If the value is passed as 32 bit value, then wmbusmeters considers it as Wh

image

But if it is a 64 bit, then wmbusmeters considers it as kWh

image

Or something like that...


weetmuts:
The 64 bit value is correct: 

```
(gdb) p 0x01FD6C0A / 1000.0
$3 = 33385.482000000004
```
You divide with 1000 since the value sent is Wh but we want to present kWh.

The 32 bit float (real) value also seems to be correct, BUT the meter is making a mistake?

The hex for the float is `4702697F`
got to https://www.h-schmidt.net/FloatConverter/IEEE754.html
and paste this into the ex field.
You get 33385.496 BUT the vif says this is Wh so we have to divide it with 1000 to get kWh.
E.g. we get 33.385496

Did the meter make a mistake...or should wmbusmeters ignore the scale factory Wh in the VIF?
Seems unlikely, I am betting on the meter making a mistake...

PR #1887, opened 2026-04-26
Migrate apator172 driver from C++ to XMQ format

Uzlopak:


drivers/src/apator172.xmq
```diff
@@ -0,0 +1,56 @@
...
+                difvifkey = 0413
+            }
+        }
+        field {
+            name      = total
+            quantity  = Volume
+            info      = 'The total water consumption recorded by this meter.'
+            calculate = '0.1 counter * round(10000 counter * raw_total_m3 / 3 counter)'
+        }
+    }
```
Uzlopak, line 41: i have the feeling that this could be done in using less fields?

weetmuts:
@Uzlopak Thanks! I simplified the driver a bit since we could drop the rounding, I am not sure why I bothered about the rounding, but this is a very old driver. Kept the new round command in the formula calculator, perhaps it will be useful for other things. https://wmbusmeters.org/drivers/apator172.xmq.html

Renamed media_type to force_media_type.

PR #1888, opened 2026-04-27
chore: extract crc16, improve perf

Uzlopak:


weetmuts:
@Uzlopak Thanks, but this is yet another example of why Claude assists the writer of the code but slows down the reviewer of the code.

If this had been a simple re-factor (ie split out the crc functions) I could have approved this in a few seconds.

Now this was combined with a rewrite of the crc code claimed to be a perf improvement, which it perhaps is.

But why is it a perf improvement? It replaced the internal loop with a table lookup. This code was most likely
learned/copied from some other examples on the internet.... so it is probably right. However no comments in the code, nor in the PR why this code i faster, I had to read the source. This sucks.

Also there are no perf tests in the PR to verify the speedup and if it is worthwhile.

I ran a speed test myself on the replaced loop, 1 billion calculations took 2.3 seconds on my laptop. That means the speedup is at most 2.3 nanoseconds, now the speedup is less since you replace this execution with with a memory lookup.

Lets just assume we save 2 ns per byte, on average a telegram is 100 byte, that is we save 0.2 microsecond per telegram.

An average wmbusmeters setup receives maybe 1-10 meters, some have 100 HCA:s listening to them.
Assume meter sends a message once per minute, some more often, some less.
So we have 10 telegrams per minutes on average per setup.
Then we perhaps have 10000 wmbusmeters setups worldwide. Just guessing based on the number of stars 1.3k.

For one hour of running... this perf-improved wmbusmeters would have saved 0.2*10000*60 = 120000 microsconds = 0.12 seconds for all wmbusmeters installations in the whole world. Now I have spent 2 hours thinking about this PR, 
annoyed by AI code etc, thinking about how to require so called perf improvements to be based on an actual need and actual performance test in a real world situation, or a simulated reasonable real world situation, thinking about the Claude cost in electricity etc.

Just taking my 2 hours of of thinking, ie 7200 seconds divided by 0.12 seconds of saving per hour = 60000 hours.

I.e. to recover my two lost 2 hours and break even in time, it takes 60000 hours, that is ~6.8 years. Arghhh.

I did not merge this pull request, I redid the refactoring in 10 minutes manually. Refactoring is probably valuable because it speeds up future compile times and development, which will be much more than the 0.2 microseconds per telegram.

So I will chalk up these 2 hours on the lets learn what not to use AI for account and that I will require perf measurements before accepting perf improvements from AI.

Uzlopak:
@weetmuts 

I think my expectations were not communicated, which i should have. I was expecting a general feedback like "Hey, i dont agree with that lookup table, remove that". 

But I purposely did this PR with two commits. 

The first commit extracts the crc16 out of utils.h. This is what you would have had accepted in seconds. 

The second commit replaces the calculations and moves them into the pre generated lookup table. 

So if you would have given me the feedback that you disagree with the lookup table, i would have just removed the last commit, and had the PR you would expect. 

So I will now revert the last commit and create a new PR to extract the crc16 logic.

Uzlopak:
ah! You did it already

weetmuts:
We are all learning how to deal with AI generated contributions. Some projects have banned them completely: https://simonwillison.net/2026/Apr/30/zig-anti-ai/ 

This one was interesting, because the AI made it so easy to generate a perf improvement, alas to analyze the PR took longer time than all the future savings. (~7 years were if we counted the savings over all wmbusmeter users, saving 0.1 us per telegram for a single use, then we are looking at 70000 years.... :-/ before we recoup the time. :-D

So thank you for this PR @Uzlopak ! Without it I would not have experienced this problem firsthand... now we need to create performance testing as part of the build. Then a simple "make perftest" will show if a fix actually helps or not.

PR #1890, opened 2026-04-27
fun

Uzlopak:


weetmuts:
Thanks! Very interesting to see Claude optimize the source code based on the output from -fdump-tree-optimized

The PR is way too much and messing with the code everywhere, as expected, since Claude got free reins! 
So I tried to cherry pick the performance relevant changes.
 
First I added scripts/perftest.sh that will collect telegrams parts to perf test on and then build src/perf.cc
and run this perf test to calculate the number of nanoseconds spent per byte parsed in dvparser.cc

First I switched from Os to O3 which did give a significant speedup, but increases the binary.

Then I used unordered_map instead of map for dv_entries. 
Then I avoided snprintf inside the dv string generation.
Then I used emplace to avoid copying the DVEntry.

Now the PR code hovers around 210 ns per byte
The master branch now hovers around 220 ns per byte.

Close enough for now. :-)

The PR optimizes quite a lot of other stuff that is really not necessary, but I suppose it is because
it looks at the output from -fdump-tree-optimized

weetmuts:
Also, just pasting parseDV into chatGPT and for hints for speed performance improves gives these tips plus some more.

Uzlopak:
Nice! :)

How much was the perf on master before?

weetmuts:
The speed increase was from 310 ns to 220 ns per byte parsed.

So for a 100 byte telegram, you save ~10 micro seconds. If you receive 100 thousand telegrams per day, you will save 1 second per day. This perf improvement is probably barely noticable because most of the time is spent in field matching and conversion of data, not in the dvparser.

However the unordered_map does speed up other parts of wmbusmeters which is good.

Uzlopak:
Well i think this PR was useful then ;)

PR #1891, opened 2026-04-28
chore: migrate gwfwater driver from C++ to xmq

Uzlopak:
Dont know if we really needed that mfct_error_decoder but hey, there it is.

Also fixing a type from UKNOWN_MFCT_STATUS to U__**N**__KNOWN_MFCT_STATUS, error code hexadecimal is upperacase and not lowercase.


weetmuts:
Thanks! I had to think a bit about this one. First generate UNKNOWN_MFCT_STATUS=02.... and put it into the status is not something I want to do right now. It feels like a risk that values will be exported without proper unit tracking, and that is something we do not want to do.

So I rewrote the new xmq a bit to do without any c++ changes, the new status message for 02 telegrams is now
UNKNOWN_MFCT_STATUS_2

I also merged the two ixml grammars into one.

So, lesson learned, even if Claude manages to solve the problem of porting c++ to xmq, it might do it in a too complex way, because the real solution is to simplify the output where it does not matter.

weetmuts:
The new driver can be seen here: https://wmbusmeters.org/drivers/gwfwater.xmq.html

PR #1892, opened 2026-04-28
chore: migrate mkradio3, mkradio3a, mkradio4  driver from C++ to XMQ format

Uzlopak:
added alot of operators to the ixml to represent this hot mess

fixed the current_date field to derive the year from prev_date


weetmuts:
Very nice! Thank you very much!

PR #1893, opened 2026-04-29
chore: migrate vario451 driver from C++ to XMQ format

Uzlopak:


weetmuts:
Very nice! Thank you!

PR #1894, opened 2026-04-29
chore: migrate compact5 driver from C++ to XMQ format

Uzlopak:


weetmuts:
Very nice! Thank you!

PR #1895, opened 2026-04-30
feat: introduce `critical()` and wmbusmeters specific exitcodes

Uzlopak:
In my opinion, `warning()` and `error()` are not "correctly" implemented. 

In my opinion we should use following systematic approach:
- `warning()` -> something irregular happened, but does not affect the processing, e.g. a settingg was set wrong and we do a fallback to the default value
- `error()` -> something happened, which affects the processing of the current task, but the overall processing is ensured, e.g. invalid telegram got received and we drop it
- `critical()` -> something happened, which not only affects the processing of the current task, but is so disruptive, that running wmbusmeters in sane and deterministic way is not possible

As first step i implemented critical. And changed the occurence of most error calls to critical. 

Next step would be to make some `warnings()` calls to be `error()` calls

btw. no use of LLM.

### DIFF
--- a/src/bus.cc
+++ b/src/bus.cc
@@ -185,7 +185,7 @@ shared_ptr<BusDevice> BusManager::createWmbusObject(Detected *detected, Configur
     {
     case DEVICE_AUTO:
         assert(0);
-        critical("Internal error DEVICE_AUTO should not be used here!\n");
+        critical(EXIT_BUS_DEVICE_ERROR, "Internal error DEVICE_AUTO should not be used here!\n");
         break;
     case DEVICE_MBUS:
         verbose("(mbus) on %s\n", detected->found_file.c_str());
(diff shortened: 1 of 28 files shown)


weetmuts:
Thank you for your enthusiasm but this is mostly a subjective opinion on the meaning of info, warning, error.

info = print always. 

warning = print always because it is something the user should now about.
It might be a single meter fails, but the rest should be running.
Irregular non-compliant telegrams are also warned for.

error = cannot run, exit wmbusmeters.

What you are suggesting is two levels of warnings, but I do not think this is necessary right now.
Thanks anyway!

Uzlopak:
Can we atleast salvage the exitcodes?

weetmuts:
Yes! A more proper list of exit code is valuable!

PR #1896, opened 2026-05-01
chore: avoid try catch in getConfiguredBaudRate

Uzlopak:


weetmuts:
Thanks, but it is easier to change stoi into atoi. :-) The second try/catch can then just be removed.

weetmuts:
https://github.com/wmbusmeters/wmbusmeters/commit/2870575fe801bdbc5b737a01a8e79fc22d94fa69

Uzlopak:
beautiful

PR #1898, opened 2026-05-02
fix: improve  usage of LIST_OF_BUS_DEVICES

Uzlopak:
I am not yet convinced that this is very good approach. I wonder if I can make it more DRY. 

EDIT:
I need this to potentially make rtlsdr and other parts optional.


weetmuts:
This makes it a little bit too much with the X macros.... it is too complicated. No need, these lists change so seldom anyway. I think we should close this one.

Uzlopak:
let me investigate this. i put it into draft

PR #1899, opened 2026-05-03
chore: extract crc16 from util.h

Uzlopak:


Uzlopak:
ah, you did it already

PR #1900, opened 2026-05-03
chore: migrate tsd2 driver from C++ to XMQ format

Uzlopak:
the second and third commit changes meter.cc so that we have a fallback for status-fields if the payload is too small

Or is there some systematic I didnt see?


src/meters.cc
```diff
@@ -1513,6 +1526,14 @@ bool MeterCommonImplementation::handleTelegram(AboutTelegram &about, vector<ucha
     // Invoke any calculators working on the extracted fields.
     processFieldCalculators();
 
+    // Inject ERROR for any STATUS+REQUIRED field that was not matched.
+    if (required_status_field_index_ >= 0)
+    {
+        FieldInfo &fi = field_infos_[required_status_field_index_];
+        if (string_values_.count(fi.vname()) == 0)
+            string_values_[fi.vname()] = StringField("ERROR", &fi);
```
Uzlopak: Maybe this should be configured somewhere... I will see if I can use default_message for this

Uzlopak: I introduced fallback_message

weetmuts:
Thank you! I implemented this in a different way, if an ixml field named mypayload has REQUIRED and the parse fails,  
then add DECODING_ERROR_mypayload to the STATUS field. 

I also fixed the problem with a formula calculating a date which becomes NaN then generate the correct
`"prev_date":null` (instead of the bad  `"prev_date":"null"`

weetmuts:
https://github.com/wmbusmeters/wmbusmeters/commit/6070a684332b92c51add7cebdef6b4641c06a082

PR #1903, opened 2026-05-04
feat: improve cp48 date parsing robustness

### LINKED ISSUE #1364: 32 bit datetime 1 hour offset
### Type of request

Bug

### OS version

linux

### wmbusmeters version

latest(online)

### Your message goes here

Hello,

I had cross-checked one of my decodings on the online decode and think there is an issue with decoding of 32 bit date time which have an 1-hour-offset.

This is the oms field:

```
036   : 04 dif (32 Bit Integer/Binary Instantaneous value)
037   : 6D vif (Date and time type)
038 C!: 3AA81139 ("meter_datetime":"2024-09-17 08:58")
```

I'm not 100% sure because different online decoders gives me a time of 08:58, 09:58 or 10:58 (lobaro decoder).

But the 2nd Byte (0xA3) has bit 7 set (so the 15th bit of the whole filed, starting with zero) and the Spec (EN13757-3-2018, page 37: Date+Time Type F, CP32) says:

```
SU B1 [15]
SU < 0 > : Normalzeit ;
SU < 1 > : Sommerzeit
```

"Sommerzeit" is german for "Daylight saving time": https://en.wikipedia.org/wiki/Daylight_saving_time

So this should be in "Summer time" which is IMO normal time +1h

Uzlopak:
Fixes #1364

### DIFF
--- a/drivers/src/amiplus.xmq
+++ b/drivers/src/amiplus.xmq
@@ -191,7 +191,7 @@ driver {
             args     = 'MyElectricity2 amiplus 00254358 NOKEY'
             comment  = 'amiplus/apator electricity meter with three phase voltages'
             telegram = 5E44B6105843250000027A2A005005_2F2F0C7835221400066D404708AC2A400E032022650900000E833C0000000000001B2B9647000B2B5510000BAB3C0000000AFDC9FC0135020AFDC9FC0245020AFDC9FC0339020BABC8FC100000002F2F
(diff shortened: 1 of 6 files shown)


weetmuts:
It looks nice. What do you want to do more here?

Uzlopak:
@weetmuts 
Some tests are failing. So there is still something odd there

Uzlopak:
This doesnt work, because the time is not in utc but local time.

Uzlopak:
@weetmuts 

This is actually ready for review. But to be honest it is alot of braingymnastics to understand if I did it right. 

The tests tun with UTC timezone for normalized results. And the spec itself says that DST flag is only informational. So should it change in the tests or not? 

Yesterday it made sense.

I have to reconsider it.

weetmuts:
It seems like both the DST and the leap year bit is purely informational. Why do we care? This information
is not going to be printed in the output json anyway.... 

Do you have an example of a situation where the current decoder fails?

Uzlopak:
@weetmuts 

Three weeks since i touched that code is a long time. I try to remember what our issue here is.

If you look at the tests, you see that some telegrams are 1 hours off. Our tests runs with UTC timezone. In such a case DST needs to be accounted and the offset needs to be corrected.

weetmuts:
Yes, I am reading the other issue where there are test cases. Perhaps it is because the daylight savings bit must be accounted for when converting to wmbusmeters default unit for PointInTimes which is unix timestamps.

Need to carefullt study the docs here.

PR #1905, opened 2026-05-04
chore: migrate izar driver from C++ to XMQ format

Uzlopak:


weetmuts:
I think we should make a new version of the izar driver. It is pretty old and does not conform the the new standards.
In particular the flags should be upper case and separated by space and all stored in the status field.

So no need for a separator setting (comma here) yet for lookups. leakage previously becomes PREV_LEAKAGE

Also instead of -1 for wildcards in mvt:s lets use the asterisk `*`

The default fields should be much fewer:
`name,id,status,total_m3,target_m3,timestamp`
the default fields should always be as few as possible.
target_m3 should be the new name for last_month_total_m3
target_date should be the new name for last_month_measure_date
battery_y for remaining battery
current and previous alarms are both merged into status
manufacture_y or manufacture_date ?

So lets create izarv2 and mark the old for deprecation.

PR #1907, opened 2026-05-04
Fixes for the KEM importer

jktjkt:
### KEM import: fix working with relative paths
    
Previously, the default behavior would take an "absolute path" such as `/home/foo/wmbusmeters`, strip the leading slash, which yields `home/foo/wmbusmeters`, and use that as a path relative to the current working directory. This in turn created files on disk named, e.g., `/home/foo/wmbusmeters/home/foo/wmbusmeters/etc/wmbusmeters.d/12345678`.

### KEM: store meter type/config as a comment as well
    
 Right now these fields are of no use for wmbusmeters, but I think it's generally useful to keep information about the configuration of any given device. Given that Kamstrup decided to put this info into the KEM  files, let's at least put that into a comment field.

### KEM: generate a more generic config for modern Kamstrup water meters
    
The meter IDs are a bit more generic, so let's detect them based on a generic prefix. Also point to a generic driver name, not to an alias that implies that it's a specific driver for an older generation of meters.

### DIFF
--- a/utils/kem-import.py
+++ b/utils/kem-import.py
@@ -178,6 +178,7 @@ def print_meter(meterName,meterType,meterNum,meterSerial,meterVendor,meterConfig
         if (wmbusmeters_driver is not None):
             try:
                 f = open(args.config+meterSerial, 'w')
+                f.write("# %s type %s config %s\n" % (meterName, meterModel, meterConfig))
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks!

PR #1909, opened 2026-05-05
chore: set optimization to -O2, set -flto and -fdump-tree-optimized

Uzlopak:
Optimization level -O3 is too excessive  and extend the compile time unnecessarily. Potential gains, if there are any, are neglectable.

Activating -flto reduces the size of the final release executable by about 500 kb.

-fdump-tree-optimized "dumps" the preprocessed and by the compiler generated constexpr files in the build-folder. This can be used to examine potential performance and size improvements.


weetmuts:
Thanks!

weetmuts:
With this fix, the compile time is slower. Using time we get:
Before fix and -O3: user time 3m24s
After fix with -O2 and -flto: 4m36s
After fix with -O3 and -flto: 4m48s

No wonder, -lto makes the linking take a lot more time.

The time difference between O2 and O3 is minimal.

PR #1910, opened 2026-05-05
fix: implementation of "exp", rename to "pow" and change operator character to "**"

Uzlopak:
exponentiation was since for ever broken.

I fixed it and because it never worked in the first place, took the chance and replaced the operator name from `exp` to `pow` and the "^" character to "**" characters. 

As it was broken from the beginning, there is no chance of introducing a regression. 

The name `exp` and `exponentiation` is misleading, because exponentiation means e^x, while e is Eulers number, and pow means x^y. 

The reason for changing the character is the problem, that in c-like languages `^` is used for XOR. C and C++ do not have an pow operator, while other languages like python and javascript use the `**`. I guess because  exactly the same reason, because `^` would collide with the XOR operator.

### DIFF
--- a/src/formula.cc
+++ b/src/formula.cc
@@ -377,7 +377,7 @@ const char *toString(TokenType tt)
     case TokenType::BOR:  return "BOR";
     case TokenType::LAND: return "LAND";
     case TokenType::LOR:  return "LOR";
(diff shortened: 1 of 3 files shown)


weetmuts:
Thanks!

PR #1911, opened 2026-05-05
fix: make should not set `-fdump-tree-optimized` if on non linux system

Uzlopak:
The CI is red, because the compiler for mac is not know `-fdump-tree-optimized`. So we just add `-fdump-tree-optimized` if we are on Linux

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,11 @@ DEFAULT_CONF_DIR?=/etc
 DEFAULT_DEAMON_DRIVER_DOWNLOAD_DIR?=/var/lib/wmbusmeters/wmbusmeters.drivers.d/downloaded
 DEFAULT_USER_DRIVER_DOWNLOAD_DIR?=.local/share/wmbusmeters/wmbusmeters.drivers.d
 
-DUMP_FLAGS=-fdump-tree-optimized
+DUMP_FLAGS=
+
+ifeq ($(shell uname),Linux)
+    DUMP_FLAGS=-fdump-tree-optimized
+endif
 
 include $(wildcard build/*/spec.gmk)
 


jktjkt (changes requested):
First of all, this particular compiler option is GCC-specific, not OS-specific in any way. If you want to make its activation conditional, the appropriate way is to detect your *compiler*, not the *operating system*. Also, this detection should happen at *configure* time, not at *build* time.

But the more important question is *what do you need this option for in the first place*? [Your PR which added this option](https://github.com/wmbusmeters/wmbusmeters/pull/1909) only says:

> This can be used to examine potential performance and size improvements.

That's quite vague. It doesn't explain how you're going to use this output and through which tools, nor why you think the other developers should have that option activated, nor whether there are any measurable gains from enabling it.

TL;DR: just drop it unconditionally. If you think you need it, enable it yourself for whatever purpose you require; it's a trivial modification of CXXFLAGS after all.

Uzlopak:
I am not like you and always on master branch. Its quite annoying to always patch the makefile. 

I will investigate this regarding configuring at configure time

jktjkt:
> I am not like you and always on master branch. Its quite annoying to always patch the makefile.

There’s no need to modify any files; just export whatever you need in CFLAGS/CXXFLAGS as an env var in your shell.

weetmuts:
Fix pushed: https://github.com/wmbusmeters/wmbusmeters/commit/dc6cc26bb70278482cb570f88b5a1acc9752f963

weetmuts:
I also remove -flto from the default settings. It is already passed by Fedora build as CXXFLAGS on the command line.
It is passed as CXXFLAGS when building the wasm build. 
It conflicts with the docker build settings.

weetmuts:
I kept the dump optimized tree because I find it interesting, but again it might be removed because it is not needed for most builds.

@Uzlopak you just do: export CXXFLAGS="-flto -fdump-tree-optimized"
before "make -j4"

PR #1917, opened 2026-05-06
migrate fhkvdataiii driver from C++ to XMQ format 

Uzlopak:
#1800 needed changes in c++, this one is pure xmq.


weetmuts:
Thanks!

PR #1919, opened 2026-05-07
hcae2: real-meter tests, error decoding, half-monthly delta history

felixhauptmann:
## Summary

Overhaul of the `hcae2` driver. Essentially a full rewrite.

- cleans up `eurisii` copy-paste artifacts from PR #956
  - replaces synthetic/irrelevant tests with real-meter captures
  - removes redundant parser logic not applicable to HCAe2
  - removes consumption_at_set_date storages that carried half-monthly deltas, not billing snapshots
- migrates to using library functions for common fields
- decodes the manufacturer-specific error byte
- exposes the half-monthly delta history

## Open decisions

- **Breaking removal of `consumption_at_set_date_{2..17}` without deprecation.** Fields carried half-monthly deltas mislabelled as billing snapshots; deprecation cycle would prolong wrong behaviour.
- **Half-monthly delta fields tagged `attributes = DEPRECATED` as "do not rely yet" signal.** No `EXPERIMENTAL` PrintProperty exists. (Worth introducing one at some point?)
- **Review and merge now, or wait for the captures below?**

## Captures needed

The half-monthly delta model has open assumptions (documented in-source). This does not affect parsing/field names though. To settle them:

- ActiveLong follow-up (2026-06) from the same meter (via @Skysurfer-14)
- annual billing, no reset (x2, ~1 month apart)
- monthly billing, with reset (x2, ~1 month apart)
- monthly billing, no reset (x2, ~1 month apart)

If anyone has long-running meters with the respective configuration, feel free to help!


weetmuts:
Very nice! My only request is that you name this new driver hcae2v2 (ie the suffix v2 is added)
and the old driver remains.  Then move the detect mvts to the new driver and remove them from the old driver.

The reason is that the most important value has its key name changed from `current_consumption_hca`
to `consumption_hca`.

This is a good change, but wmbusmeters tries to avoid changing the important fields names within the same driver name.

I will add deprecated warnings to old drivers in the future.

felixhauptmann:
That's why I deprecated `current_consumption_hca`. It is still there.

I do break one thing though: I did not deprecate the fields that were previously called `consumption_at_set_date_N` which actually do not contain "consumption at set date" (those are half-monthly delta values; completely useless if interpreted as meter readings).

Instead I fully removed them apart from `consumption_at_set_date_1` which contains correct data. I can bring those back and deprecate them instead. In that case to my knowledge the driver would be fully backwards compatible.

If you still want to go the v2-route just tell me and I'll update the PR.

weetmuts:
Ah, yes I see. Hmmm, deprecating a field in an existing driver, is that worse than deprecating a whole driver...
no probably not. Someone will have to change something before a certain date.

So, yes keep the hcae2 name.

We will have to propagate warnings about deprecated fields and drivers to the end user.
Should we add this to the json:
"warning":"deprecated field current_consumption_hca will be removed 2026-11-01" 

and for a whole driver:
"warning":"this driver is deprecated and will be removed 2027-02-01"

or something else?

felixhauptmann:
> Someone will have to change something before a certain date.

Agreed!

> We will have to propagate warnings about deprecated fields and drivers to the end user. Should we add this to the json: "warning":"deprecated field current_consumption_hca will be removed 2026-11-01"
> 
> and for a whole driver: "warning":"this driver is deprecated and will be removed 2027-02-01"
> 
> or something else?

I would say emitting a warning in the log would be good enough: if a consumer does not look at their logs, they certainly won't look into raw JSON output.

I don't know how wmbusmeters is typically used, although I have seen some mentions of HomeAssistant. Does HomeAssistant read and surface logs? If so, and if that applies to other dependents as well, logging should suffice, right?

If machine-readable parser-related messages are needed, it should probably be an array, so something like this works:

```json
{
    "diagnostics": [
        { "warning": "deprecated field current_consumption_hca will be removed 2026-11-01" },
        { "warning": "this driver is deprecated and will be removed 2027-02-01" },
        { "info": "use of experimental field delta_1_half_months_ago" }
    ]
}

PR #1921, opened 2026-05-08
Compact frame formats

Uzlopak:
While working at the TPL 73 PR, I realized that the compact frames should be part of the xmq and not of wmbus.cc. 
It is currently in draft because of two aspects:
1. kamheat and kamwater have compact formats where it is not clear if it should be for water or heat. needs to be differentiated
1. should we already put mvt as prefix of the compact formats? So if two devices have the same crc it doesnt matter because we differentiate by mvt.


weetmuts:
I think storing the signature per driver is reasonable...

As you say, since the hash is short, there is a small risk that two formats map to the same signature if we maintain a global database of signatures. Much less risk within a single driver.

weetmuts:
Separating with mvt seems a bit too specific, the meter manufacturers put out new versions all the time...

jktjkt:
Since the compact format signature/hash is the CRC of the DIF/VIF/VIFE sequence from a full packet, would it make sense to not list the hash in the XMQs and instead compute it at startup (or at compile time if you want to be very fancy) from the original sequence?

weetmuts:
@jktjkt Good point, also we could have the driver wait to calculate the crcs of the formats stored in the driver until a compressed telegram actually arrives.

weetmuts:
I also think that the test cases should exercise the formats, so that we have actual telegrams with the format supplied.
If we have a signature+format which is not exercised in the test cases, then a warning/error should be printed.

Uzlopak:
@jktjkt 
@weetmuts 

PTAL

Uzlopak:
It does not calculate the crc16 at compilation time, as drivers should be able to be loaded from a remote source. So it makes sense to let them run in the same procedure

I wonder if the current implementation is really better:

Right now.
```
    compact_frame_formats {
        // Kamstrup Multical 303
        difvif = 0406041404FF0704FF080259025D023B02FF22026C44064414426C
    }
```

I wonder if my previous approach is maybe better

```
    compact_frame_formats {
        compact_frame_format {
            // Kamstrup Multical 303
            difvif = 0406041404FF0704FF080259025D023B02FF22026C44064414426C
        }
    }
```

Then we could improve it by setting the mvt to narrow it down or so. But I am inconclusive.

weetmuts:
Very nice! Thanks!

weetmuts:
I think that per driver registered mvts is fine. Does this work if you force a driver for some other mvt?

PR #1923, opened 2026-05-09
chore: remove error-artifact

Uzlopak:
error should always use a exit-code. previous commit did not remove the overloaded error function but commented it out. this removes the commented out code.


weetmuts:
Thanks!

PR #1924, opened 2026-05-10
chore: avoid try catch in `DriverDynamic::load`

Uzlopak:
avoids try-catch-logic 
Also seems to be better encapsulating the logic.

The next step would be to implement something similar for the DriverDynamic constructor. 

Then we should be basically done in removing the exception handling.

### DIFF
--- a/src/driver_dynamic.cc
+++ b/src/driver_dynamic.cc
@@ -26,11 +26,13 @@
 
 using namespace std;
 
-string check_driver_name(const char *name, string file);
-string check_aliases(const char *aliases, string file);
-MeterType check_meter_type(const char *meter_type_s, string file);
-string check_default_fields(const char *fields, string file);
-void check_detection_triplets(DriverInfo *di, string file);
+bool process_driver_name(DriverInfo *di);
+bool process_aliases(DriverInfo *di);
+bool process_meter_type(DriverInfo *di);
+bool process_default_fields(DriverInfo *di);
+bool process_force_media_type(DriverInfo *di);
+bool process_mvt(DriverInfo *di);
+bool process_mfct_tpl_status(DriverInfo *di);
 
 string check_field_name(const char *name, DriverDynamic *dd);
 string check_field_ixml(const char *ixml, DriverDynamic *dd);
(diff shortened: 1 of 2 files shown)


weetmuts:
I think that we either want exceptions or long jump here. 

This is a typical case where we just want to escape the driver parsing without further ado when we find an error deep inside the driver. The exception/long jump is the best way to do this programmatically.

Since we are removing the exceptions (there were very few to start with) from wmbusmeters to reduce the wasm binary, I suggest we try long jumps here. Assuming it actually reduces the size of the wasm binary.

Uzlopak:
I have just a slight idea what you probably mean with long jump. 

But isnt this what is done here?

The throw is basically interrupting the flow and we jump to the error handling. 
Here we we exit early, because if process_x is returning false, the following process_y function wont be called. So then we just  release the xmqDocument by removing the reference to it by setting it to NULL.

weetmuts:
What is the size saving in the wasm binary if we get rid of exceptions? Must all exception handling be removed or is it a saving per place where we use exceptions? This is the only remaining one as far as I know. 

My questions is, is it worthwhile to do this?

Uzlopak:
I am not targetting primarily wasm. If we remove exception handling wmbusmeters will reduce by atleast 200kb. my assessment is from 2.6 MB to 2.26 MB.

weetmuts:
Yeah, I have thought about it now and I do not think it is worthwhile. 200Kb is just the cost of C++.

The driver_dynamic loading code becomes less readable. You have not removed all throws in the PR either, so you have quite a lot of more work to do in the PR. :-)

The saving is not guaranteed. 
https://dev.to/sandordargo/binary-size-and-exceptions-2cp3

And perhaps we will need exceptions some time in the future and then we will have to reintroduce them but the driver_dynamic code will stay the less readable version.

weetmuts:
wmbusmeters is still smaller than the average web page in 2026.... :-D

https://httparchive.org/reports/page-weight

Uzlopak:
Ok, let me fix the other parts too.

weetmuts:
Please do not spend any more time on this.

weetmuts:
Thank you very much for your time thinking about this, and I have thought about it a lot too.

But I prefer to keep the exceptions...for now. I will keep this in mind however.

PR #1926, opened 2026-05-11
chore:  Migrate hydrodigit driver from C++ to XMQ format

Uzlopak:
I needed to changed the values of fraud_type, because the xmq drivers will sort the values alphabetically.

So "no type info" became "NO_TYPE_INFO".


weetmuts:
Amazing work @Uzlopak! But why did you have to introduce mkdate? 
Previously you used:
`2000-01-01+(x * 1y)+(y * 1m)+(z *1d)`
To generate a new date, is that not possible here?

I am not saying mkdate is a bad idea, just curious.

weetmuts:
Ah, now I see. You need to generate a null date when when month and day are 0....
Ok, I have added year adding to formulas anyway. :-) 
So now `'2000-05-05' + 2y` results in the date `'2002-05-05'`

weetmuts:
Thanks!

weetmuts:
Pretty impressive what can be done with ixml+wmbus+formula decoding! 

Some things can be simplified:
`calculate = 'backflow_raw_counter / 1000counter * 1m3'`
this can be calculated more efficiently with `force_scale=0.001` no need for a formula.

These I do not understand:
`calculate = '((january_total_raw_counter < 10000000counter) * (january_total_raw_counter / 100counter)) * 1m3'`

If the total exceeds a certain value, then total is fored to zero forever?

PR #1928, opened 2026-05-11
fix: tests should use UTC timezone

Uzlopak:
Testing should use UTC timezone

I think this would make it possible to merge 
#1927 
#1903


weetmuts:
Thanks!

PR #1929, opened 2026-05-12
chore: reduce warnings

Uzlopak:
here a convoluted pr for reducing warnings. cherry pick if you like :)


Uzlopak:
we have probably better things to do

PR #1930, opened 2026-05-12
chore: remove `language()` from util.cc

Uzlopak:
`language()` from util.cc is not used anywhere in the code.


weetmuts:
The drivers are supposed to have multi-language descriptions, but we can re-add this if we get there.

PR #1933, opened 2026-05-13
fix: make generated_database a header file, generate proper copyright header

Uzlopak:


weetmuts:
Thanks!

weetmuts:
Github no longer can handle large diffs so the merging in the github ui is blocked! Yikes.
Anyway, I merged your improvement separately, however I kept generated_database.cc

PR #1934, opened 2026-05-13
fix: make extract_days have const parameter and fix "say" to "sat"

Uzlopak:
I guess, you dont like it, but this makes it for me read easier and we are not manipulating the string


weetmuts:
Heh, you are right. :-) I see no reason to merge this PR. In general, focus on real refactoring. Splitting large files into smaller files. Changing this kind of low level stuff on code paths like this, just risks introducing bugs.

PR #1939, opened 2026-05-14
chore: improve fuzzing

Uzlopak:
I have no deep understanding if the bug fixes are proper.

adds a github workflow for fuzzing
adds a fix for a bug in dvparser
adds a fix for two crashing bug in formula
add some more fuzzer

See https://github.com/wmbusmeters/wmbusmeters/actions/runs/25853748718/job/75966333509 for a workflow run.


Uzlopak:
Acciddently merged master. Will rebase

weetmuts:
Great work! Please comment out the fuzzing of formulas and addresses, and the fixes.
But keep them in the code. I will take a look if those
are real problems or artifacts of the fuzzing. 

In particular, I am suspicions of dropping an assertion
since asserts mean that the code is not used in the expected way.
I will re-enable the fuzzing of formulas and addresses.

You can remove the fuzzing of frames. These are completely handled inside fuzzing of telegrams already.

weetmuts:
I fixed the real assert crash caused by a conflict between a synthetic and real telegram difvif entry by
adding a new wmbusmeters specific Combinable Synthetic which is added to the synthetic entries.
The match rule to match is then required to use add_combinable = Synthetic 

The reduces the risk for conflict to zero for all normal telegrams. For bad telegrams, the fix of adding more
suffixes is used and a warning is printed.

Modified driver:
https://wmbusmeters.org/drivers/caltose.xmq.html

weetmuts:
https://github.com/wmbusmeters/wmbusmeters/commit/97d26246dfbffdaf9b15417a0b0374396e1769d4

Uzlopak:
@weetmuts 

Should I also revert the asserts in formula_implementations.h?

Uzlopak:
@weetmuts 

Is there something in this PR you would like to integrate? Should I/you close it?

weetmuts:
Yeah, I do not want to fuzz address and formula right now. They are supplied from the wmbusmeters config side, ie by someone who already has control over the system and software.

The fuzzing is to make sure it is hard to crash wmbusmeters with a maliciously crafted radio telegram....

I do not think github is really happy with me running a fuzzing sessions for a looong time inside github.

When I run fuzzing, I let it run for several weeks on my own server.

Uzlopak:
Oh, ok. But the potential findings were solved, right?

weetmuts:
Yes! I fixed them.

PR #1943, opened 2026-05-15
Changes related to Apator

Uzlopak:


weetmuts:
This code has the crc16 changes that I said no to before. Can you remove please.

Uzlopak:
@weetmuts 

I did not change crc16 like the last PR. Apator uses  a  crc16 with different poly and init values. So I unified the two crc16 functions and made the crc16 use msb instead of lsb.

weetmuts:
Can you split them please? I want to keep the wmbus standard compliant code separate from manufacturer specific code.
The apator code might go away in the future, when there are no more old apators running.

PR #1944, opened 2026-05-16
chore: add more status fields for kamwater, add compact format for flowIQ2200

Uzlopak:


drivers/src/kamwater.xmq
```diff
@@ -49,22 +51,78 @@ driver {
...
+                map {
+                    name  = HARDWARE_FAILURE
+                    value = 0x00004000
+                    test  = Set
+                }
+                map {
+                    name  = TAMPER_DISPLAY
+                    value = 0x00008000
                     test  = Set
                 }
```
jktjkt, line 127: This part of the diff is about the "cold env", and in there the there are no alarm bit for `TAMBIENT_HIGH`, `INTERNAL_ERROR`, `FLOW_CAVITATION`, `HARDWARE_FAILURE` or `TAMPER_DISPLAY`.

drivers/src/kamwater.xmq
```diff
@@ -80,26 +138,82 @@ driver {
...
+                map {
+                    name  = HARDWARE_FAILURE
+                    value = 0x4000
+                    test  = Set
+                }
+                map {
+                    name  = TAMPER_DISPLAY
+                    value = 0x8000
                     test  = Set
                 }
```
jktjkt, line 218: This is called the legacy info code, and these bits actually store an accumulated time over which the DRY/REVERSE/LEAK/BURST/TAMPER conditions were happening. There are no error bits for, say, a low battery (or really any error conditions that you're adding here) in this particular register.

drivers/src/kamwater.xmq
```diff
@@ -113,27 +227,84 @@ driver {
...
             }
```
jktjkt, line 310: Isn't it better to leave a deprecated field as-is?

jktjkt (changes requested):
Where did you take these bitfield descriptions from?

The doc which Kamstrup sent me a few months ago says that there are five possible status fields:
- `02ff20`, which is the legacy info
- `04ff23`, aka info registers for the cold environemnt
- `04ff24`, aka info registers for hot env
- `02ff25`, info registers for fixed networks (no time info)
- `12ff25`, info registers for fixed networks (no time info, but with logged info codes)

There's quite a lot of errors in this patch, so I wonder what source of information you used?

Uzlopak:
Let me investigate this.... will come back to you.

Contacted you on linkedin :)

jktjkt:
> Let me investigate this.... will come back to you.
> 
> Contacted you on linkedin :)

Please drop me a mail at jkt@jankundrat.com if you need something which should not be done over GitHub. I’m not really following LinkedIn I’m afraid.

weetmuts:
I have tried to avoid adding fields to drivers unnecessary in wmbusmeters. All new fields should be driven by real telegrams collected by users and their desire to read the values. The users have then been able to verify that the fields are correct.

The reason is that it is hard to guarantee correctness otherwise. 

It might be tempting to ask Claude to make the adjustments to a driver based on documents, but I think this is dangerous.

It takes a lot of manual time to verify that the fields are correct......

Better with fewer fields that are correct, than many fields where some are wrong.

I am going to implement testing requirements that a driver is not valid unless
there are test cases for each field. (This is a problem for the status bits though,
but the requirement will have to be a bit more relaxed there.)

Issue #1575, opened 2025-08-26
ei6500 dust levels


H4PPYX:
### Type of request

Existing device improvement

### Meter description

I have information that the device dust level is 0-31 and not 0-15. 

### Logged telegrams from meter in decrypted format

```shell
5E442515410364070C1A7AD50050252F2F_0BFD0F000102046D0E0A3A3802FD17000082206C3538426C01018440FF2C000F11008250FD61000082506C01018260FD6100008360FD3100000082606C01018270FD61030082706C35382F2F2F2F
```

Uzlopak:
You are definetly correct.

in the izar@mobile app ei6500 the dust level goes from 0 to 31 and above 28 recommend to clean it.

weetmuts:
The code is correct since the bitmask is 0x1f which allows 0-31.
The comment is wrong since it says 0-15.

PR #1946, opened 2026-05-19
feat: improve ei6500 driver

Uzlopak:
ERROR_FLAGS are now set properly. dust_level_counter is actually 0-31 as it was correctly claimed here: https://github.com/wmbusmeters/wmbusmeters/issues/1575 When Dust counter is >= 28, add CLEANING_RECOMMENDED  as status flag


weetmuts:
We cannot change the behavior of an existing driver, if we want to do that then we must create a new driver.
We can add new fields in an existing driver.

Uzlopak:
@weetmuts 

Should be now non breaking change. dust_level is just marked as deprecated.

weetmuts:
Very nice!

PR #1949, opened 2026-05-19
fix: run specific hub workflows only on wmbusmeters/wmbusmeters repo

Uzlopak:
Committing and pushing on forks result in failed workflow runs, as those workflows should only run i then wmbusmeters/wmbusmeters repo.


weetmuts:
Thanks!

PR #1950, opened 2026-05-19
chore: move TestBit to translatebits.cc/.h, OutputFormat to wmbus.h

Uzlopak:
They actually belong in those files and not into util.h


weetmuts:
Thanks!

PR #1956, opened 2026-05-22
fix: dlms and also dlms workflow

Uzlopak:
the point of the workflow was to have automatic updates. Till today it didnt had one successful PR 

This should update the manufacturer codes and also fix the workflow.

### DIFF
--- a/.github/workflows/update_manufacturers.yml
+++ b/.github/workflows/update_manufacturers.yml
@@ -1,8 +1,6 @@
 name: Update DLMS list
 
 on:
-  schedule:
-    - cron: '0 14 * * 5' # Every Friday at 14:00 UTC
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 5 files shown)


weetmuts:
But you are removing the cron job from the workflow? This means it will be disabled. Why?

Uzlopak:
Because you have to enable the option to create PullRequests by Github Actions in the Repository Settings. 

https://github.com/wmbusmeters/wmbusmeters/settings/actions

Scroll down and you should see the checkbox `Allow GitHub Actions to create and approve pull requests`

image

If you activate it, then the cron job can actually work.

Uzlopak:
If you want i can revert the cron job removal commit.

weetmuts:
Yeah, thanks for the effort on the dlms. But I reverted it because there were so many problems, so now the
"make update_manufacturers" does the correct thing and I can just commit the diff generated.

Issue #459, opened 2022-01-31
Claim multical403 is the correct driver when multical603 more correct


petterreinholdtsen:
I have successfully converted my .zip.kem file to a wmbusmeters.d, and get some values from my meters.  But when I start it up, it complain that the correct type should be multical403 and not multical603, while only using multical603 will give me a non-zero value for total_energy_consumption_kwh.

Here is an example run and output:

```
./build/wmbusmeters --useconfig=/home/pere/.config/wmbusmeters/ --device=rtlwmbus
Started config rtlwmbus[00000003] listening on none
(meter) 71054899: meter detection did not match the selected driver multical603! correct driver is: multical403
(meter) Not printing this warning again for id: 71054899 mfct: (KAM) Kamstrup Energi (0x2c2d) type: Heat volume at inlet meter (0x0c) ver: 0x34
{"media":"heat volume at inlet","meter":"multical603","name":"71054899","id":"71054899","total_energy_consumption_kwh":32175,"total_volume_m3":3591.16,"volume_flow_m3h":0,"t1_temperature_c":31.45,"t2_temperature_c":35.79,"at_date":"2022-01-31 00:00","current_status":"","energy_forward_kwh":153189,"energy_returned_kwh":124887,"timestamp":"2022-01-31T15:05:00Z","device":"rtlwmbus[00000003]","rssi_dbm":140}
[...]
```

Perhaps some detection logic should be adjusted?

weetmuts:
Can you post a logtelegrams=true output here:
ie telegram=|.....|

Do you have a multical603 meter?

petterreinholdtsen:
[Fredrik Öhrström]
> Can you post a logtelegrams=true output here:
> ie telegram=|.....|

I'm not quite sure how to enable this.  Tried adding it to my
~/.config/wmbusmeters/etc/wmbusmeters.conf but got no such dump.

> Do you have a multical603 meter?

My meter modell is 403WD0244 according to the front of the meter, but
using the multical403 driver did not give me the number I needed.
-- 
Happy hacking
Petter Reinholdtsen

weetmuts:
You could run it from the command line just to get a couple of logged telegrams.
wmbusmeters --logtelegrams rtlwmbus:c1,t1 MyMeter multical403 71054899 <yourkey>

petterreinholdtsen:
[Fredrik Öhrström]
> You could run it from the command line just to get a couple of logged telegrams.
[...]

Issue #344, opened 2021-09-24
Add Multical 602 Heat Meter


LarsBorris:
Hello!

What I want to do:
I have a Kamstrup Multical 602 heatmeter.
I want to use wmbusmeters to read the wm-bus telegram and feed it to my home assistant using mqtt. This is my goal.

I have the Amber 8665-M dongle.
Flashed the latest firmware on it yesterday from windows.
All ok!

I have installed (build from source) wmbusmeters on my laptop running Ubuntu.
The dongle is located in /dev/ttyUSB0:

mesg | grep tty
[    0.140155] printk: console [tty0] enabled
[  660.963692] usb 1-4: FTDI USB Serial Device converter now attached to ttyUSB0

(I have read that the AMB8465 and AMB8665 are the same but the AMB8665 has external antenna, is that correct?)

CONF:
loglevel=normal
device=/dev/ttyUSB0:amb8465:c1,t1
logtelegrams=false
format=json
meterfiles=/var/log/wmbusmeters/meter_readings
meterfilesaction=overwrite
logfile=/var/log/wmbusmeters/wmbusmeters.log

Output when running wmbusmeters:
wmbusmeters --debug auto
(wmbusmeters) version: _
(config) number of meters: 0
(main) device /dev/ttyUSB0 not currently used, detect contents...
(serial) EVENT thread interrupted
(amb8465) could not open tty for detection
(serial) EVENT thread interrupted
(serial) EVENT thread interrupted
(serial) EVENT thread interrupted
(main) ignoring /dev/ttyUSB0, it does not respond as any of the supported wmbus devices.
No wmbus device detected, waiting for a device to be plugged in.
(shell) exec (capture output) "/bin/pidof"
(shell) arg "wmbusmetersd"
(shell) output: >>><<<
(shell) return code 1
(shell) 14223 exited
(shell) exec (capture output) "/bin/pidof"
(shell) arg "wmbusmeters"
(serial) EVENT thread interrupted
(shell) output: >>>14219
<<<
(shell) return code 0
(shell) 14224 exited
(serial) registered regular callback HOT_PLUG_DETECTOR(0) every 2 seconds
No meters configured. Printing id:s of all telegrams heard!
(serial) waiting for stop
(serial) EVENT thread interrupted
(serial) TIMER thread interrupted
^C(serial) stopping manager
(serial) TIMER thread interrupted
(serial) EVENT thread interrupted
(serial) event loop stopped!

[...]

Issue #181, opened 2020-11-04
Support for Multical 803 heat meter


Nikodemski2002:
Hi Weetmuts. 

As you can guess, I'm trying to add support for Kamstrup Multical 803 heat meter. I've read through the #165 issue, but my inexperience in terms of m-bus protocol, or overlooking something left me stuck. 

1. So as in #165 I've made a copy of the meter_multical403.cc file, and replaced all instances of 403 with 803.
2. I've added the lines for multical 803 in meters.h:

LIST_OF_METERS:
_X(multical403,C1_bit, Heat, MULTICAL403, Multical403)
X(multical603,C1_bit, Heat, MULTICAL603, Multical603)
X(multical603,C1_bit, Heat, MULTICAL803, Multical803)_

_METER_DETECTION:
X(MULTICAL403,MANUFACTURER_KAM, 0x0d, 0x34)
X(MULTICAL603,MANUFACTURER_KAM, 0x04, 0x35)
X(MULTICAL803,MANUFACTURER_KAM, 0x04, 0x39)_

3. I've set the meter file to:
_name=79806477
type=multical803:c1
id=79806477
key=00112233445566778899AABBCCDDEEFF_
4. When I used the `wmbusmeters /dev/ttyUSB0:im871a:c1 79806477 multical803:c1 79806477 00112233445566778899AABBCCDDEEFF` 
I didn't receive any data, also when I tried resetting the daemon or the whole RPi, the same scenario when I changed the meter type in the meter file to multical603:c1 (and of course in the listening command too).
5. When I changed the meter type in the meter file, and command to multical403, I received mixed data, as shown below:
```
Started config im871a on /dev/ttyUSBO listening on c1
(meter) 79806477: meter detection did not match the selected driver multical403! correct driver is: unknown!
(meter) please consider opening an issue at https://github.com/weetmuts/wmbusmet ers/
(meter) to add support for this unknown mfct,media,version combination
(meter) 79806477: meter detection did not match the selected driver multical403! correct driver is: unknown!
(meter) please consider opening an issue at https://github.com/weetmuts/wmbusmeters/
(meter) to add support for this unknown mfct,media,version combination
(meter) 79806477: meter detection did not match the selected driver multical403! correct driver is: unknown!
(meter) please consider opening an issue at https://github.com/weetmuts/wmbusmeters/
(meter) to add support for this unknown mfct,media,version combination
[...]

Issue #48, opened 2019-11-20
multical302 volume unit issue


mbursa:
There seems to be an issue with volume units in multical302 meter.

I am getting the following data: {"media":"heat","meter":"multical302","name":"67947613","id":"67947613","total_k
wh":305,"total_volume_m3":1830,"current_kw":0,"timestamp":"2019-11-20T18:29:15Z"
}
The display of the device shows 18.30 m3. The value reported by wmbusmeters thus should be divided by 100.0
I have the meter for two weeks now and its counters had started from zero, so volume around 18 m3 is much more sensible than 1830 m3.

weetmuts:
Can you provide a few log telegram lines from your log when you have enable --debug or --logtelegrams. They look something like this:

(multical302) log "telegram=|4E44.....00|+0" 

Thanks!

weetmuts:
I got some other logs and could add proper support for Mulitcal302. 

Note that the output will be zero until you have received a full telegram.
After the full telegram has been received, then it is possible to decode the compressed
telegrams. I believe every 8th telegram is a full telegram.

Please test and verify that this. If it works properly, then we can add some
hardcoded hashes->formats for multical302 so that it can immediately start
decoding the compressed frames.

mbursa:
telegrams from meter no. 1:
|25442D2C4576946730048D20B791A20B20|D37E79E7F194310500000000002A0000FFFF0A0000|+1
|25442D2C4576946730048D20C0C0A20B20|923F79E7F1694E0500000000002A0000FFFF090000|+19
|25442D2C4576946730048D20DF40A30B20|1EE979E7F17E880500000000002C0000FFFF0C0000|+148
|25442D2C4576946730048D20E351A30B20|A87279E7F10CB60500000000002C0000FFFF0B0000|+4
|25442D2C4576946730048D20E452A30B20|A87279E7F10CB60500000000002C0000FFFF0B0000|+21
|25442D2C4576946730048D20E552A30B20|A87279E7F10CB60500000000002C0000FFFF0B0000|+37

telegrams from meter no. 2:
|25442D2C1376946730048D20BB31A70B20|154D79E7F1887F8E0100000000CA0900FFFF010000|+14
|25442D2C1376946730048D20BC31A70B20|154D79E7F1887F8E0100000000CA0900FFFF010000|+30
|25442D2C1376946730048D20BD32A70B20|154D79E7F1887F8E0100000000CA0900FFFF010000|+46
[...]

PR #1957, opened 2026-05-22
feat: add compact frame KAM,1c,04; KAM,30,04; KAM,30,0c; KAM,34,0c and KAM,39,04

### LINKED ISSUE #1484: kamheat: MULTICAL 403 (MC403) doesn't work - compressed and signature unknown
I have already used the util to decrypt data from my `multical21`, which does work as expected. Thank you, for this great solution!

Now I'm ready to get my Multical 403 to work, too.  

But I'm facing a wierd problem. At least, that's what I think?

Here's a telegram from the Multical 403:

```
3F442D2C9534477134048D20375A64FD224054B02BCC391E9E87C7E177960D24AA60C0D15E13B3DC23AB541A98BF5A40357105FAE61A865FACE2A89E1D845005 (64) C1 B
```

It does match the correct serial number, **however**, it's not able to extract the data.

I've provided the decryption key in the webpage, but it doesn't work.

I have the signature details (value?), as part of my details related to the decryption key (DEK), but I don't want to share it here. I'm willing to provide it directly, if that's needed.

What can I do?

```
Auto driver    : kamheat
Similar driver : unknown 00/00
Using driver   : kamheat 00/00
000   : 3f length (63 bytes)
001   : 44 dll-c (from meter SND_NR)
002   : 2d2c dll-mfct (KAM)
004   : 95344771 dll-id (71473495)
008   : 34 dll-version
009   : 04 dll-type (Heat meter)
010   : 8d ell-ci-field (ELL: Extended Link Layer II (8 Byte))
011   : 20 ell-cc (slow_resp sync)
012   : 36 ell-acc
013   : 3564fd22 sn (AES_CTR)
017   : c1c3 payload crc (calculated c1c3 OK)
019   : 79 tpl-ci-field (EN 13757-3 Application Layer with Compact frame (no tplh))
020   : 292e format signature (unknown)
022 CC: E1873FF1000015BA010060E90000F7AD02003119180C58000000230000009C0000001ACC000000000000 compressed and signature unknown

{
    "_":"telegram",
    "media":"heat",
    "meter":"kamheat",
    "name":"",
    "id":"71473495",
    "timestamp":"1970-01-01T00:00:00Z"
}


Using: wmbusmeters: 1.17.1-70-g41a2b48
41a2b4808f00367af7bbaac57e221539825babbb
```

I've looked at https://github.com/wmbusmeters/wmbusmeters/pull/147, but I'm not really sure if that's related. I've checked https://github.com/wmbusmeters/wmbusmeters/blob/master/src/driver_kamheat.cc too, but I'm not really sure where to start.

By giving the tool a wrong key, I get:

```
013   : af65fd22 sn (AES_CTR)
017   : f971 payload crc (calculated 61b4 ERROR)
019 CE: EECF9184A5D4A41D9D0DBA41A18F82CED315FB14DB8434C0FF23380B7C8010D02B93247DB60E748E4BFDA5642B failed decryption. Wrong key?
```

So I guess it's something with how the data is decoded, right?

I've also checked #1242, but I'm not sure if I'm in the same boat.

My hardware is approximately 2-3 meters away, from both the multical21 (which does work), and the multical403.

Data from both devices are picked up with the following solution:
https://github.com/SzczepanLeon/esphome-components/

Uzlopak:
fixes #1484

Used Telegrams from #48 #459 #181 #344


weetmuts:
Thanks! Note that the compact frame formats are mostly dependent on the configuration set for the meter, which is not the same thing as the mvt triade. I.e. each meter can have tons of different configuration settings and these settings result in different compact frames.

PR #1958, opened 2026-05-22
fix: only store compact formats when they have sane values

Uzlopak:
while scanning for compact formats, i realize that i get alot of 0xffff frames because format_bytes is empty.

Avoid this by adding sanity checks.


weetmuts:
Thanks!

PR #1959, opened 2026-05-22
feat: add compact frame for kampress, KAM,01,18

Uzlopak:
;)


jktjkt:
Instead of ~~writing this all by hand~~ using AI to extract them one-by-one from the existing test cases and storing them in the same XMQ file, why not extend the build system so that these DIF/VIF/... signatures are extracted from each provided test packet? That way you won't have to keep these two synced.

(also, Kamstrup pressure monitor is sadly EOLed)

Uzlopak:
This is a smart solution, but the wrong way to do it. The tests should validate the code and not be the source for the code. It would basically make it a circular validation.

jktjkt:
What I actually proposed is automating the gathering of packet signatures from the existing database of *long* packets which are currently in the test suite. You can still test this by, e.g.:
- having a long packet X in the test suite,
- then feeding the test suite with a short telegram (with no manually written hash) Y,
- checking that the decoding of this packet matches a known payload Z).

If this test passes, you've just tested this feature with no circular validation as far as I can tell.

You can then add more features on top of that, such as explicitly checking whether the hash -> long_sequence_of_difvif has an entry for a given hash, etc.

Uzlopak:
@jktjkt 

You write it yourself: "long packets in the test suite". Test data is test data.
And where do you store then the extrapolated result? in the xmq? Obviously not according to you. So there is another place to store them? Or do you want to "pollute" the xmq drivers with compact frame formats, even though you have no proof that the meters sends any compact frames.

Also we know only that kamstrup is sending the data in the same order without the difvif information. Is there even a guarantee, that the compact frame "must" contain the same data as the long frame? If so, can you point me to the place in the OMS or EN13757 spec, please? 

Also you need the long frames in the first place. If you dont have them, you are lost anyway.

To make the kamwater long frames pr possible I extracted 3000+ telegrams and +700 AES keys to find the long and compact telegrams to be able to add them.

jktjkt:
> And where do you store then the extrapolated result?

In the current master, the format of "full M-Bus frames" is specified as `/driver/compact_frame_formats/difvif`. What I propose is to *also* read them from `/driver/tests/test/telegram`. The proposed code would take a look at these telegrams, extract their DIF/VIF/... sequence, and then proceed as-if this sequence was given under the `compact_frame_formats` element. There's no explicit storage of format signatures anywhere else in the source code; they will be loaded dynamically when a driver's XMQ is loaded to memory.

Essentially, I propose to have code in wmbusmeters which will do the same thing as you did when you created this PR -- take a look at the existing full frames which are provided in the test suite, and to extract and remember their formats.

> Test data is test data.

A test for this newly proposed feature would involve adding a new compact frame (CI 0x79, 0x7b or 0x73) to the test suite. This newly added packet would check whether this proposed code works correctly. Obviously it is only necessary to add one or two checks for each packet format, not for all occurrences in all drivers.

> is there even a guarantee, that the compact frame "must" contain the same data as the long frame? If so, can you point me to the place in the OMS or EN13757 spec, please?

I believe that this is the actual definition of the compact format :). Take a look at [this preview of EN 13757-3:2018, Annex G](https://www.normsplash.com/Samples/DIN/122282135/DIN-EN-13757-3-2018-en-2.pdf). There's a lot of info in there (including a special "M-Bus-Format frame" which sends just the format, not the data), but I think it's clear that the order of the individual DIFVIF fields is expected to be stable. Here's the relevant part from the link above:

image
image
image


> Also you need the long frames in the first place. If you dont have them, you are lost anyway.

What I proposed was reusing the *known* long frames which exists in the testsuite today, and compute the format signature from these known-good frames. Doing that would make this PR (and #1957 and #1960) redundant because wmbusmeters would simply use the information which it already has to do this work at startup time.

Uzlopak:
I assessed your proposal and I think it is just covering half of the problem.

Or lets say, it is not to my liking.

You have tests, which you want utilize: Smart.

Somewhere in the wmbusmeters code and sometime in the wmbusmeters runtime the compact frames should be calculated and be remembered. 

You are not answering the question where and when this should happen.

I dont think that when the driver is loaded we should waste cycles to calculate the compact Format. 

So again, the test code is test code. Not production code. 

If I would need to write a concept for this, i would say, that we take care of this at compilation or pre release time. We write a script which extracts the compact_frame_formats from the existing long frames and adds them to the xmq. 

But hey, this is your proposal. Maybe you implement it the way you, and it gets accepted and everybody is happy.

PR #1963, opened 2026-05-24
feat: parse_TPL_7B

Uzlopak:
this PR adds parse_TPL_7B which is necessary for parsing some Apator telegrams.

I made som changes in the other parse_TPL_7x functions, to show that they are pretty much "the same", which small differences.


weetmuts:
Thanks!

PR #1964, opened 2026-05-24
feat: unify crc logic, add ability to create new crc16 variants

Uzlopak:
This PR unifies `crc16_EN13757()` and `crc16_CCITT()` logic, by using in both cases the MSB  method.

I personally would have preferred moving the logic of the new `uint16_t crc16(uint16_t poly, uint16_t init, uchar *data, size_t len)` into crc16.h and mark it as `__inline`. As such it would not be possible to call the `crc16()` in different places and would force the compiler to definetly inline it at compile time.

Anyhow: There are afaik two manufacturers, who use special CRC16 variants, IIRC it is kamstrup and apator.  So this PR enables developers to add their CRC16 variants to wmbusmeters-

### DIFF
--- a/src/crc16.cc
+++ b/src/crc16.cc
@@ -15,71 +15,64 @@
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#include"crc16.h"
-#include<cassert>
+#include "crc16.h"
 
-#define CRC16_EN_13757 0x3D65
+#include <cassert>
 
-static uint16_t crc16_EN13757_per_byte(uint16_t crc, uchar b)
+uint16_t crc16(uint16_t poly, uint16_t init, uchar *data, size_t len)
 {
-    unsigned char i;
-
-    for (i = 0; i < 8; i++) {
-
-        if (((crc & 0x8000) >> 8) ^ (b & 0x80)){
-            crc = (crc << 1)  ^ CRC16_EN_13757;
-        }else{
-            crc = (crc << 1);
-        }
-
-        b <<= 1;
-    }
-
-    return crc;
-}
-
-uint16_t crc16_EN13757(uchar *data, size_t len)
-{
-    uint16_t crc = 0x0000;
+    uint16_t crc = init;
 
     assert(len == 0 || data != NULL);
 
-    for (size_t i=0; i<len; ++i)
+    for (size_t i = 0; i < len; i++)
     {
-        crc = crc16_EN13757_per_byte(crc, data[i]);
+        uchar b = data[i];
+        for (int j = 0; j < 8; j++)
+        {
+            if (((crc & 0x8000) >> 8) ^ (b & 0x80))
+            {
+                crc = (crc << 1) ^ poly;
+            }
+            else
+            {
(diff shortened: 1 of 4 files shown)


jktjkt:
Where does Kamstrup use a special CRC implementation?

Uzlopak:
@jktjkt 

When the Kamstrup is not sending a compact frame with CI=0x79 but CI=0xED.

weetmuts:
Your optimizations on performance like inline is just totally unnecessary.

Especially since a previous change you made caused wmbusmeters to have 30% slower startup time, ie
because you scan and decode the entire driver database unnecessarily to find the compact frame signatures.
The loading of the compact frames from the driver should be done when the first telegram arrives, ie lazy.

I noticed this because fuzzing usually has 700 telegram tests per second, but it has now dropped to 400 per second.

I do not want to change the crc code. I will not change the crc code.
Changes to the crc code will be driven by new telegrams from users, where this is actually a need.
We will not implement new features without a user who actually complains.

Refactoring and moving the crc code to a separate source file is ok.

Uzlopak:
@weetmuts 
May I ask which commit made the perf regression? Just that I can learn from it?!

weetmuts:
Of course, it is a4a5a1db48b2da52281d01f5adeb2ebaa10c9b51

Uzlopak:
Ok, then I have found the right commit then. 

I guess the culprit is https://github.com/wmbusmeters/wmbusmeters/blob/a4a5a1db48b2da52281d01f5adeb2ebaa10c9b51/src/drivers.cc#L130

I think I added that because before the compact frames were globally applied. Then we decided to make them MVT specific. This made actually the loading of all compact frames prematurely obsolete. 

I think we can savely remove that.

Uzlopak:
@weetmuts 

Provided PR to fix #1969 the perf regression.

jktjkt:
> When the Kamstrup is not sending a compact frame with CI=0x79 but CI=0xED.

I didn't know about these (and I don't see such examples in either of `kamwater`, `kamheat`, `kampress`, `omnipower` or `em24` drivers). Got an example and/or more info about what kind of meter is that?

Uzlopak:
@jktjkt 

Unfortunately I also dont have any also

jktjkt:
I am confused now. You said that *some* Kamstrup meters send compact frames with CI=0xED, and that these use a non-standard CRC. If you don't have the details, *how do you know* that these are Kamstrup frames with non-standard CRC? Is it possible that you might have simply received a malformed frame over the air, for example?

In our environment (148 pcs of Multical 303 heat meters, 428 pcs of KWM2231 water meters, one EM24-compatible electricity meter in our buildings, *and* a ton of Zenner/Qundis/Maddalena heat cost indicators & water meters in the neighborhood), it's quite common to receive a packet that is very obviously malformed, even though it passes the initial CRC check. I know that these are malformed because nobody else has Kamstrup meters in the RF vicinity of our site, and we sometimes receive a packet with meter ID such as `KAS27284508` instead of the correct `KAW27284508`. A CRC will only do so much...

PR #1965, opened 2026-05-24
chore: restructuring, extract Alarm logic into /utils/alarm.cc and .h

Uzlopak:
This should mark the beginning of the restructuring by first extract a small struct Alarm from utils.cc and .h into ./utils/alarm.cc and .h. 

if this gets approved and merged, we can split the util.cc further. 

IMHO following aspects should be considered in this PR and some  aspects also in following PRs:

- is the name `utils`  for the folder ok? Should it be maybe get another other generic name like `helpers` or `lib`?
- is the extracted logic somewhat "self-encapsulated" and therefore meaningful extracted?
- is the name of the file appropriate?


weetmuts:
Thanks!

Issue #1951, opened 2026-05-20
possible improvements for XMQ drivers


jktjkt:
After reviewing #1944 I wanted to add support for the "modern" status fields of Kamstrup water meters myself. I came up with a "wishlist" of features which I think would make this a bit easier. I also quickly went through `driver_dyanmic.cc` and some existing `.xmq` files. Are there better docs about all the constructs which can be used in the XMQ sources?

Anyway, here's a list of suggestions that I would *love* to have available:

- Being able to specify the `loopkup` -> `BitToString` -> `map` -> `value` as a bit index/offset instead of a raw value (i.e., `bit = 8` instead of `value = 0x100`.
- Being able to reuse some parts of existing definitions in a similar manner to [YAML's anchors/references](https://blog.daemonl.com/2016/02/yaml.html). Kamstrup water meters currently have five possible status registers where large parts are "similar".
- Being able to bitshift the incoming value before the bit mapping is applied. This is for, e.g., kamwater's `time_dry`, `time_reversed`, etc where these times are always the same (i.e., `0` is "nope", `1` is 1-8 hours,  `2` is 9 to 24 hours, etc) -- as long as the extraction gets applied to a bitshifted value.

Also, my use case is feeding all of the received data to a [time series database](https://en.wikipedia.org/wiki/Time_series_database) which will be then used for data analysis. TSDBs work extremely well with, well, series of *numbers* (or even flags), which means that I will have to convert the `status` field into a bunch of sub-fields. Would you accept a CLI option which changes the JSON printing of bitfields from this:
```json
{
  "status": "ERROR_FLAGS_800"
}
```
to this:
```json
{
  "status": {
    "DRY": false,
    "OK": false,
    "TAMPER": false,
    "LOW_BATTERY": false,
    "ERROR_FLAGS_800": true,
  }
}
```
I.e., all fields would be listed explicitly, along with their current values. This will make the wmbusmeters -> TSDB extractor much easier because it will not have to keep track of all possible values which, e.g., a status field *might* have.

@weetmuts , are these features a good fot for wmbusmeters?

weetmuts:
All ideas are very good. I have pushed a fix for the simplest fix, you can now write bit=0 to test 0x1 bit=1 to test 0x2 etc.
Is it the standard to start with bit 0?

[...]

PR #1967, opened 2026-05-25
feat(xmq): introduce `tpl` token for `match_entire_frame`

Uzlopak:
The goal is to remove the `match_entire_payload` operator by `match_entire_frame` operator. This should make the drivers significantly more powerful on the long run. 

Migrating from `match_entire_payload` is easy, as you just replace the `match_entire_payload` with `match_entire_frame` and add the tpl token to the ixml instructions. See e.g. `apator08.xmq` as an example migration. 

Now I look at izar and apatorna1 driver, as they have still some magic in their code. If they can be migrated, we can assume that the driver logic can truely be encapsulated in the xmq.


weetmuts:
Thanks, but this is not a good idea. 

It creates an unholy alliance between the ixml grammar and the wmbus implementation which will make it hard to debug in the future. 

You want to be able to run the ixml parsing outside of wmbusmeters to debug and test it. With the tpl rule added inside wmbusmeters, this becomes impossible.

I suggest we change the name of the match_entier_frame and match_payload into this setting:

ixml_sees = field (this is the default)
ixml_sees = payload (jumps to the mfct specific part at the end of the telegram after 0xF or the tpl ci A)
ixml_sees = telegram (for really weird old meters)

Uzlopak:
how about `scope` 

Also i can move the tpl part directly into the driver itself.

Uzlopak:
Lets discuss this in #1951 please

PR #1968, opened 2026-05-25
feat: add `default_keys` to xmq.

Uzlopak:
Adds the opportunity to add `default_keys` to xmq, which then can represent factory keys or keys which are widely used in an area.


weetmuts:
Thanks!

PR #1969, opened 2026-05-25
fix: avoid loading built-in xmq drivers prematurely

Uzlopak:
@weetmuts

This should fix the perf regression.

### DIFF
--- a/scripts/measure_runs.sh
+++ b/scripts/measure_runs.sh
@@ -1,0 +1,155 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Detect UTF‑8 support
+if [[ "$(locale charmap 2>/dev/null)" == "UTF-8" ]] || [[ "$LANG" =~ UTF-8 ]]; then
+    USE_UNICODE=true
+    FULL_BLOCK='█'
+    EMPTY_BLOCK='░'
+else
+    USE_UNICODE=false
+    FULL_BLOCK='#'
+    EMPTY_BLOCK='-'
+fi
+
+# Determine number of runs
+if [[ $# -gt 0 && "$1" =~ ^[0-9]+$ ]]; then
+    RUNS="$1"
+    shift
+else
+    RUNS=10
+fi
+
+# Determine command
+if [[ $# -eq 0 ]]; then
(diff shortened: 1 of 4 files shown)


Uzlopak:
added a perf measuring script

master:

```sh
Measuring startup time of: ./build/wmbusmeters
Runs: 100

⠏ Progress: [████████████████████████████████████████] 100%  (100/100) 

Results in milliseconds:
  Runs    : 100
  Min     : 4 ms  (4,001,027 ns)
  Max     : 7 ms  (7,255,302 ns)
  Average : 4 ms  (5,152,673 ns)
  Std Dev : 0 ms  (641625 ns)
```

PR:

```sh
Measuring startup time of: ./build/wmbusmeters
Runs: 100

⠏ Progress: [████████████████████████████████████████] 100%  (100/100) 

Results in milliseconds:
  Runs    : 100
  Min     : 3 ms  (3,110,240 ns)
  Max     : 5 ms  (5,201,805 ns)
  Average : 3 ms  (3,806,525 ns)
  Std Dev : 0 ms  (494287 ns)
```

Seems about right. 

My daughter pulled the power cord, so i am not running with max performance :D

weetmuts:
Thanks! But when you look at this fix, does it not strike you that Claude did something truly silly by letting this code remain....
I mean, it is really really stupid code, that was not even necessary to keep there.... :-/

I suppose, this is why we see regressions in github:s ui behaviour, it gets slower by they day, because MS uses a lot of AI to develop their SaaS.

weetmuts:
Nice script by the way! @Uzlopak

Uzlopak:
@weetmuts 

Well It was "good" when i did the first iteration, where the compact format needed to be installed "global" to match the previous iteration. I actually tried to mitigate a potential perf regressions by "just" loading a subset of the drivers.  Because I did not want to introduce a functional regression, because I decided to do it differently than before. 

Then in second iteration, after we agreed that putting the compact frame format in the driver was an accepted approach,  we agreed  on only "activating" the compact frames per matching mvt. So we could ignore the global compact frames and load them when we hit the right MVT. And at that moment i forgot to remove this artifact.

PR #1970, opened 2026-05-26
`kamwater` improvements

jktjkt:
The main goal is to modernize the status field parsing (#1951). So far, I've used the `bit` instead of `value`. When more of the proposed changes in the core driver/extractor are implemented, I'll be happy to make use of them.


Uzlopak:
There is a bit option?!

jktjkt:
> There is a bit option?!

7b1d72c0

weetmuts:
Thanks @jktjkt !

PR #1971, opened 2026-05-26
fix: `--listmeters` should contain builtin drivers.

Uzlopak:


weetmuts:
Thanks! But the fix is too complicated. I do not understand it. 

I made a simpler fix instead: 706c1fdc70a07a77c48e18fc07013cbbfe9b17fc

PR #1972, opened 2026-05-28
feat: use HCI  reset for robust resetting iu891a

Uzlopak:
Imho not really testabe, because I think you need for atest to be in an environment with alot of wmbus devices so

I tested with: 
```
./build/wmbusmeters --resetafter=10s --verbose auto:t1
```


lucaszebrowsky-work:
Hi. We currently have a POC in a street with around 400+ wmbus devices. I will compile your branch and test if this fixes the problem, as we are facing the same issue.

lucaszebrowsky-work:
> Hi. We currently have a POC in a street with around 400+ wmbus devices. I will compile your branch and test if this fixes the problem, as we are facing the same issue.

Hello again. It has been three weeks now, and I’ve also used this PR on multiple other devices. Everything has been running without any problems so far.

The POC in the street I mentioned earlier, with 400+ wM-Bus devices, is also still running strong without any issues.

So from my side, this can be considered verified and ready to be merged

Uzlopak:
@weetmuts 

Wdyt :)?

PR #1979, opened 2026-06-03
chore: extract download logic, add unit test, fix bugs

Uzlopak:

PR #1980, opened 2026-06-03
fix: process env variables under macos properly

Uzlopak:
This PR fixes the regression of #1979 

Actually there was never the necessity to install curl for running the tests.

### DIFF
--- a/src/shell.cc
+++ b/src/shell.cc
@@ -87,7 +87,8 @@ void invokeShell(string program, vector<string> args, vector<string> envs)
         restoreSignalHandlers();
         close(0); // Close stdin
 #if (defined(__APPLE__) && defined(__MACH__)) || defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


Uzlopak:
@weetmuts 

Here the green test run on macos
https://github.com/Uzlopak/wmbusmeters/actions/runs/26888548622

weetmuts:
Super, thanks!

weetmuts:
@Uzlopak But now you have broken "make DEBUG=true"

Uzlopak:
@weetmuts 

What do you mean? I can build "make DEBUG=true" and run 
` ./tests/test_download_drivers.sh ./build_debug/wmbusmeters` without any issue.

weetmuts:
You are right, I only needed to remove the build_debug directory first. Thanks! @Uzlopak

PR #1981, opened 2026-06-03
feat: benchmark

Uzlopak:


Uzlopak:
This PR introduces benchmarks which can be called by doing`make benchmark xyz` 

Basically every benchmark is its own executable.

To write a benchmark you utilize benchmark.h

weetmuts:
Thanks! A lot of interesting code.

PR #1984, opened 2026-06-07
chore: fix wasm build, add working replacement for download

Uzlopak:
the wasm build was broken
I also used the chance to replace the download functionality with the browser equivalent.


weetmuts:
Great! Thanks!

PR #1985, opened 2026-06-08
KEM import: add support for EM24 electricity meter

jktjkt:
For some reason the KEM file which we got from the vendor has empty `<ConfigNo>` and `<TypeNo>` XML elements. Let's handle this properly.


weetmuts:
Thanks!

PR #1987, opened 2026-06-09
chore: remove util.h from translatebits.h

Uzlopak:
Removing util.h from translatebits.h to decouple the code further.


weetmuts:
Thanks!

PR #1990, opened 2026-06-10
chore: extract signal handling

Uzlopak:


weetmuts:
Thanks! I will just change the name of the new files to use underscore instead of minus.

PR #1994, opened 2026-06-14
Test framework

Uzlopak:
yes, yes, yes... it is very heavy change. 

I started by fixing the SLIP implementation and needed unit testing.  And I was annoyed by testinternals.cc being so convoluted.

So you write the unit tests in the header files, then include them in testinternals.cc and it will run as expected.

Looking for feedback.


weetmuts:
You have to split out the SLIP fix separately, in its own PR. 
The test framework idea is very nice!

Uzlopak:
#1996 is the slip protocol changes, this one contains only test framework

Uzlopak:
@weetmuts 

Should i resolve the merge conflict?

weetmuts:
I will take a look next week. A bit busy now!

weetmuts:
@Uzlopak Can you see if you can make this build again? I have made a lot of other changes regarding printMeter etc.

Uzlopak:
I will take care of it this weekend ;)

PR #1996, opened 2026-06-15
fix: slip processing according to spec

Uzlopak:
only slip fixes

### DIFF
--- a/src/utils/slip.cc
+++ b/src/utils/slip.cc
@@ -25,6 +25,8 @@
 
 bool slipAllEND(std::vector<uchar>& msg)
 {
+    // Returns true for an empty vector (vacuous truth). Callers that need to
+    // treat an empty buffer differently should check msg.size() first.
     for (size_t i = 0; i < msg.size(); ++i)
     {
         uchar c = msg[i];
@@ -44,6 +46,7 @@ ssize_t slipFrameSize(std::vector<uchar>& msg)
         if (c != SLIP_END) break;
     }
 
+    // Returns the encoded (wire) byte count — escape pairs count as two bytes.
     size_t from = i;
     for (; i < msg.size(); ++i)
     {
@@ -76,7 +79,7 @@ void addSlipFraming(std::vector<uchar>& from, std::vector<uchar> &to)
     to.push_back(SLIP_END);
 }
 
-void removeSlipFraming(std::vector<uchar>& from, size_t *frame_length, std::vector<uchar> &to)
+SlipFrameResult removeSlipFraming(std::vector<uchar>& from, size_t *frame_length, std::vector<uchar> &to)
 {
     *frame_length = 0;
     to.clear();
@@ -95,7 +98,23 @@ void removeSlipFraming(std::vector<uchar>& from, size_t *frame_length, std::vect
     for (; i < from.size(); ++i)
     {
         uchar c = from[i];
-        if (c == SLIP_END)
+        if (esc)
+        {
+            // esc is checked before SLIP_END so that ESC+SLIP_END (0xDB 0xC0) is
+            // caught as a framing error, not silently accepted as a frame terminator.
+            esc = false;
+            if (c == SLIP_ESC_END) to.push_back(SLIP_END);
+            else if (c == SLIP_ESC_ESC) to.push_back(SLIP_ESC);
+            else
+            {
+                // Invalid escape sequence (includes SLIP_END after ESC): corrupt frame.
+                // Report bytes consumed so the caller can resync by scanning forward.
+                to.clear();
+                *frame_length = i + 1;
+                return SlipFrameResult::CORRUPT;
+            }
+        }
+        else if (c == SLIP_END)
         {
             found_end = true;
             i++;
@@ -105,26 +124,21 @@ void removeSlipFraming(std::vector<uchar>& from, size_t *frame_length, std::vect
         {
             esc = true;
         }
-        else if (esc)
-        {
-            esc = false;
-            if (c == SLIP_ESC_END) to.push_back(SLIP_END);
-            else if (c == SLIP_ESC_ESC) to.push_back(SLIP_ESC);
-            else to.push_back(c); // This is an error......
-        }
         else
         {
             to.push_back(c);
         }
     }
 
-    if (found_end)
+    if (!found_end)
     {
-        *frame_length = i;
-    }
-    else
-    {
-        *frame_length = 0;
+        // No frame terminator yet, including the case where a trailing ESC byte has
+        // arrived but its second byte has not. Treat as partial — the caller should
+        // accumulate more data. frame_length stays 0.
         to.clear();
+        return SlipFrameResult::PARTIAL;
     }
+
+    *frame_length = i;
+    return SlipFrameResult::OK;
 }
--- a/src/utils/slip.h
+++ b/src/utils/slip.h
@@ -23,14 +23,32 @@
 #include<unistd.h>
 #include<vector>
 
+enum class SlipFrameResult
+{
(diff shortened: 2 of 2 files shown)


weetmuts:
So, what is the problem that is fixed?

Uzlopak:
The escaping is processed wrong, resulting in wrong terminated packages. We are basically dropping packages.

weetmuts:
No, what you are saying does not make sense. First the SLIP format is not perfectly specified for what to do when there are protocol errors in the stream. (https://www.rfc-editor.org/rfc/rfc1055.html) 

The current code does the right thing when there are no protocol errors. So the escaping is not processed wrong..

The current code does not drop packages if there is a protocol error. Not necessarily wrong, the RFC is fuzzy.

The proposed fix will drop packages with a protocol error. Not a bad idea, but.

Should the code drop packages if there are protocol errors? Perhaps yes, perhaps no.

However! I do not think it matters, because there should be no protocol errors of this kind. 
Unless there are bugs in the dongle.

Since the dongle sends the serial data immediately to the USB interface. There should be no noise in these millimeter long traces. And there are way more than a single bit toggle to get the kind of protocol error we would see here. A whole
byte must be lost.

So, have you actually experienced the problem or was the code analyzed by claude and determined to have a bug?

weetmuts:
Also, if there is noise introduced in the serial data, it could either be
ESC -lost byte- END thus the frame should end there as the current code does
or 
0xdc END thus the single bit was toggled and 0xdc became 0xdb which is ESC and the fframe should end there like what
the current code does.

ESC ESC_END  
or 
ESC ESC_ESC
where multiple bit-flips happened to convert 0xdc or 0xdd into 0xc0, ie this is more unlikely. 

The current code seems to optimize for the most reasonable assumption that the frame ended there...

Uzlopak:
I used following prompt:

> In /home/aras/Downloads/WM-Bus Gateway - Encrypted Package R16 (1)/WM-Bus Gateway - Package R16/WM-Bus Gateway HCI Example Code/HCI_ExampleCode/CPlusPlus we have SLIP Decoder and encoder. Consider it to be the spec! Compare the implementation from there with the implementation in src/utils/slip.cc
> 
> Is it implementd correctly here?

against master:


> Constants — identical ✅
> 
> Both use 0xC0 (END/BEGIN), 0xDB (ESC), 0xDC (ESC_END), 0xDD (ESC_ESC).
> 
> Encoder (addSlipFraming vs SlipEncoder::Encode) — correct ✅
> 
> The logic is identical:
> 
> Prepend 0xC0
> For each byte: escape END → ESC ESC_END, escape ESC → ESC ESC_ESC, otherwise pass through
> Append 0xC0
> Decoder (removeSlipFraming vs SlipDecoder::Decode) — one behavioral difference ⚠️
> 
> The core framing logic is correct, but there's one divergence in invalid escape sequence handling:
> 
> Scenario	Spec (SlipDecoder)	wmbusmeters
> ESC followed by unexpected byte	Abort frame, reset to Initial state	Pass the byte through as data (with // This is an error...... comment)
> Spec (SlipDecoder.cpp:111-113):
> 
> 
> else // error
> {
>     // abort frame reception -> return to initial state
>     _State = SlipDecoder::Initial;
> }
> wmbusmeters (slip.cc:113):
> 
> 
> else to.push_back(c); // This is an error......
> The spec says abort the frame on a bad escape; wmbusmeters silently swallows the byte as data.
> 
> Verdict: The implementation is functionally correct for all valid SLIP streams. The only divergence from the spec is in malformed-input error handling — the spec aborts the frame and resets on a bad ESC sequence, whereas wmbusmeters passes the offending byte through as data. For a gateway that sends well-formed SLIP, this will never trigger. If you want strict spec compliance, slip.cc:113 should clear to and return an error instead of pushing the byte.
> 

against this pr-branch in a fresh session:

> The implementation is correct for the happy path. Here's a detailed comparison:
> 
> Encoding (addSlipFraming vs SlipEncoder::Encode)
> They are identical in behavior:
> 
> Step	Spec	wmbusmeters
> Frame start	Append 0xC0	Append 0xC0
> 0xC0 in data	→ 0xDB 0xDC	→ 0xDB 0xDC
> 0xDB in data	→ 0xDB 0xDD	→ 0xDB 0xDD
> Other bytes	Pass through	Pass through
> Frame end	Append 0xC0	Append 0xC0
> Decoding (removeSlipFraming vs SlipDecoder::Decode)
> The core logic matches, but there are three design differences worth knowing:
> 
> 1. ESC + 0xC0 — both handle it as an error ✅
> The spec's EscState falls through to Initial when it sees 0xC0 (not 0xDC or 0xDD). wmbusmeters checks esc before checking for SLIP_END (slip.cc:101-105), so ESC 0xC0 correctly returns CORRUPT. Same outcome.
> 
> 2. Error recovery strategy — different approach, both valid
> Spec: On bad escape → return to Initial state, silently drop the frame, wait for the next 0xC0
> wmbusmeters: Returns CORRUPT with frame_length = bytes consumed, so the caller can resync explicitly
> The wmbusmeters approach is actually more explicit and safer for the caller.
> 
> 3. Leading 0xC0 handling — subtle difference, benign
> Spec: Has an Initial state that requires a 0xC0 to begin — if the very first byte is not 0xC0, it's silently discarded until one appears
> wmbusmeters (slip.cc:93-96): Skips any leading 0xC0 bytes, then immediately starts collecting data
> This means wmbusmeters is slightly more lenient — if you hand it a buffer with no leading 0xC0 at all, it would try to decode it. In practice this doesn't matter since valid SLIP frames always start with 0xC0.
> 
> 4. Empty frame (0xC0 0xC0) — minor edge case
> Spec: Fires no callback (because output is empty)
> wmbusmeters: Both 0xC0s get skipped as "leading", nothing is left → returns PARTIAL
> PARTIAL instead of OK with empty payload is a semantic difference, but neither implementation does anything useful with empty frames, so this is inconsequential.
> 
> Verdict
> The implementation is correct. Encoding is byte-for-byte identical to the spec. Decoding handles all the important cases: escape sequences, invalid escapes treated as errors, leading END bytes skipped, frame terminator detection. The differences are design choices (stateless function vs. stateful class) that don't affect correctness for real traffic.
> 

And PR-branch in the same claude session as the master with the following prompt:

> Now i switched the branch. What is your assessment now?


Result:

> This branch (slip, commit 28faaccd) fixes both issues I flagged. Here's the updated assessment:
> 
> The two bugs from before are now fixed:
> 
> 1. Invalid ESC handling — fixed ✅
> 
> Previously it silently passed the byte through. Now it clears the buffer, sets frame_length so the caller can resync, and returns SlipFrameResult::CORRUPT — matching the spec's "abort frame" behavior.
> 
> 2. ESC + 0xC0 mishandled — fixed ✅ (subtle, important)
> 
> This is the more interesting fix. The previous code checked c == SLIP_END before checking esc, so a 0xDB 0xC0 sequence would have been treated as a valid frame terminator (the 0xC0 would exit the frame). Now the esc flag is checked first (line 101), so ESC + 0xC0 falls through to the else branch → CORRUPT. That matches the spec's EscState which handles EscEnd/EscEsc only and aborts on anything else.
> 
> What's still different from the spec (intentional, not a bug):
> 
> The spec is a streaming state machine (processes one byte at a time, fires a callback per frame). wmbusmeters is buffer-based (processes a whole chunk, returns one frame with a length). This is an architectural choice, not a defect.
> The PARTIAL return value is a wmbusmeters addition (no equivalent in the spec, which just waits for more bytes). It's a clean API improvement for the buffer-based model.
> Verdict: correct. The implementation now faithfully implements the spec's byte-stuffing semantics, including proper error recovery on malformed escape sequences.

PR #1997, opened 2026-06-15
fix: dont "touch" authors.h if unchanged

Uzlopak:
Running `make` would always result in building despite no content was changed, because authors.h was generated autmatically. So we check before writing authors.h if it actually had a change.

### DIFF
--- a/scripts/generate_authors.sh
+++ b/scripts/generate_authors.sh
@@ -72,16 +72,14 @@ fi
     echo "#endif"
 } > "$NEWOUT"
 
-if [ ! -f $OUT ]
-then
-    # The new author files does not exist, write it.
-    cp $NEWOUT $OUT
+if [ ! -f "$OUT" ]; then
+    # First-time creation.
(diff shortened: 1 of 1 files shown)


weetmuts:
Hmm, the code is already working, you only change from diff to cmp and from cp to mv.
Why?

Uzlopak:
In master the script is always overwriting authors.h. So if you run `make` multiple times, you will always build wmbusmeters.

With these changes, we wont overwrite authors.h if the content is not changed. So if you run `make` multiple times, you will only build wmbusmeters once, after that it will report that the files are unchanged thus it wont build wmbusmeters again.

jktjkt:
Can you please share a full output log when this happens for you? I'm asking because it works correctly for me (and looking at the code, I agree with @weetmuts, it looks correct as-is). How can I reproduce what you're seeing?

Uzlopak:
STRANGE! I cant reproduce it either on my machine. my bad. sry

PR #2004, opened 2026-06-23
Fix M-Bus Type G invalid date parsing (0xffff)

bartman081523:
This pull request fixes the parsing of invalid M-Bus Type G dates (0xffff) by decoding them to null (for numeric/PointInTime fields) or empty strings (for string/Text fields) instead of overflowing/extrapolating to year 2127 or 2128. Test assertions have been updated accordingly.


weetmuts:
Thank you very much! But where is the test for 0xffff ? I was expecting a change in extractDate in dvparser.cc somewhere?

weetmuts:
Ah, the month > 12 test in dvparser.cc extractDate will return false for 0xffff and thus trigger the NaN and empty string generation.

Yes, that should work. :-)

PR #2005, opened 2026-06-23
test: add unit tests for Type F and Type I date/time decoding

bartman081523:
This PR adds internal unit tests for M-Bus Type F (4-byte CP32) and Type I (6-byte CP48) date/time structures to verify that hour, minute, and second values are extracted correctly.


weetmuts:
Are you an expert on the date/time formats? I read this proposed PR, but I could not see that it really makes a difference...
https://github.com/wmbusmeters/wmbusmeters/pull/1903

PR #2006, opened 2026-06-23
feat: stabilize hcae2 half-monthly history comparison fields

bartman081523:
The `hcae2` driver's half-monthly history fields were originally marked
`DEPRECATED` with `EXPERIMENTAL:` info prefixes because the on-wire layout was
not yet fully understood at the time they were added.

After validating the fields against real ELV/eQ-3 hcae2 device telegrams, the
semantics are now confirmed:

- **`half_month_history_start`** — date of the oldest snapshot in the rolling
  history. Its left edge is aligned to the start of the billing cycle containing
  that snapshot, so it may span more than exactly 15 days.
- **`delta_{N}_half_months_ago_hca`** — difference between two consecutive
  half-monthly snapshots (not an absolute reading). The 30 fields together cover
  ≥ 15 months. Summing forward from `delta_30` gives the cumulative reading at
  any half-month boundary.

### What this PR changes

Removes `attributes = DEPRECATED` and the `EXPERIMENTAL:` prefix from both
field definitions in `drivers/src/hcae2.xmq`, and updates the auto-generated
`src/generated_database.cc` accordingly.

No functional change — field names, quantities, and DIF/VIF matching are
identical. Only the metadata annotations are updated.

### Why remove DEPRECATED now?

Keeping `DEPRECATED` causes downstream consumers (Home Assistant, MQTT bridges)
to silently suppress these fields by default, even though the values are correct
and the layout is stable.

### Testing

All existing driver tests pass. The decoded values for `half_month_history_start`
and `delta_*_half_months_ago_hca` are verified by the existing hcae2 test
telegrams.


Uzlopak:
what is the reasoning?

weetmuts:
Sorry, this is just strange AI stuff. The following text is just wrong.

"Keeping DEPRECATED causes downstream consumers (Home Assistant, MQTT bridges)
to silently suppress these fields by default, even though the values are correct
and the layout is stable."

The DEPRECATED attribute is meant to be propagated into the json, but it isn't. So HA cannot silently
suppress, because nothing in the json reveals that it is deprecated....

Closing this for now.

Though, we should implement something like `"warnings":"Deprecated fields: x y z will be removed 2026-12-10."`
or for the whole driver: `"warnings":"Deprecated driver, upgrade to ..."`

PR #2008, opened 2026-06-23
Add --addrawfield option to include raw hex frame in JSON output

Uzlopak:
Introduces an opt-in _raw field containing the full raw telegram as a hex string, placed right after the _ key in the JSON output.

- Add add_raw_field flag to Configuration, MeterInfo, and Printer
- Parse --addrawfield CLI flag, addrawfield=true in config files
- Thread flag through Printer::print -> printMeter
- Emit "_raw":"<hex>" as the second JSON key when enabled
- Add tests verifying presence, absence, and value correctness
- Update README with documentation and examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ into human readable:
 or into csv:
 `MyTapWater;33225544;123.529;0;2024-03-03 19:36:45`
 
-or into json:
+or into json (with `--addrawfield` the raw hex frame is also included):
 ```json
 {
     "_":"telegram",
@@ -27,7 +27,8 @@ or into json:
     "id":"33225544",
     "max_flow_m3h":0,
     "total_m3":123.529,
-    "timestamp":"2024-03-03T18:37:00Z"
+    "timestamp":"2024-03-03T18:37:00Z",
+    "_raw":"2e44931578563412330389a2000000"
 }
 ```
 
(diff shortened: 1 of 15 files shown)


Uzlopak:
This would make my current project probably easier

weetmuts:
Yes, previously I have refused to pass on the raw (but decrypted) hex telegram.

The reason is that this has been requested by employees of companies who want to use wmbusmeters for collection but does not want to share how they decode the actual telegrams..... :-/  Not fair play....

Maybe wmbusmeters is big enough now so that it does not matter any more?
What do you think @Uzlopak ?

Uzlopak:
It is behind a flag. So they can still not provide the raw telegram. 

But it is so, that without the raw telegram i have to run two instances of wmbusmeters so that i have the telegrams and the json.

Uzlopak:
@weetmuts 
Can you consider merging this PR :). 

I personally dont like the name "addrawfield" but other than that, it should be good to go.

Uzlopak:
Well. I implemented in my solution it without this flag with 
logtelegrams=true... lol. Would be cooler if  this flag would be inside it. But well...

weetmuts:
Ok, I added this to the main branch: --addtelegramhex or in config addtelegramhex=true
this will add "hex": "...." the telegram hex (potentially decrypted if you had a key)
to the json. Can you test?

weetmuts:
I also added the --telegramdetails=always (or never or first) which will add details to the json about the quantity and unit and purpose of the field. This will be used to automatically generate mqtty discovery messages.

I also had to rewrite the entire output generation from naive string concatenation to using my libxmq library 
to compose the json. So adding the hex for the telegram is this easy now:

```
    if (getAddTelegramHex())
    {
        string hex = bin2hex(t->frame);
        xmqAddKeyValue(doc, telegram, "hex", hex.c_str(), NS_PARENT);
    }
```

weetmuts:
@Uzlopak By the way, thank you again so very very much for rewriting all the drivers into xmq format!

PR #2009, opened 2026-06-23
feat: implement --analyze=json output

bartman081523:
## What

`wmbusmeters --analyze=json:<driver>:<key>` (the JSON output mode of `--analyze`) returned a hard-coded placeholder:

```c
string renderAnalysisAsJson(vector<Explanation> &explanations)
{
    return "{ \"TODO\": true }\n";
}
```

This replaces the placeholder with the actual per-byte-range explanations as a JSON array:

```json
[
  { "pos": 0, "len": 1, "info": "18 length (24 bytes)", "ixml_parse": "", "kind": "PROTOCOL", "understanding": "FULL" },
  ...
]
```

String fields (`info`, `ixml_parse`) are JSON-escaped.

## How (TDD)

- `5c2ab5ae` adds a failing unit test (`test_analyze_json` in `src/testinternals.cc`) that asserts the array shape, the `pos`/`len`/`kind`/`understanding` fields, and that an embedded `"` in `info` is escaped. Against the stub it aborts on `assert(j.find("TODO") == string::npos)`.
- `823c84af` implements `renderAnalysisAsJson` + a small `escapeJson` helper and declares the function in `wmbus.h` so the test (and the existing `Telegram::analyzeParse` JSON path) can reach it.

```
./build/testinternals analyze_json   -> Test analyze_json
./test.sh build/wmbusmeters          -> All tests ok!
```

End-to-end, on a sample telegram:
```
$ wmbusmeters --analyze=json 1844AE4C...2B0000
Auto driver    : iperl
...
[
  { "pos": 0, "len": 1, "info": "18 length (24 bytes)", "ixml_parse": "", "kind": "PROTOCOL", "understanding": "FULL" },
  ...
]
```

## Scope

This is split out from #2007, which bundled several unrelated changes. `renderAnalysisAsJson` is the only piece not already covered elsewhere. The CP32/CP48/Type-G date-time work is covered by #1903.

### DIFF
--- a/src/testinternals.cc
+++ b/src/testinternals.cc
@@ -87,6 +87,7 @@ bool verbose_ = false;
     X(formulas_stringinterpolation)             \
     X(formulas_rounding)                        \
     X(formulas_extended_ops)                    \
+    X(analyze_json)
 
 #define X(t) void test_##t();
 LIST_OF_TESTS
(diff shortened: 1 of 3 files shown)


Uzlopak:
Imho should not be implemented in wmbus.cc.

weetmuts:
You are right @Uzlopak , also we already have two escapeJsonString already, which is one too many! Now @bartman081523 wants to add a third. ;-)

We should use the libxmq library to generate json, i.e. compose a DOM tree and then use the print function instead of just concatenating the text. However I need to add the necessary functions to libxmq to composing the DOM, so I am doing that right now. 

libxmq is also going to be used to generate the json for the telegrams eventually. I just have not gotten there yet.

weetmuts:
I pushed a fix. Please test.

PR #2014, opened 2026-06-26
parser + measurements output

veronika-ihnatova-asg:


weetmuts:
Can you explain what you want to achieve with this PR?

PR #2019, opened 2026-06-30
kadenwater: detect VIP-coded Kaden water meters + decode the water-leak alarm

Hymbajs:
Two related improvements to the `kadenwater` driver (Kaden/Ecomess type 07 water meters).

## 1. Detect Kaden water meters sold under the VIP manufacturer code

The same Ecomess type 07 water meter ships both under the **KDN** code and under the Kaden **VIP** code (`0x5930`), version `0x01`, type `0x07`, with an **identical data layout**. Only `KDN,01,07` was detected, so VIP-coded units fell through to the unknown driver — no readings, no status. Adds `mvt = VIP,01,07`.

Verified against **67 live VIP water meters**: all now autodetect as `kadenwater` (total 21–338 m³, status OK), no collision with the `kaden` HCA driver (`VIP,1E,08`).

## 2. Decode the water-leak alarm in the TPL status byte

The Kaden water meters signal a permanent water-leak alarm in the **unencrypted TPL status byte** (per the Kaden wM-Bus OMS spec, same encoding as the Picoflux AiR):

- bits 0+1 (value 3) → leak alarm — already decoded as `ALARM` by the standard OMS bits
- **bit 5 (0x20)** → during the leak the flow stayed **below 100 l/h**
- **bit 6 (0x40)** → during the leak the flow **exceeded 100 l/h** (both may be set)

Previously bits 5/6 surfaced as `UNKNOWN_xx`. Adds an `mfct_tpl_status_bits` block naming them `LEAK_FLOW_LOW` / `LEAK_FLOW_HIGH`.

Verified: healthy meters stay `OK`; status byte `0x63` → `"ALARM LEAK_FLOW_HIGH LEAK_FLOW_LOW"`.

## Verification
- `testinternals` passes; `make pretty` clean (lint_xmq)
- tests added for the VIP variant and for the leak path

The matching leak-status mapping for the related `picoflux` driver is in #2022.


Hymbajs:
Hi @weetmuts, gentle ping 🙂 CI is green and the PR is conflict-free — whenever you find a moment for a review, it would be much appreciated. We have a fleet of these VIP-coded Kaden water meters currently not decoding, so this one matters most to us. Thanks!

weetmuts:
Thanks!

PR #2020, opened 2026-06-30
hydroclimav2: surface the STS status word instead of discarding it

Hymbajs:
## What

The `hydroclimav2` IXML grammar parsed the BMeters **STS status word** (16-bit, in the `0F` manufacturer block) but discarded it — only the TPL status byte reached the `status` field. Device faults carried in the STS word were therefore invisible.

This maps `StatusWord` to error flags (`02FD17`) and lets the `status` field decode it with a raw `BitToString` lookup (default `OK`), mirroring the existing `picoflux`/`kadenwater` pattern. Non-zero STS now surfaces as `ERROR_FLAGS_xxxx` so faults are **visible** and can be named once their meaning is known.

The STS bit meanings are **not documented by BMeters** (verified against the Hydroclima 2 ITN spec and the BMeters RF-OMS 2.9 telegram spec — both only show STS as an example value). Rather than guess, the raw value is surfaced, with a `Help needed` comment, exactly like `picoflux`.

## Verification

On real fleet telegrams:
- 49/52 healthy devices stay `OK` (STS = `0x0000`)
- fault devices now expose their status, e.g. `93001021` → `"ERROR_FLAGS_100 UNKNOWN_40"` (was just `"UNKNOWN_40"`); the v0.53 example `68036198` reveals `"ERROR_FLAGS_4300"` that was previously hidden

`testinternals` passes; embedded tests updated; `make pretty` clean (lint_xmq).

## Files
- `drivers/src/hydroclimav2.xmq` — `@DV_status` (02FD17) on StatusWord + status field match/lookup + updated tests
- `src/generated_database.cc`, `tests/generated_tests.xmq` — regenerated


Hymbajs:
Hi @weetmuts, gentle ping — CI green, no conflicts, ready for review whenever you have time. Thanks!

weetmuts:
Thanks!

PR #2022, opened 2026-06-30
picoflux: decode the water-leak alarm in the TPL status byte

Hymbajs:
## What

Ecomess Picoflux AiR signals a **permanent water-leak alarm in the unencrypted TPL status byte** (per the Ecomess/Kaden wM-Bus OMS spec, shared by the Kaden water meters):

- bits 0+1 (value 3) → leak alarm — already decoded as `ALARM` by the standard OMS status bits
- **bit 5 (0x20)** → during the leak the flow stayed **below 100 l/h**
- **bit 6 (0x40)** → during the leak the flow **exceeded 100 l/h** (bits 5+6 may both be set)

Previously bits 5/6 were unmapped and surfaced as `UNKNOWN_xx`. This adds an `mfct_tpl_status_bits` block naming them `LEAK_FLOW_LOW` / `LEAK_FLOW_HIGH`, so leak severity is visible. The separate `02FD17` error register is reserved by the spec (`currently 0x00, future per OMS`) and is left as the raw `ERROR_FLAGS` fallback.

## Verification

- healthy meters stay `OK` (status byte `0x00`)
- a telegram with status byte `0x63` decodes as `"ALARM LEAK_FLOW_HIGH LEAK_FLOW_LOW"`; `0x23` → `"ALARM LEAK_FLOW_LOW"`; `0x43` → `"ALARM LEAK_FLOW_HIGH"`
- added a leak test (real 4.492 m³ telegram with the status byte set to `0x63`); `testinternals` passes; `make pretty` clean (lint_xmq)

## Files
- `drivers/src/picoflux.xmq` — `mfct_tpl_status_bits` + doc comment + leak test
- `src/generated_database.cc`, `tests/generated_tests.xmq` — regenerated

The same encoding applies to the `kadenwater` driver (Kaden KDN/VIP water meters) — happy to follow up with the same mapping there.


Hymbajs:
Hi @weetmuts, gentle ping — CI green, no conflicts, ready for review whenever you have time. Thanks!

weetmuts:
Thanks!

PR #2026, opened 2026-07-08
Update emerlin868.xmq

TimurSaitov:
Added newer version 12.
Added corresponding test.
In my dataset the "target_m3" resolves to "null" and  "target_date" resolves to "2000-00-00", probably indicating that initial meter setup were never properly initiated.

### DIFF
--- a/drivers/src/emerlin868.xmq
+++ b/drivers/src/emerlin868.xmq
@@ -6,6 +6,7 @@ driver {
     default_fields = name,id,total_m3,target_m3,timestamp
     detect {
         mvt = ELR,11,37
+        mvt = ELR,12,37
     }
     fields {
         field {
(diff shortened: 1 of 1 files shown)


weetmuts:
Thanks!

weetmuts:
The default value is the maximum negative binary value for target_m3.....
we can add null_value=-2147483.648
to the target field... this will detect this max negative value and replace it with null.
However the other target_m3 is 0 even thought the target_date is 2000-00-00

Oh, well, it is a temporary problem.

TimurSaitov:
Glad to help with this awesome project!
My plan is to wait for >1 year, to see if this field actually reports previous year total consumption. At the moment the meter is brand new, so no previous-year-data is available.

PR #2028, opened 2026-07-11
kamheat: recognize the combined MC303 heating & cooling meter

jktjkt:
...and extract the monthly cooling energy record. This fixes #1674 as reported by @ernstjerome.

@weetmuts, I need maintainer's help on the following two points:
- running `cd drivers; make install` produces a rather big diff in `../src/generated_database.cc` and `../tests/generated_tests.xmq`
- running `make test` succeeds (as in "the return code is 0") even though the run reports the following failures:

```
$ PATH=~/work/prog/wmbusmeters/build:$PATH make test

[...]

Testing src/kamheat.xmq
18a19
>   "target_energy_backward_kwh": 0,
Testing src/kampress.xmq

[...]

Testing src/qsmoke.xmq
5d4
<   "duration_removed_h": 0.004722,
11a11
>   "some_sort_of_duration_h": 0.004722,

[...]
```

PR #2030, opened 2026-07-17
XMQ driver order and a new `kamheat` compact telegram hash

jktjkt:
1) Fix two more missing place where the driver order was locale-dependent (see https://github.com/wmbusmeters/wmbusmeters/pull/2028#issuecomment-4995018741)

2) Add a compact packet hash for `kamheat`, showing that the DB regen now works for me, yay


jktjkt:
@weetmuts , I need your help with the tests. The new order which I implemented here puts `ultrimis` before `ultrimisv2`, which looks like a good alphabetical order to me.

However, the tests fail. This packet decoding in the test suite:

 https://github.com/wmbusmeters/wmbusmeters/blob/7daef4a8e01d1b8902e241101f781dc5c358a5fa/simulations/simulation_t1.txt#L207 contains `"meter": "ultrimis"` as the expected JSON, and the expected log also lists `ultrimis`:

https://github.com/wmbusmeters/wmbusmeters/blob/7daef4a8e01d1b8902e241101f781dc5c358a5fa/tests/test_listen_to_all.sh#L227


...but when I (or the CI) run that test, the *log output* says `driver: ultrimisv2`, yet the JSON output says `ultrimis`. How come that these outputs are different? How to fix that?

weetmuts:
Many thanks!

jktjkt:
After I pushed this, I realized that this breaks the test suite -- see my comment above. I do not understand the relation between the `meter` key in the JSON output and the `driver` field in the text output, and these two now diverge for one test -- which is why the CI started failing now. Sorry.

weetmuts:
So, the meter=kamheat should really be renamed to driver=kamheat.....
old field name from when I did not know better.

weetmuts:
But it seems to build and test properly? It is all green now. :-)

weetmuts:
The ultrimis.xmq should not have the mvt detect that ultrimisv2.xmq has.... that is the problem...

weetmuts:
I need to add something that prints a big warning if two drivers register the same mvt..... I thought I had implemented that, but it stopped working some time ago.

Issue #2003, opened 2026-06-21
Decoding error on flowiq2200


Omri-Kleynhans:
Explain the problem.....

https://wmbusmeters.org/analyze/3444372c595613543c168d200050048c20217a79f3a997d5013a0000150400000a0400004136000000000000055000065000808011:flowiq2200

Decoding of 044 what does this mean?
044 C?: 055000065000

Is this possible the acoustice noise value?

Thanks

weetmuts:
@jktjkt

jktjkt:
Yeah, `06 ff 1b` is the (proprietary) acoustic noise measurement. I'll send a PR.

Issue #1176, opened 2024-02-10
Register of unknown size


ssuukk:
### Type of request

Bug

### OS version

n/a

### wmbusmeters version

built from current master

### Your message goes here

Please open an issue at https://github.com/wmbusmeters/wmbusmeters/
and report this telegram: 5144B4091828912218078C00A87A3D0040052F2F066DA405130A32000C13135214000F05000000005A37002F1000E61000FB14007318002F1C00622000762400972800CE2C00783000EC3300002F2F2F2F2F

weetmuts:
Are you sure you are reporting the correct telegram? That telegram decodes fine, and is not an apator162 instead it is a iwmtx5 B-Meters water meter.

The error you are referring to are for apator162 water meters.

ssuukk:
I have more than 200 w bus devices in the air, this was just a random log
message I noticed, so this might be indeed the case you describe Ed.

pon., 12 lut 2024, 16:57 użytkownik Fredrik Öhrström <
***@***.***> napisał:

PR #2031, opened 2026-07-17
field changes for `kamwater`

jktjkt:
# RFC -- do not merge as-is

As reported in #1798 and #1176, the field naming or Kamstrup water meters is currently wrong. This is an RFC PR for @weetmuts to decide whether it's OK to make a change like this, or whether this needs a new `kamwater_v2` driver.

Also note that there's a *lot* broken unit tests, some of them in the legacy shell code. I'll only fix these (or maybe drop the old ones?) when it's decided whether my approach is acceptable.

Here's a list of the individual commits:

## 1. kamwater: rename some fields

The driver used inconsistent naming of various fields which are related to historic min/max value of flow, flow temperature, etc. Some of these were wrong. Kamstrup uses a pattern where a given "storage nr" always corresponds to the time interval over which the stats are calculated:

- `storage_nr=1` refers to a monthly data logger,
- `storage_nr=2` is a daily one,
- `storage_nr=3` is the yearly one.

This means that it's perfectly OK to see telegrams with "gaps" in the storage fields, e.g., a telegram with `storage_nr=2`, but no `storage_nr=1`.

To ensure correctness, this commit renames the following fields:

- `max_flow`, `min_flow`, `min_flow_temperature` and `max_flow_temperature` were always read from the "daily" register, so these fields now simply gain the `_last_day` suffix.

- In contrast to that, the `max_external_temperature` was read from the "monthly" register, so it gets the `_last_month` suffix.

- The old field `flow_temperature` was matching either the monthly or the daily logger, introducing an ambiguity. The daily logger was *also* being converted to a field `min_flow_temperature`. Start converting these fields separately, with either a `_last_month` or `_last_day` suffix, depending on what data the meter actually sends.

## 2. kamwater: move field definitions

...so that related values are placed close to each other. I.e., do not put flow temperatures between the current, min and max flows.

## 4. kamwater: recognize the proprietary acoustic noise measurement

The publicly available docs only clarify that it's encoded as 48bits of something. We don't know the internal structure of this field, so let's treat it as a blob.

Bug: https://github.com/wmbusmeters/wmbusmeters/issues/2003


weetmuts:
This looks good, and I think they all conform to being bug-fixes. So this should be ok to merge into the current driver.
What do you think Jan?

weetmuts:
I merged it and adjusted all the tests. A lot of places since the Multical 21 was the first meter I got, so its telegrams are part of the test-suite everywhere! :-D Please test!

jktjkt:
> I merged it and adjusted all the tests. A lot of places since the Multical 21 was the first meter I got, so its telegrams are part of the test-suite everywhere! :-D Please test!

You're my hero, I was putting this off due to all that redundancy. Thanks!

Issue #2036, opened 2026-07-28
amiplus with Gama350 and proper meter data time with reactive power


4SiB:
Explain the problem.....

https://wmbusmeters.org/analyze/be44ed145145583001027ae700b005a3c60e5d52ab6b2c2cfdd8368f408830f50f6d698041d61a76fd4e01988dde5d601f0e425682c43566156c932fc9d5661d01832707d90236c6ea2fddd6579a1533fe1b637082c2458939e96cc2cbd46410e6880a7990f7219a4b2c546b21eb2cddfda202aaf448ba01ab92e07ac68dd339d860b483fecdbd25851d938c002a3fe654875a0e6d8fa6692de3a66afc2dc3cff669d6a581ae7903f299e037ea81a7ae842fd4fe4ba4c9f5f01b60277a6861:amiplus:fba95687613b73604c320a16aa8892ea

I cannot use device_datetime but the device_date_time works for decode timestamp from electric meter Gama350.
I can decode two more 
current_rective_power_l_var
current_rective_power_c_var
from that telegram.

I use ESPHome with https://github.com/SzczepanLeon/esphome-components by CC1101 for decode electric meter GAMA350 in Poland.

```
[00:55:07.591][V][wmbusmeters:855]: (meters) trying field info device_date_time(Text)[7]...
[00:55:07.601][V][wmbusmeters:869]: (meters) using field info device_date_time(Text)[7] to extract 066D at offset 19

[00:55:07.826][V][wmbusmeters:855]: (meters) trying field info current_rective_power_l(Reactive_Power)[17]...
[00:55:07.837][V][wmbusmeters:869]: (meters) using field info current_rective_power_l(Reactive_Power)[17] to extract 0BFB14 at offset 140
[00:55:07.849][V][wmbusmeters:1735]: (meter) Any current_rective_power_l_var decoded var default var value 67 (scale 1)
[00:55:07.858][V][wmbusmeters:855]: (meters) trying field info current_rective_power_c(Reactive_Power)[18]...
[00:55:07.868][V][wmbusmeters:869]: (meters) using field info current_rective_power_c(Reactive_Power)[18] to extract 0BFB943C at offset 147
[00:55:07.879][V][wmbusmeters:1735]: (meter) Any current_rective_power_c_var decoded var default var value 187 (scale 1)
```

BR Marcin

weetmuts:
[...]

PR #2038, opened 2026-07-30
amiplus: add reactive power and reactive energy fields, detect Gama 350

Kustonium:
Ref #2036 — this implements what was asked for there.

Apator Otus 3 and Apator Gama 350 electricity meters report reactive power and
reactive energy in manufacturer specific records that the amiplus driver
currently ignores:

| difvifkey    | raw value      | added field                          |
|--------------|----------------|--------------------------------------|
| 0EFB8273     | 646163010000   | total_reactive_energy_inductive      |
| 0EFB82F33C   | 924805050000   | total_reactive_energy_capacitive     |
| 0BFB14       | 230300         | reactive_power_inductive             |
| 0BFB943C     | 180300         | reactive_power_capacitive            |

The VIFEs (FB 82 and FB 14) are Reserved in the standard, so `vif_scaling = Auto`
cannot work here: `vifScale()` falls through to `default: return -1`, and since the
value is divided by the scale, the result comes out unscaled *and negated*
(-1636164 instead of 1636.164). The fields therefore use `vif_scaling = None`
with an explicit `display_unit` and `force_scale`, the same shape iem3000 uses
for its manufacturer specific records.

Scaling is inferred from the other records in the very same telegram: this meter
reports active power in W (0B2B = 68 W) and active energy in Wh (0E03 = 6265020 Wh),
so the reactive records are var and varh. `force_scale = 0.001` converts them to
kvar/kvarh, consistent with `current_power_consumption_kw` and with abbb23/iem3000.

`default_fields` is deliberately left untouched — most amiplus meters do not send
these records, and adding them would only add null columns to the fields output.

Two tests from real telegrams are included, one per meter model; all existing
amiplus tests are unaffected. The Gama 350 telegram is the one posted in #2036 —
that meter sends only the reactive power records, not the reactive energy ones,
which is a useful check that the new fields stay silent when absent.

**Also in this PR: `mvt = EGM,01,02`.** The Gama 350 in #2036 identifies as EGM,
which no driver claims, so `--analyze` reports `Auto driver: not found!` and the
user has to name the driver by hand. With that line added it auto-detects as
amiplus. Happy to split this into its own commit or drop it if you would rather
keep the PR to the fields.

**Naming.** The pairs are the inductive and capacitive components, not import
and export. I first named them consumption/production because the only
difference between the two records of a pair is the BackwardFlow combinable
(3C), the way this driver already separates 0E03/0E833C and 0B2B/0BAB3C for
active energy. The data says otherwise:

- In the Otus 3 telegram *both* reactive power records are non-zero at the same
  instant (0BFB14 = 323, 0BFB943C = 318). Instantaneous reactive power is a
  single signed quantity, so an import/export pair cannot both be non-zero;
  two quadrant accumulators can.
- @4SiB, who owns one of these meters, computes net reactive power from the pair
  as L - C ("positive = inductive, negative = capacitive") and derives power
  factor and tan phi from it:
  https://forum.arturhome.pl/t/komponent-wm-bus-do-esphome-wersja-5-x-watek-ogolny/15057/588
- The out-of-tree patch that first described these registers names them _l / _c.

I spelled the words out instead of using _l / _c, which means nothing to a reader
outside the trade. Say the word if you prefer the short form.

Update: the `current_` prefix is dropped as requested, so the fields are
`reactive_power_inductive` and `reactive_power_capacitive`. The active power
fields keep their names, since renaming them would break existing setups.

**Credit.** The DIF/VIF keys and the Otus 3 test telegram come from a patch by
Witold Olechowski in the SzczepanLeon/esphome-components fork (commit 37f9a5e,
2026-04-02, GPL-3.0), written against the old C++ amiplus driver that has since
been replaced by the XMQ one. The Gama 350 telegram and key are from @4SiB in #2036.

A full `make test` passes. `tests/test_list_envs.sh` uses amiplus as its example
driver, so its expected env and field lists are updated alongside (the field name
column is right aligned, so the longer names re-pad every line there).

The analysis and the patch were worked out together with Claude (AI), following
your AGENTS.md — hence the Co-Authored-By trailer. Every claim above was checked
by running it: the scaling, the two telegrams, the full test suite. If anything
here still reads as guesswork, say so and I will go back to the binary for it.


weetmuts:
Nice, but can you avoid the "current" in the fields? Since it also can mean the amperage (current) I really try to avoid it all cost.
If something is the instantaneous (current) measurement I usually just use the plain name. Typically power_w etc.

Are all amiplus compatible meters compliant and behaving the exact same way? Do these field, if they exist in the other manufacturers meters behave the same because of the amiplus specification?

weetmuts:
So the new fields can be reactive_power_inductive (no need to add current_ first)

4SiB:
> So the new fields can be reactive_power_inductive (no need to add current_ first)

instantaneous_value_
now_
peak_
current_ <- seams the best for me, even if have similarity to amperage.. but People have IQ, they understand.

weetmuts:
Or just no prefix at all for instantaneous values. That is the wmbusmeters standard. :-)

weetmuts:
Do you know anything about the amiplus standard itself? My other question.

Kustonium:
Prefix dropped — the fields are now reactive_power_inductive_kvar and reactive_power_capacitive_kvar. I left the active power fields alone: current_power_consumption_kw and current_power_production_kw predate this PR and renaming them would break existing setups, but say the word and I will do it separately.

On the amiplus standard: no, I do not have one, and I would not claim these fields behave the same across manufacturers.

What I can say is limited to Apator. The four records use VIFEs that are Reserved in EN 13757, so they are manufacturer specific by definition — there is no standard text that would make another vendor encode them the same way. My evidence is two real telegrams, an Otus 3 and a Gama 350, plus two independent descriptions of the same registers: the out-of-tree patch by Witold Olechowski, which arrived at the same DIF/VIF keys from a different meter, and @4SiB, who reads the pair off his own meter and gets a sensible power factor out of L - C. That is enough for me to trust the mapping on Apator hardware, and not enough to say anything about anyone else's.

As far as I can tell "amiplus" here is a driver that grew to cover Apator meters and their rebrands, not a published specification that other manufacturers implement. If you know otherwise I would rather hear it than guess.

If I am wrong the blast radius is small: the fields only appear when a meter actually sends those difvifkeys, and default_fields is untouched, so meters that do not send them are unaffected.

Still open from my side: do you want mvt = EGM,01,02 in this PR, split into its own, or dropped?

weetmuts:
Thanks!

Issue #2042, opened 2026-08-01
Sanxing S34U28 (mfct=KPL) electricity meter sends non-2F2F decrypt-verification bytes


bryszard:
# Sanxing S34U28 (mfct=KPL) electricity meter sends non-2F2F decrypt-verification bytes

## Summary

A Sanxing S34U28 three-phase electricity meter (Tauron/Poland AMI deployment) uses TPL
security mode 5 (AES_CBC_IV), but the first two bytes of the decrypted content are **not**
the OMS `2F2F` verification marker. With the correct per-meter key (obtained from Tauron's
own portal), decryption is genuinely successful — the DV record stream that follows is
byte-for-byte identical to the existing `amiplus`/Apator meter family — but wmbusmeters
rejects the telegram because of the `2F2F` check, and permanently ignores the meter.

```
(wmbus) WARNING!! decrypted content failed check, did you use the correct decryption key?
Permanently ignoring telegrams from id: XXXXXXXX mfct: (KPL) KPLC Power, Kenya (0xae0c)
type: Electricity meter (0x02) ver: 0x01
```

## Device

- Model: Sanxing S34U28, 3-phase electricity meter
- Deployment: Tauron (Poland) AMI rollout, wM-Bus T1 mode
- Declared wM-Bus manufacturer/version/type: `KPL`, `0x01`, `0x02`

Worth noting: `KPL` is **not** Sanxing's registered manufacturer code. Per the DLMS FLAG ID
registry (and already present in this repo's own `src/manufacturers.h`), `KPL` is registered
to `KPLC Power, Kenya`; Sanxing's own registered code is `AUX`. I don't know why this
particular device/firmware declares `KPL` — flagging the mismatch as a fact, not a claim
about the cause.

## Evidence

I have two real captures, several hours apart, using the correct key for this meter
(obtained from Tauron's portal). IDs below are anonymized (`78563412` / `12345678`
replacing the real 4 bytes consistently everywhere they appear, including inside the
decrypted fabrication-no field, so the two stay internally consistent). Meter name is
also a placeholder. AES key is intentionally **not** included.

### Frame A — captured 2026-07-28 13:02:00 (local)

```
[...]

PR #2043, opened 2026-08-02
fix: support Sanxing S34U28 (KPL mfct code) amiplus-compatible electricity meter

bryszard:
Ref #2042 — full evidence (real captures, timestamp cross-checks, why this isn't a wrong
key) is there.

## Update (rebuilt per review)

@weetmuts — implemented as suggested: a standalone `sanxing` driver plus a
`transform_payload` hook, same shape as the existing `diehl_prios` case, instead of the
mfct/version/type-gated exception inside `amiplus.xmq`/`wmbus.cc` from the first version
of this PR. `amiplus.xmq` is now untouched relative to `master`.

One naming deviation from the suggestion: the hook is `transform_payload=sanxing_609B`
and `setSanxing609BDecode(bool)`, not `buggy_sanxing_609B` / `setBuggySanxing609BDecode`.
Dropped "buggy" to keep the identifier to the dry, checkable fact (device emits a fixed
`0x609B` marker instead of `2F2F`) rather than a judgment call — happy to rename if you'd
rather match your suggestion literally.

## What

Sanxing S34U28 three-phase electricity meters (Tauron/Poland AMI deployment, wM-Bus
manufacturer code `KPL`, version `0x01`, type `0x02`) use TPL security mode 5
(AES_CBC_IV), but don't put the OMS `2F2F` verification marker first. With the correct
key, decryption is genuinely successful and the DV record stream that follows is
byte-for-byte identical to the existing `amiplus`/Apator meter family — no new fields
needed — but wmbusmeters rejects the telegram on the `2F2F` check and permanently
ignores the meter.

Three independent captures (different access counters, therefore completely different
IVs and ciphertext each time — see #2042) all decrypt to the same two leading bytes:
`0x60 0x9B`. That's a fixed marker this device emits, not arbitrary noise, so the fix
requires that exact value rather than accepting anything non-`2F2F`.

## Shape of the fix

- `drivers/src/sanxing.xmq` — new driver, copy of the `amiplus` field/detect layout
  (`mvt = KPL,01,02`), with `transform_payload = sanxing_609B` added.
- `src/driver_dynamic.cc` — when a driver declares `transform_payload=sanxing_609B`,
  calls `setSanxing609BDecode(true)` on construction (mirrors the existing
  `diehl_prios` branch right above it).
- `src/meters_common_implementation.h` — `setSanxing609BDecode(bool)` /
  `sanxing_609b_decode_` flag, same pattern as `diehl_prios_decode_`.
- `src/meters.cc` — in `handleTelegram`, if the flag is set, sets
  `Telegram::permit_sanxing_609b_quirk = true` before parsing.
- `src/wmbus.cc` (`Telegram::potentiallyDecrypt`) — the `0x609B`-marker check now only
  applies `if (permit_sanxing_609b_quirk)`, i.e. only for telegrams being decoded by a
  driver that opted in. It still requires the exact `0x609B` value and the trailing
  `2F2F` padding at the end of the frame, for the same false-accept-rate reason as
  before: a wrong key gives pseudorandom leading bytes, so accepting any non-`2F2F`
  value would pass whenever those two random bytes and the two trailing bytes both
  happened to land on `2F2F` — roughly 1-in-65536 per telegram. Requiring `0x609B`
  first drops that to 1-in-~4.3 billion, on top of now also requiring the driver
  assignment itself to opt in.

## Precedent for this shape of change

`src/manufacturer_specificities.cc` already does the same kind of thing for the Diehl
family (DME/EWT/HYD/SAP/SPL): `DiehlLfsrCheckMethod::HEADER_1_BYTE` verifies successful
decryption against a fixed non-OMS byte value instead of the standard marker, and
`CHECKSUM_AND_0XEF` verifies it via a checksum instead of a marker at all. Both are
called from `Telegram::preProcess()`, the same hook this fix uses (via
`potentiallyDecrypt`).

## Why `KPL`, not Sanxing's own `AUX` code

Also worth noting (not something this PR needs to resolve, just documented for whoever
looks at this later): `KPL` is not Sanxing's registered wM-Bus manufacturer code —
that's `AUX` (`Ningbo Sanxing Smart Electric Co., Ltd.`, per the DLMS FLAG ID registry,
and already present in this repo's own `manufacturers.h`). `KPL` is registered to `KPLC
Power, Kenya`. Flagging the mismatch as a fact; #2042 covers what I checked and didn't
find an explanation for.

## Test

`drivers/src/sanxing.xmq` has one test. The telegram is **synthetic**: real DV records
from one of the captures in #2042, re-encrypted under a fabricated throwaway key with
the meter id changed — no real key or live household data is committed. Full
construction and reasoning is in the test's `comment` field.

Verified in both directions:
- Fix applied → `cd drivers && make testonly DRIVER=sanxing` passes clean, no diff.
- Fix fully reverted (`src/wmbus.cc`/`src/meters.cc`/`src/meters_common_implementation.h`/
  `src/driver_dynamic.cc` back to identical to `master`) → the same test fails (telegram
  rejected, no JSON emitted) — confirms the test actually exercises this code path
  rather than passing regardless.

Also re-tested against the real meter/key on my own deployment (custom HA add-on build
from this branch) after the rebuild — decrypts and reports correctly with `type:
sanxing` configured.

`make pretty`, `drivers && make install`, and a full `make test` were run — all pass;
`git diff --quiet drivers` is clean (no formatting drift).

## Note on how this was put together

The investigation and this patch (including the rebuild after your review) were worked
out together with Claude (AI), following this repo's AGENTS.md — hence the
Co-Authored-By trailer on the commits. Happy to go into more detail on any part of the
reasoning in #2042 or here.


weetmuts:
Hmm, clearly they have a very buggy implementation.... the 2F2F is required for verifying the integrity of a decrypted AES messages..... very strange....

But implement it like this instead, create a new driver, which is a copy of amiplus add C++ support for:
transform_payload=buggy_sanxing_609B 
to the driver. (We already support the izar meters with transform_payload=diehl_prios for their very specific protocol....)
add  void setBuggySanxing609BDecode(bool v).
Then only if this boolean is set, then the check will also permit 609B after it has tested for 2F2F.

bryszard:
@weetmuts Thank you for the constructive feedback. I'll rebuild the implementation.

bryszard:
Ok, I've rebuilt it, checked it against my local setup (still works) and changed the PR description. 

I'm going for holiday on Monday for 2 weeks. If any more changes will be needed, I can address that in the last week of August.

weetmuts:
Many thanks. I change the name of the transform to `transform_payload = buggy_sanxing_609B`
and I added some more fields that were present in the telegram:

https://wmbusmeters.org/analyze/5e440cae4433221101027a01005005c86fBe60c581f909a1c76c4adB96896111c7629254158644f874083d105c737f6f56812ce5e4f9B60cdfBe01cdB2a76e38fa17d0ef61f683fe189a57Bf35d23c7B8631f22fB5dd64e41acac3df2476a0:auto:38852fe01b9427ffcdf549b50190721a

Please test!

PR #2044, opened 2026-08-03
q400: add detect for Axioma Qalcosonic W1 dll-version 0x20, decode error code + error-free time

Hymbajs:
## Summary
- Add `mvt = AXI,20,07` to `q400`'s `detect{}`. Real-world telegrams from an Axioma Qalcosonic W1 (dll-version 0x20) were not auto-detected since only versions `0x01` and `0x10` were listed.
- Add two fields documented in Axioma's official "QALCOSONIC W1 Wireless MBus specification" (v1.8, 2020-12-28) that were previously left undecoded:
  - `error_code_counter` — instantaneous error code (DIF `34` VIF `FD 17`)
  - `error_free_time_h` — time since the meter last recorded an error (DIF `04` VIF `24`)
- Add a regression test (`AxiomaW1v20`) built from a real device telegram, and update the existing `MM` test's expected JSON now that the two new fields are decoded.

Verified the decoded status/volume/temperature values against an independent LoRaWAN payload decoder for the same physical device (status "Dry / empty spool", volume 0 m³) — decoding was already correct, only auto-detection and two documented-but-unparsed fields needed fixing.

## Test plan
- [x] `cd drivers && make pretty && git diff --quiet drivers` (lint)
- [x] `cd drivers && make testonly DRIVER=q400` — all 5 driver tests pass
- [x] `cd drivers && make test` — full driver test suite passes, no other drivers affected
- [x] `./configure && make` — builds cleanly
- [x] Verified `--format=json stdin:hex ... auto ...` now auto-detects `q400` for a real dll-version 0x20 telegram (previously failed with "meter detection could not find driver")


weetmuts:
Many thanks!

PR #2045, opened 2026-08-03
ultraheat: add target_kwh and actuality_duration_s

cgommel:
Real LUG 04/04 meters send a storage 1 energy record (`4C 06`) and an actuality duration (`02 74`) that the driver currently leaves unclaimed — `--analyze` marks both red. Both are covered by existing library entries, so this adds two `use` lines to the library block. `default_fields` is left untouched so the `--format=fields` column order does not change for existing users.

Existing test 1 also carries these records, so its expected JSON gains `target_kwh` and `actuality_duration_s`; the same applies to the ultraheat entry in `simulations/simulation_mbus.txt`. Test 2 is unaffected.

A third test is added from a real meter. The payload was decrypted with `--logtelegrams` and the frame uses the usual `_` split, so no key is needed; the meter id in the header was anonymised.

Still open: the record `8C 01 06` (storage 2, 51 kWh in this sample) remains unclaimed. I could not establish its meaning for this device — `istaheat` treats storage 2 as the monthly value, `pollucomf` treats storage 1 as the set date, and the observed values fit neither cleanly. I would rather leave it unmapped than guess at a field name. If anyone has the Landis+Gyr documentation, happy to follow up.


weetmuts:
Many thanks!

Issue #200, opened 2020-12-10
Adding support for Techem Smoke detector 2


Kosygor:
Hello All

I have this Techem Smoke detector 2 or (Funk-rauchwarnmelder 2) and I am trying to make use of it in wmbusmeters.
I have read issue #165 and [https://github.com/weetmuts/wmbusmeters/blob/master/HowToAddaNewMeter.txt](How to add manual) but I am willing to do testing but need some guidance.

I have manage to receive that kind of message:
```Received telegram from: 91633569
          manufacturer: (TCH) Techem Service (0x5068)
                  type: Unknown (0xf0)
                   ver: 0x76
                device: rtlwmbus[00000001]
                  rssi: 76 dBm
                driver: unknown!
telegram=||294468506935639176F0A0009F2789290060892900000501D7311AF93E1BF93E008DC3009ED4000FE500|+282
```

Ok so I have listened for few messages and copy paste them to tht file like this: 
```
telegram=||294468506935639176F0A0009F2782290060822900000401D6311AF93E1BF93E008DC3009ED4000FE500|
telegram=||294468506935639176F0A0009F2782290060822900000401D6311AF93E1BF93E008DC3009ED4000FE500|
telegram=||294468506935639176F0A0009F2782290060822900000401D6311AF93E1BF93E008DC3009ED4000FE500|
```

from this file I got:
```
wmbusmeters --debug simulation_dymr.txt Wasser lansensm 91633569 NOKEY

(simulation) from file "294468506935639176F0A0009F2782290060822900000401D6311AF93E1BF93E008DC3009ED4000FE500"
(wmbus) parseDLL @0 42
(meter) Wasser: for me? 91633569
(meter) Wasser: meter detection did not match the selected driver lansensm! correct driver is: unknown!
(meter) please consider opening an issue at https://github.com/weetmuts/wmbusmeters/
(meter) to add support for this unknown mfct,media,version combination
(meter) Wasser: yes for me
(meter) Wasser lansensm handling telegram from 91633569
(meter) Wasser 91633569 "294468506935639176F0A0009F2782290060822900000401D6311AF93E1BF93E008DC3009ED4000FE500"
(wmbus) parseDLL @0 42
(telegram) DLL L=29 C=44 (from meter SND_NR) M=5068 (TCH) A=91633569 VER=76 TYPE=f0 (Unknown) (driver unknown!) DEV= RSSI=0
(wmbus) parseELL @10 32
[...]

PR #2050, opened 2026-08-05
fix: add DEFECT status support to tsd2 driver

docbobo:
I just noticed in my local testing, that one of the Techem Smoke Detectors I was receiving reported `STATUS_80`, which - according to https://github.com/wmbusmeters/wmbusmeters/issues/200 - should indicate `DEFECT` instead.


weetmuts:
Many thanks!

PR #2052, opened 2026-08-10
kadenwater: add ECM,01,07 detect (real water meter was undecodable)

Hymbajs:
## Summary
- A real-world Ecomess/Kaden water meter reporting under manufacturer code **ECM** (0x146d), version **01**, type **07** falls through autodetect as "unknown". `kadenwater.xmq` currently only detects `KDN,01,07` and `VIP,01,07`; `picoflux.xmq` detects `ECM,05,07` (different version).
- Verified the telegram's DIF/VIF layout matches `kadenwater`'s field set exactly (100% byte coverage: status, fabrication_no, total_m3, volume_b_m3, volume_flow_m3h, max_flow_m3h, actuality_duration_s, duration_b_s, meter_datetime) — `picoflux` leaves several of these unmapped.
- Fix: add `mvt = ECM,01,07` to `kadenwater.xmq`'s `detect{}` block, plus a regression test built from a real captured telegram (device id and fabrication_no anonymized; AES key is the manufacturer's public default key, same one already used by `picoflux`'s ECM tests).

## Test plan
- [x] `cd drivers && make pretty` — no formatting diff
- [x] `cd drivers && make install` — regenerated `src/generated_database.cc` / `tests/generated_tests.xmq`, diff limited to the new mvt entry + test block
- [x] `make` — builds clean
- [x] `cd drivers && make testonly DRIVER=kadenwater` — passes
- [x] `./test.sh build/wmbusmeters` — full suite passes (`All tests ok!`)
- [x] Confirmed autodetect now resolves `ECM,01,07` to `kadenwater` via `--format=json stdin:hex ... auto ...`


weetmuts:
Thanks!

PR #2054, opened 2026-08-11
fix: find compact profile base value regardless of its data field coding

cgommel:
A compact profile and the base value it refers to are separate data records. `findCompactProfileBaseValue()` only accepted a candidate whose DIF data field coding matched the coding of the profile elements. That rules out the normal case, because a reading needs more range than an increment does — [OMS Spec Vol.2](https://oms-group.org/wp-content/uploads/2024/10/OMS-Spec_Vol2_Primary_v501_01.pdf) Annex G, Table G.4 combines an 8 digit BCD base value with 4 digit BCD increments.

When no base value was found, no reconstruction took place and the raw increments were stored as readings. There is no warning and the numbers look plausible, which is what makes it easy to miss.

Fixing that exposed a second problem: reconstructed readings were re-encoded using the width of the *increment*. In the Annex G example that happens to fit (209 into 4 digit BCD), but a meter sending 1 byte increments on top of a 64 bit base value overflows, so the profile is dropped from its first element onwards — and where a second profile for the same quantity exists (import/export), that one silently takes over the keys. Reconstructed readings now carry the coding of the base value record, which is also what Annex G, Table G.7 shows for the converted single data points.

Base value lookup now also compares the combinable VIFEs, so that a telegram carrying both a forward and a backward profile for the same quantity picks the correct base value for each.

**Tests** — two vectors at dvparser level:

* OMS Spec Vol.2 Annex G, Tables G.6/G.7: inverse compact profile of a water meter, with the expected result printed in the document. The generated keys `CC 04 13`, `8C 05 13`, `CC 05 13`, `8C 06 13` are exactly the data information blocks listed in Table G.7.
* An EMH electricity meter (EMH 00/02) sending TAF7 data as described in [OMS Spec Vol.2 Annex R](https://oms-group.org/wp-content/uploads/2024/05/OMS-Spec_Vol2_AnnexR_A101.pdf): 64 bit base value of 95772 Wh with 1 byte increments, two profiles for import and export. Before this change the profile decoded as 0.000 / 0.001 / 0.000 kWh, now it decodes as 95.772 / 95.771 / 95.771 kWh.

The 15 existing compact profile tests are untouched and still pass; they all code base value and elements with the same width, so nothing changes for them. `make test` is green.

Both OMS documents are free downloads, so the vectors can be checked without access to the underlying paid standard.

I have two further findings from the same documents — element ordering of the forward compact profile (VIFE 1Fh), and variable length records with an LVAR of C0h or above. I would rather submit those separately, the first one needs re-measuring on top of this change.


weetmuts:
I do not understand. You must provide a telegram that demonstrates what is wrong.

weetmuts:
Again, telegrams that demonstrate the problem!

cgommel:
Fair enough. Two telegrams, both unencrypted, both runnable as one line.

**1. The example from the standard**

The compact profile from OMS Spec Vol.2, Annex G, Table G.6: a water meter, base value 1013 l on 2008-05-31, four monthly increments. Table G.7 of the same document prints what it has to expand to — 755, 423, 209 and 65 litres.

```
// demo.xmq
driver {
    name = democp
    meter_type = WaterMeter
    default_fields = name,id,total_m3,timestamp
    detect { mvt = MAD,01,07 }
    fields {
        field { name = total quantity = Volume
                match { measurement_type = Instantaneous vif_range = Volume storage_nr = 0 } }
        field { name = 'history_{storage_counter}' quantity = Volume
                match { measurement_type = Instantaneous vif_range = Volume
                        storage_nr = 1,4 add_combinable = Synthetic } }
    }
}
```

```
wmbusmeters --analyze=demo.xmq 264424340102030401077a010000200C1313100000026C1F150D93130A7AFE5802320314024401
```

master:
```
"history_1_m3":0.258, "history_2_m3":0.332, "history_3_m3":0.214, "history_4_m3":0.144
```
with this PR:
```
"history_1_m3":0.755, "history_2_m3":0.423, "history_3_m3":0.209, "history_4_m3":0.065
```

The dates come out right either way, only the values are wrong. The base value is coded as 8 digit BCD while the increments are 4 digit BCD, and the lookup demanded both use the same coding, so it found no base value and stored the raw increments as if they were readings.

**2. A real meter**

An EMH electricity meter sending TAF7 data. TAF7 is the meter reading profile use case from BSI TR-03109-1, specified for wireless M-Bus in OMS Spec Vol.2 Annex R. The meter has no clock, so instead of timestamped history it sends a base value and a series of increments running backwards at a fixed spacing, here 8 seconds. The purpose is redundancy: the profile covers about three transmission intervals, so the gateway still has a gap-free series if two telegrams are lost on air.

This is that meter's own payload, rewrapped into an unencrypted frame, meter id anonymised. Fields: storage 3 is the base value, storage 4 onwards the profile.

```
wmbusmeters --analyze=demo2.xmq AE44a8153412711300027a010000002F2F0700285EB5050000000007803C34080000000000000728905F01000000000004208F84160101FD971D00C1017402C401208D841601C701031C76010000000000CD01831324410800010000000001000000000100000001000000000100000000010000000001000000C701833C0200000000000000CD0183933C244108000000000000000000000000000000000000000000000000000000000000000000
```

| | base | history 1..5 |
|---|---|---|
| master | 95.772 kWh | 0, 0, 0, 0, 0 |
| this PR | 95.772 kWh | 0.002 x 5 |
| this PR + #2056 | 95.772 kWh | 95.772, 95.771, 95.771, 95.771, 95.771 |

A meter reading 95.772 kWh whose history says zero. This telegram carries two profiles, import and export, so it needs #2056 as well to come out right — on its own this PR gets as far as picking up a base value, but it picks the export one. The first telegram above has a single profile and isolates this defect on its own.

Why this could sit unnoticed: `caltose`, `fhkvdataiv` and `uiws` do consume compact profile output, but in each of their test telegrams the base value and the increments use the same data field coding, so none of them takes this path. That is also why the suite stays green with this change.

weetmuts:
Many thanks! I added a test case tests/test_inverse_compact_profile.sh based on your simulated telegram.

No changes with this kind of significance can go into wmbusmeters without test-cases and example telegrams. :-D

Issue #2055, opened 2026-08-11
Compact profiles without a base time have no time reference in the output


cgommel:
When a compact profile comes with a base time, `dvparser` already generates a date for every profile point. When it does not, the points end up in the output with no time reference at all: the spacing is read in `decodeCompactProfileHeader()`, used to expand the profile and then dropped, so no driver can see it.

That is not an exotic case. TAF7 meter reading profiles as described in [OMS Spec Vol.2 Annex R](https://oms-group.org/wp-content/uploads/2024/05/OMS-Spec_Vol2_AnnexR_A101.pdf) are meant for meters without a clock — R.3.2 explicitly accepts an invalid base time and points at the actuality duration as the reference instead. The receiving system is expected to supply the absolute time.

### Concrete case

An EMH electricity meter (EMH 00/02, which the `nzr` driver already claims) sends two inverse compact profiles on storage 3, spacing control `41h`, spacing value `08h`, 34 elements each, plus an actuality duration of 2 s. The profile covers the last four and a half minutes, but nothing in the decoded output says how far back any point lies.

The spacing of that telegram is confirmed three ways: OMS Spec Vol.2 Annex N, N.13 prints exactly the same two bytes and labels them "Increments + 1 Byte" and "8 seconds spaceing"; the profile increments by 1 Wh every fifth element, which is 1 Wh per 40 s; and the same telegram reports 90 W as current power, which is the same number. Any other spacing breaks that agreement.

### Question

Should `dvparser` emit a relative time per profile point when no base time is available, mirroring what it already does with dates? An actuality duration entry (VIF `74h`) under the point's storage number fits the existing machinery, and a driver could then match it like any other field. `74h` also happens to be a single byte VIF, so `makeSyntheticStorageKey()` works unchanged, whereas `DurationSinceReadout` would be `FD 2C` and needs it extended.

A candidate implementation is on a branch: https://github.com/cgommel/wmbusmeters/tree/compact-profile-relative-time

It is deliberately not a pull request. It changes what drivers can match on, and I would rather have a decision on the intended output format before that lands anywhere. It builds on #2054, and `make test` is green with it.

The alternative I considered and discarded was letting each driver hardcode the spacing of its meter. That works until someone reconfigures the meter, and OMS uses 6 s and 24 s in its own worked examples, so the value really has to come from the telegram.

cgommel:
Two notes while building the candidate.

[...]

PR #2056, opened 2026-08-11
fix: keep the combinable vifs of a compact profile on its generated points

cgommel:
Builds on #2054 and should go in after it — it uses the helper introduced there, so until that one lands the diff shown here contains its two commits as well.

A telegram can carry two compact profiles for the same quantity, told apart only by a combinable vif such as `BackwardFlow`. The points generated from them dropped that combinable and kept only the `Synthetic` marker, so both profiles ended up as entries that a field matcher cannot distinguish — `FieldMatcher::matches()` compares the combinable set exactly, in both directions. A driver asking for the profile storage numbers therefore receives whichever profile the hash map happens to yield first.

How it showed up: an EMH electricity meter (EMH 00/02) sends an import and an export profile as described by [OMS Spec Vol.2 Annex R](https://oms-group.org/wp-content/uploads/2024/05/OMS-Spec_Vol2_AnnexR_A101.pdf), both on storage 3 with 34 elements. A driver matching those storage numbers got all 34 import readings filled with the export value: 0.002 kWh where 95.772 kWh belongs. No warning, and the numbers look like a meter that simply has not consumed anything.

The generated points now carry the combinable vifs of the profile they came from, plus the `Synthetic` marker. A driver can then address the two separately, with `add_combinable = Synthetic` and `add_combinable = Synthetic,BackwardFlow`.

**Test** — the existing EMH vector gains two assertions on the generated points, using a small `tst_combinable()` helper in the style of `tst_subunit()`. Reverting the change alone makes it fail with `key 8702037F77_2 is missing combinable BackwardFlow`. `make test` is green.


weetmuts:
Many thanks! It seems to work. This is quite complicated stuff. 
I wrote an improved driver for the EMH meter and your reference telegram. Check the decode here:

https://wmbusmeters.org/analyze/AE44a8153412711300027a010000002F2F0700285EB5050000000007803C34080000000000000728905F01000000000004208F84160101FD971D00C1017402C401208D841601C701031C76010000000000CD01831324410800010000000001000000000100000001000000000100000000010000000001000000C701833C0200000000000000CD0183933C244108000000000000000000000000000000000000000000000000000000000000000000

Check the driver itself here:
https://wmbusmeters.org/drivers/ehzp.xmq.html

It seems like there is no way to calculate the back dates for the historic at values?

cgommel:
Yes, it can be calculated — the information is in the telegram, it just gets dropped. The spacing value byte of the profile (`08h`, 8 seconds here) is read in `decodeCompactProfileHeader()` and discarded once the profile is expanded, so no driver can see it.

This meter sends no base time, which is normal for TAF7 profiles: OMS Spec Vol.2 Annex R, R.3.2 accepts an invalid base time and uses the actuality duration as reference instead. So the points can only be placed relative to the transmission. The base value lies `actuality_duration_at_1` before it (2 s here), and point *n* another *n* × spacing before that — 10 s, 18 s, 26 s … for this telegram.

I described this in #2055, with a candidate that emits a relative time entry per profile point. I'll rebase it onto current master and `ehzp` and send it as a PR with a test case.

PR #2057, opened 2026-08-11
qwaterv2: detect QDS 19/06, 35/06 and 35/07

cgommel:
Qundis radio converters announce the adapter in the link layer (`QDS,35,37`) and the meter itself in the transport layer (`QDS,35,06` or `QDS,35,07`). `pickMeterDriver()` prefers the transport layer triple when a long transport header is present, so those are the ones that have to be listed — but the detect block only had the link layer form `35,37`. Version 19 has the same hole the other way around: `19,07` is listed, `19,06` is not, while every neighbouring version (16, 17, 18, 1A, 1D) is listed with both media types.

The result is that meters the driver decodes perfectly well are simply not found.

The test block already contains the proof. Two of its telegrams carry a transport header of `35,07` and `35,06`:

```
$ wmbusmeters --analyze 3C449344682268363537726666666693443507720000200C13670512004C13611003...
Auto driver    : not found!
$ wmbusmeters --analyze 3C449344333333333537723333333393443506B80000200C13350000004C13000000...
Auto driver    : not found!
```

Both decode completely once the driver is named, which is why this went unnoticed: driver tests pass the driver explicitly and never exercise detection.

I ran a set of captured telegrams through `--analyze` and hit all three variants in the field, `19,06` once and `35,06` and `35,07` three times each, all decoding fully (7/7 and 20/20 records understood) once the driver was forced.

**Test** — one new case for `19,06`, a walk by telegram with manufacturer data and twelve `volume_delta` values; that frame shape (no long transport header) was not covered yet. No new tests for `35,06` and `35,07` since the telegrams quoted above already cover them. The meter id in the added telegram is anonymised.


Uzlopak:
LGTM

Uzlopak:
Screams for a follow up, regarding testing that the driver is actually detected

cgommel:
Thanks for looking!

Agreed, and I think the gap is narrower than it sounds: driver tests always pass the driver name, so `pickMeterDriver()` never runs in the test suite at all. That is precisely how two telegrams sitting in `qwaterv2`'s own test block could stay undetectable without anyone noticing.

A cheap check would be to take every telegram from the driver tests and assert that auto detection returns the driver the test belongs to. That alone would have caught this one, and I suspect it is not alone — sweeping a few hundred captured telegrams here turned up six more manufacturer/version/type combinations that no driver claims, several of them one line away from a driver that already handles the family.

Happy to put that together. Where would you want it to live — inside the framework from #1994, or next to the existing `test_case.sh` for now? I would rather not build across what you have in flight there.

And since you asked for feedback on #1994: having tests sit next to the code they cover would have helped me last week. I added a handful of compact profile vectors and they went into `testinternals.cc`, which is by now a long way from `dvparser.cc`.

weetmuts:
What a mess. But we can just add more mvt:s. So I can merge this. 

We should improve the testing so that the tests inside the driver fails if the auto does not match the tested driver....

PR #2060, opened 2026-08-17
Update ista meter drivers (IT-480)

bbiegun:
This PR  updates two Ista meter drivers (`istaheat` and `istawater`) to support new models.  The PR includes telegrams from new models of meters in the tests and enhances the drivers' documentation.

I've added possible compatible products based on the one I own and the [list of models](https://www.ista.com/fileadmin/media_ista/polska_pl/Documents/Technologies/Heat_meter/sensonic3_092025.pdf). Sadly, I failed to find same document in English. 
In the (EU declaration of conformity)[https://www.ista.com/fileadmin/media_ista/ceska-republika_cz/Support/Download/POS_sensonic_3.pdf] there is even longer list of models, though, I have no means to assume which one would be compatible with the OMS mode. Therefore, I've limited possible supported models to the ones which I'm mildly  confident with.

New `radio net 3` devices includes two additional fields (date, volume flow). I don't see in any Ista document what they may be representing - left them unparsed for now. The [OMS documentation](https://oms-group.org/wp-content/uploads/2025/03/L.83.0186_Technisches-Datenblatt_Funkmodul-fuer-modulare-Wasserzaehler-radio-net-3_DE.pdf) list only current value, value at end of last month and the reference date.


weetmuts:
Thanks!

Issue #2025, opened 2026-07-08
# qwaterv2: encrypted WalkByDataSet (0DFF5F) of 2026 Qundis QWater 5.5 is randomly mis-decoded as garbage `total_m3` (~1/256 per telegram)


ghost:
### Type of request

Existing device improvement

## Environment

* wmbusmeters `3.0.0-7-gff7c7d28` (built from a git checkout), Debian 13 on Raspberry Pi
* Dongle: IMST iU891A (`device=/dev/ttyACM0:iu891a:t1,c1`)
* Meter: Qundis QWater 5.5 cold water, built 2026, `QDS,1A,07`, id `60555885`, driver `qwaterv2`, mode-5 key configured (regular `CI=7A` frames decrypt fine)

## Symptom

Sporadically the meter reading jumps by several orders of magnitude for a single update, with `status: OK`, and then returns to normal. From `meterfiles` (meter is at ~1.4 m³):

```
2026-07-08T05:40:07Z total_m3=1.387     target_m3=1.248     target_year_m3=0
2026-07-08T05:41:58Z total_m3=15430.611 target_m3=34228.176 target_year_m3=10377.559   <-- garbage
2026-07-08T05:46:00Z total_m3=1.387     target_m3=1.248     target_year_m3=0
```

This silently corrupts downstream consumers (e.g. Home Assistant long-term statistics / utility meters).

## Root cause

The 2026-generation QWater 5.5 **encrypts** the proprietary WalkByDataSet block (`0DFF5F`, CI=78 frames). Note that CI `78` carries no TPL header, so there is no wM-Bus-level encryption on these frames — this is proprietary encryption *inside* the manufacturer block. Observed layout:

```
00 82 <cnt> 00 35 <48 bytes high-entropy data>      (53 bytes total)
      ^shares the access-counter sequence with the CI=7A frames
```

**Confirmed by decryption** (details below): the 48-byte body is AES-128-CBC encrypted with the **same meter key** already configured for the CI=7A frames, using the standard mode-5 IV construction `M-field(2) + A-field(6) + cnt×8`, where `cnt` is header byte `[2]`. `byte[4]` changed from `0x00` (plaintext generation) to a constant `0x35`. The decrypted 48 bytes are byte-for-byte the **legacy WalkByDataSet layout** starting at old offset 5 (`<flags(2)> 07 C1 13 <err_date> <total> <ty_date> <ty> <t_date> <target> <12 deltas> 2F`).

[...]

PR #1941, opened 2026-05-15
feat: qcaloric, qheat and qwater add processing of WalkByDataSet

Uzlopak:
=D

How did I do that... you may ask...

Issue #1916, opened 2026-05-06
Qundis Q Water 5.5


EmKayP:
Explain the problem.....
Hi,

First, thanks a lot for your work on this add-on.

Please ignore my non-technical terminlogies I've used here - I'm not an expert, so I'm trying to explain the problem so as others benefit from what I learn. Hopefully my message is clear. Would be happy to share more information if required.

I've noticed on the logged telegrams by the HA wmbusmeters add-on, each Q Water 5.5 meter appears to be sending 3 different format of the telegrams.
1st short-form which appears to be decoded correctly
2nd long-form which the wmbusmeter appears to be truncated when checking the MQTT feed
3rd a weird telegram which has some manufacturer specific blocks which appears to be ignored entirely.

My this report is about the 2nd long-form telegram 

See this one for eg., from the last month
https://wmbusmeters.org/analyze/49449344782863521A07780DFF5F350082720000800000C113FFFF029000003F3C192500005F330073000000000000000000000000000000004700B100B1008600A8002F046D00135E34:qwater

There are some 13 blocks of code which doesn't seem to be translated

And then the one from 1st of this month
https://wmbusmeters.org/analyze/49449344782863521A07780DFF5F350082CC0000610107C113FFFF159000003F3C192500005E34119000000000000000000000000000004700B100B1008600A800AB002F046D01064135:qwater

And shows at least 4 blocks of data preserved from the previous month with one new block of data added

<img width="936" height="54" alt="Image" src="https://github.com/user-attachments/assets/c2dae8e6-e686-44f3-a05e-2eae8d879103" />

I think part of it is the reading from the previous "billing cycle" and my theory is that it stores the data for the 13 last cycles. But however my meters are new and I don't have any way to verify all of the remaining zeros - perhaps someone else with an older meter can verify? The same behaviour is also seen on the other meter for hot water too.

I will raise a new issue when I manage to crack that third telegram because I _think_ this is causing the wmbusmeters HA add-on to crash with a Segmentation fault  (Error code 139) unless someone knows it is a known issue.

Edit:
The third weird telegram is just an heartbeat from the meters every 8 hours on the clock carrying some device identifiers but otherwise doesn't seem to carry any other useful data. And this is not related to the segfault errors. I've just seen the other thread about it too, so I'll post there or create a new issue when I have more info.

weetmuts:
Which version of wmbusmeters are you running?

weetmuts:
[...]

Issue #1497, opened 2025-02-22
qheat measures differers bei 10


hngfngzl:
### Type of request

Question

### OS version

raspberry debian

### wmbusmeters version

1.6.1.29

### Your message goes here

We face the situation that the system runs fine, but it brings different values for the same counter:

meter_id	timestamp		timeslot		verbrauch	stichtag	verbrauch_stichtag	status	rssi_dbm
67961176	2025-02-22 12:45:02.000	2025-02-22 12:45:00.000	308.700		31.12.2024	174.700			OK	135
67961176	2025-02-22 12:30:04.000	2025-02-22 12:30:00.000	3.087.000	31.12.2024	1.747.000		OK	124

The difference ist factor 10 - so far easy to tell. But we can't tell why this happens. At the moment we don't have a way to filter this, because the data so far looks fine. 
The used driver is qheat. 

The telegrams bring this a lot and we haven't found a system/reason for the shorter or the longer values. 

Any Ideas?

schmyd:
Hi,

I have the same issue, but it seems like there are multiple issues already concerning the same topic
 * #1492 (also a factor of 10 here - people just thought it is an issue about printing commas or not)
 * #1475 highlights the parsing issue when trying to interpret the "wrong" message which produces data that is a factor of 10 too low
 * #1347 Again a factor of 10. There @gotsila already concluded that only one message type should be used. Unclear if this can be filtered
 
Maybe those threads could be combined to resolve this? I am struggling with the same issue.

### Further analysis
Example telegrams of the issue
 * ["Bad" message, 10x too low, contains unparsable block](https://wmbusmeters.org/analyze/5844934457278167463778077957278167934446040DFF5F3500827C0000610107C006FFFF093500001F3C851700003F31572600003C007F00070000000000000000000000DF00B102A70268032F02FD170000046D3B0E3B32:qheat)
 * ["Good" message (also shorter)](https://wmbusmeters.org/analyze/414493445727816746377257278167934446047C0000200C06093500004C0685170000426C1F3CCC080657260000C2086C3F3102FD170000326CFFFF046D390E3B32:qheat)

It feels that the first message isn't send during nighttimes, but this is just a first guess.
Any pointers on how to debug this further or properly interpret the message would be welcome.

### Workaround
A workaround for home assistant users: adjust the value template in your mqtt sensor to automatically multiply the value by 10 if it is below a certain threshold (1000 in this example):
```
[...]

PR #2070, opened 2026-08-27
qwaterv2: fail closed on walk-by blocks with an unexpected layout

bartman081523:
## What

The `0D FF 5F` (WalkByDataSet) ixml grammar in `qwaterv2.xmq` skips the 9-byte proprietary header with `decode = quad, quad, byte, volume_l` and then matches `volume_l = '13', ...`. When a block carries the `0DFF5F` difvifkey but does **not** have the plaintext walk-by layout, the grammar falls through to best-effort matching at wrong offsets and invents values. Reported in #1916 (Q Water 5.5, "no data since ~19th June").

Reproduced against current master with @docbobo's telegram (`4dedd717`):

```
$ printf '%s\n' 49449344016135141907780DFF5F350082C100352CAA738213487EAF84ED6BFD4DA668870285651771A8D3CFB7DE5ECC22081CDD1A3FFBA6506F901027DB29B4672708D9046D160A4F38 | build/wmbusmeters --format=json stdin:hex kaltwasser qwaterv2 14356101 NOKEY
{"_":"telegram","media":"water","meter":"qwaterv2","name":"kaltwasser","id":"14356101","target_m3":204787.117,"target_date":"2052-05-05","target_year_m3":2876.976,"total_m3":80308.592,"volume_delta_2_m3":-18.481,...}
```

`total_m3=80308`, `target_date=2052`, negative deltas — published to MQTT, which makes HA drop the entity.

## Root cause (evidence-based; no public Qundis spec exists)

There is **no public Qundis/Siemens spec** for the WalkByDataSet byte layout. PR #1941 (where the walk-by ixml was written) cites nothing. The header structure is community reverse-engineering, documented by @justjanne in #1497:

```
[0]=00 reserved | [1]=0x82 magic | [2]=access | [3]=00 | [4]=00 | [5]=status_hi | [6]=status_lo | [7]=07 | [8]=metertype(0xB0=HCA,0xC0=heat,0xC1=water) | [9]=VIF(0x13=volume,...)
```

`0x82` is a constant Qundis magic number (constant across generations and device types; not the M-field 0x44, not the DLL type) — see #1497. The Qundis data sheets (Q water / heat / caloric 5.5) confirm a "Q walk-by (PDW)" proprietary block with AES-128 Mode 5/7, but do **not** document the bytes.

@docbobo's block is `00 82 C1 00 35 2C AA 73 82 13 ...`:
- `[1]=0x82` — valid WalkByDataSet magic ✓
- `[4]=0x35` — **not** `0x00`. Per @folfy in #2025, `byte[4]=0x35` is the **encrypted-frame marker** for the 2026 Q Water 5.5; plaintext frames have `0x00` there. @folfy verified on 36/36 captured frames that the block is **AES-128-CBC** encrypted with the configured OMS Mode-5 meter key (IV = M-field(2) + A-field(6) + access_counter×8). So `[7]=73`, `[8]=82`, `[9]=13` are **ciphertext**; `13` matching the VIF is a ~1/256 coincidence, and the parser invents the rest from wrong offsets.

## Fix (interim guard; not the complete fix)

Validate the 9-byte header against its fixed plaintext bytes instead of skipping it blind:

```
decode = -'00', -'82', byte, -'00', -'00', byte, byte, -'07', -'C1', volume_l.
```

A block with an unexpected header (the encrypted variant has `[4]=0x35`, `[7]≠07`, `[8]≠C1`) now **fails closed**: the invented `total_m3`/`target_*`/`volume_delta_*` fields are omitted and only the real trailing `046D meter_datetime` record is published. All existing `qwaterv2` walk-by tests (MyQWater, QWooo, QWaaa, QWddd, QQ1, QQ2, QQ5×2) carry `[4]=00, [7]=07, [8]=C1` and pass; `qheatv2`/`qcaloric`/`lse_07_17`/`lse_08` also unaffected.

**This is deliberately an interim guard, not the complete fix.** It stops the MQTT garbage that breaks HA, but it does **not** recover `total_m3`/`volume_delta_*` for the 2026 encrypted Q Water 5.5 — those values are inside the AES-encrypted manufacturer block and need core AES support, not an ixml change.

## The complete fix (out of scope here; needs C++ core)

To actually decode the 2026 encrypted Q Water 5.5, the manufacturer block must be AES-128-CBC decrypted in the C++ layer (ixml cannot do AES) before the existing ixml decoder runs. @folfy proved in #2025 that decrypting the block with the configured meter key + mode-5 IV yields the plaintext `... 07 C1 13 ...` layout, which the **unmodified** `qwaterv2` decoder then parses correctly. That belongs in `qdsExtractWalkByField()` / `src/manufacturer_specific.cc` (or its v2 successor), gated on the `byte[4]==0x35` encrypted marker — and is a separate, larger piece of work. I'll open it as a follow-up.

## Test

One new `test { }` case in `qwaterv2.xmq` (the reported telegram), expecting fail-closed JSON. `cd drivers && make testonly DRIVER=qwaterv2` passes, as do `qheatv2`, `qcaloric`, `lse_07_17`, `lse_08`. `src/generated_database.cc` and `tests/generated_tests.xmq` regenerated via `make install`.

Related: #1916, #1497, #2025.

(Note: my earlier `qwaterv2.xmq` touch #2004 only changed `0xffff`→`null`; it did not touch the `0DFF5F` grammar — flagging so the investigation points at the right commit.)


weetmuts:
Nice! What is the next step after I merge this?

justjanne:
Note thay this entire analysis seems to be AI generated. I can't find any obvious mistakes, but especially with LLMs it's common that things look OK at a surface level while being broken at a deeper level.

bartman081523:
Real-data verification note (in addition to the synthetic tests in this PR):

- **Fail closed confirmed on real encrypted telegrams (water and heat).** @docbobo's ciphertext from #1916 (`…0DFF5F3500820500 35 868E…`) now ends in `ixml:state = failed` — no invented `total_m3`/`target_*`/`volume_delta_*`, only the real `046D` meter_datetime record is published. Same result for the 10 further real encrypted walk-by telegrams I've collected from #1497/#1588/#2025/#2051 — 11/11 fail closed in total, for both `qwaterv2` and `qheatv2`.
- **The unmodified grammar stays correct for real plaintext blocks.** The three plaintext walk-by samples in #1497 (`0x00` header variant) still parse, as do all existing tests.
- **With the meter key configured, the same real telegram decodes end-to-end** (follow-up work, gated on `byte[4]==0x35`): decrypting @hngfngzl's #1497 telemetry reproduces their posted JSON exactly (`total_m3` 26.546, `target_m3` 24.086, `target_year_m3` 20.552, `target_date` 2026-01-31, `target_year_date` 2025-12-31) — which also pinpoints the CI=0x78 IV detail: `IV = DLL[2:10] ‖ ACC(block byte[2])×8`, since these frames have no TPL to take the ACC from.

So: guard = correct interim fix for the #1916 garbage, and the encrypted blocks it now refuses are provably decodable when an owner has the key. The full decode is implemented and end-to-end verified against the two real (key, telegram) pairs published in #1497 — reproduces the meter owner's posted values field-by-field. Follow-up PR to come.

bartman081523:
> Note thay this entire analysis seems to be AI generated. I can't find any obvious mistakes, but especially with LLMs it's common that things look OK at a surface level while being broken at a deeper level.

@justjanne 
Thank you so much for your review. Yes, you are correct, the whole code and analysis was generated by glm-3.5-flash running in Claude Code running my agents.md fork. I started this because #1916 guilt-tripped me into analysing whether my agent was the source of the problem or not. 
Your bitwise explanation of the qundis telegram helped a lot!
Please feel free to falsify the code and tests.

bartman081523:
> Nice! What is the next step after I merge this?

Next step: **PR #2074**. It contains this PR's commit plus the full decode — it is already verified against the two real (key, telegram) pairs that @hngfngzl published in #1497:

- water 60170138: decrypts to exactly the values you see in their log posting (`total_m3` 26.546, `target_m3` 24.086, `target_year_m3` 20.552)
- heat 68369440 (the one that stayed empty in their log): `total_kwh` 3052, `target_kwh` 2771, full 12-month history

Merging order doesn't matter:

- merge #2070 first → #2074 shrinks automatically to the AES delta (its base already contains this commit)
- or close #2070 in favour of #2074 — the guard is identical text over there

Both are CI-green and mergeable right now (#2074 = CLEAN/MERGEABLE too).

Answering @justjanne's (fair) point about deep-level correctness: the AES path in #2074 is not just reviewed-looking — it is proven against ground truth: 11/11 real encrypted walk-by telegrams from #1497/#1588/#2025/#2051 fail closed without a key (no invented values), and decrypting with the owner's keys reproduces their published readings field-by-field. The one negative result I'd highlight honestly: a sweep of ~1100 standard serial→key derivations found **no** key for either real key, so keys remain per-meter random and the existing `--key` configuration stays the correct input.

I'd also be happy to add a real-key regression test (key + telegram from #1497 are already public) if you'd like that in the PR — or keep the synthetic key vector and real-key verification out, your call.

weetmuts:
I think I merged this as part of the other PR:s right? Can you check @bartman081523 ?

bartman081523:
> I think I merged this as part of the other PR:s right? Can you check @bartman081523?

Yes — checked: the guard content is in master. Both `drivers/src/qwaterv2.xmq` and `drivers/src/qheatv2.xmq` on master carry the fail-closed `0DFF5F` header guards (`-'00', -'82', byte, -'00', -'00', …` for water, the `-'C0'` variant for heat), together with the encrypted-marker test cases (`Kaltwasser`, `HeatEnc55`).

It arrived via #2074, which was squash-merged (`d585deb9`) — that is why this PR wasn't auto-marked as merged (a squash merge keeps this PR's own commit out of master's lineage), but everything it contained is in. So this PR can be closed whenever you like.

And yes — opening the new PR with the real-key test vectors (water `60170138` + heat `68369440` from #1497) now, as requested.

PR #2071, opened 2026-08-27
Add flonidangas driver: status/id only for encrypted gas meter (FLO,50,03)

taufderl:
## What

Adds `drivers/src/flonidangas.xmq`, a minimal driver for Flonidan gas
meters identifying as `mvt FLO,50,03`.

Every telegram from this mvt combination sets `tpl-cfg` to AES-CBC
(security mode 5, confirmed via `--analyze`) - the entire data-record
only the utility that installed the meter holds. There's no way to
write (or test) field definitions for `total_m3`/etc. without a real
decrypted telegram to check the layout against.

What *is* readable without a key is the transport-layer header - meter
id and the tpl status/alarm byte - so this driver exposes just those
known here).

## Verification

`oneshot=true`:

```
{"_":"telegram","media":"gas","meter":"flonidangas","name":"ScanAll","id":"12345678","status":"ALARM UNKNOWN_20","timestamp":"...Z"}
```

re-ran the existing built-in `elster.xmq` test telegram with these new
drivers loaded alongside it to confirm no interference - unaffected.

## What I could NOT do


taufderl:
Update: the \`xmq\` CLI became available in my environment after this was opened, so I've now run the actual documented workflow (\`cd drivers && make pretty && make install\`), pushed \`src/generated_database.cc\`/\`tests/generated_tests.xmq\` regenerated for just this driver, and verified with \`make testonly DRIVER=flonidangas\`, the full \`make test\` suite, and \`./build/testinternals\` - all clean.

One side note from running the full regeneration: it surfaced pre-existing drift unrelated to this PR - \`istaheat\`/\`istawater\`'s \`.xmq\` sources have extra \`mvt\` entries that were never propagated into the compiled database (looks like a prior merge that didn't get \`make install\` run for it). I left that untouched since it's out of scope here, but flagging in case it's useful to know about.

jktjkt:
What’s the point of providing a driver which does not decode any readings? The `lansenpu` driver also provides the `a` and `b` pulse counters, which are **the** reason why somebody vuys a pulse counters. Presumably, with a gas meter, you’re interested in volume reading.

taufderl:
Good point, and fair question. Every telegram from this meter sets AES-128-CBC (mode 5), so total_m3 is genuinely locked behind a secrete per-meter key. lansenpu's a/b counters are actually sent unencrypted, so it's not quite the same situation as this one.

What is in the clear is the id and status/alarm byte, and for me that's the actual point: it turns "unknown telegram" into a named, identifiable meter and surfaces alarm codes (tamper, error, etc.) I can act on, which is genuinely useful even without volume.

Totally happy to drop this if it's not wanted upstream though — just say the word and I'll close it.

jktjkt:
> lansenpu's a/b counters are actually sent unencrypted, so it's not quite the same situation as this one.

This is factually incorrect.

weetmuts:
@taufderl sounds like an AI. These pull requests are completely useless.....
If you just want to extract the tpl_status, write your own driver: onlytplstatus and use that for any meter
for which you do not have the decryption key for.

taufderl:
sorry I bothered. Just delete them then. I appreciate wmbusmeters and will keep running it with this driver.

weetmuts:
@taufderl As soon as you have decrypted telegrams, then we can actually do something useful. Keep in touch!

PR #2072, opened 2026-08-27
Add landisgyrgas driver: status/id only for encrypted gas meter (LGB,50,03)

taufderl:
## What

Adds `drivers/src/landisgyrgas.xmq`, a minimal driver for Landis Gyr gas
meters identifying as `mvt LGB,50,03`.

Every telegram from this mvt combination sets `tpl-cfg` to AES-CBC
(security mode 5, confirmed via `--analyze`) - the entire data-record
only the utility that installed the meter holds. There's no way to
write (or test) field definitions for `total_m3`/etc. without a real
decrypted telegram to check the layout against.

What *is* readable without a key is the transport-layer header - meter
id and the tpl status/alarm byte - so this driver exposes just those
known here).

## Verification

`oneshot=true`:

```
{"_":"telegram","media":"gas","meter":"landisgyrgas","name":"ScanAll","id":"87654321","status":"ALARM UNKNOWN_20","timestamp":"...Z"}
```

re-ran the existing built-in `elster.xmq` test telegram with these new
drivers loaded alongside it to confirm no interference - unaffected.

## What I could NOT do


taufderl:
Update: the \`xmq\` CLI became available in my environment after this was opened, so I've now run the actual documented workflow (\`cd drivers && make pretty && make install\`), pushed \`src/generated_database.cc\`/\`tests/generated_tests.xmq\` regenerated for just this driver, and verified with \`make testonly DRIVER=landisgyrgas\`, the full \`make test\` suite, and \`./build/testinternals\`.

One note: the full suite has a pre-existing failure unrelated to this PR - \`istaheat\`'s committed test in \`tests/generated_tests.xmq\` has a malformed json string (missing a quote before \`"name"\`). Not something introduced or touched here.

PR #2073, opened 2026-08-27
Add elstergas driver: status/id only for encrypted gas meter (ELS,50,03)

taufderl:
## What

Adds `drivers/src/elstergas.xmq`, a minimal driver for Elster gas meters
identifying as `mvt ELS,50,03`.

## Why a separate driver from the existing elster.xmq

The existing `elster.xmq` covers `mvt ELS,81,03` and decodes `total_m3`
from an unencrypted payload. `ELS,50,03` is a different meter
version/firmware that sets `tpl-cfg` to AES-CBC (security mode 5,
confirmed via `--analyze`) on every telegram - the entire data-record
only the utility that installed the meter holds. Since the two versions
have nothing in common once one is encrypted and the other isn't, this
is a separate driver rather than an added `mvt` line on the existing
one.

There's no way to write (or test) field definitions for `total_m3`/etc.
without a real decrypted telegram to check the layout against. What *is*
readable without a key is the transport-layer header - meter id and the
tpl status/alarm byte - so this driver exposes just those
known here).

## Verification

`oneshot=true`:

```
{"_":"telegram","media":"gas","meter":"elstergas","name":"ScanAll","id":"11223344","status":"ALARM TEMPORARY_ERROR UNKNOWN_20","timestamp":"...Z"}
```

re-ran the existing built-in `elster.xmq` test telegram (mvt ELS,81,03)
with this new driver loaded alongside it to confirm no interference -
both produced correct, distinct output.

## What I could NOT do


taufderl:
Update: the \`xmq\` CLI became available in my environment after this was opened, so I've now run the actual documented workflow (\`cd drivers && make pretty && make install\`), pushed \`src/generated_database.cc\`/\`tests/generated_tests.xmq\` regenerated for just this driver, and verified with \`make testonly DRIVER=elstergas\`, \`make testonly DRIVER=elster\` (the existing driver, to confirm no interference), the full \`make test\` suite, and \`./build/testinternals\` - all clean.

weetmuts:
Closing until a decrypted telegram arrives.

Issue #1588, opened 2025-09-20
Qundis Qwater 5.5 can't be decoded


Dag0d:
### Type of request

New device support

### Meter description

Great project, I just got a new Qwater 5.5 installed but somehow the values are not decoded correctly.

Here a few examples:

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f3500825c0035dc3284b72eca9c1bb6b7331aa586f6616b72095f046b3ce6af7de7d5945d5be354d25ccc51a0872a78ff8423d12fbd63046d05063439

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f350082650035ad00a6dacea959767a122002d3bcce266598c5f6828339f03132217c545224d34215ecd8842f404ca21ab6054c6e0d1c046d0f073439

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f350082730035dbfc9be8af8ffaafeb1ef5a9f40207c76405c3118544693623e14cebc1615c689eb4792f17afe16f8d7a4970d5e5064c046d36083439

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f3500827c00356219b6ebe30566d907527a31f28334536bf63790789384e16fedb0128bfb3db531fce343b4ceaa5be37aa98ea3f4ce83046d030a3439

**Between all the reading above some water was consumed, below a few example where the water consumption stayed the same:**

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f350082650035ad00a6dacea959767a122002d3bcce266598c5f6828339f03132217c545224d34215ecd8842f404ca21ab6054c6e0d1c046d0f073439

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f3500826a003569c63d4a2161e062110587ff378177e31f54839888e9a0b3242e5443ec24634b4e4bfb81d6daa3e69a154dd4df986630046d2e073439

https://wmbusmeters.org/analyze/49449344398915531a07780dff5f350082700035b04281331cb64be146182cdcd08e9d80d534acb3ac38f03ee5f1672dde7356272c5e20eb42650c7a5d397c0a669da53b046d1e083439

I've tried to find some pattern but did not succeed, could you help me to find out whats going on here?

Thanks in advance 

### Logged telegrams from meter in decrypted format

```shell
[...]

PR #2074, opened 2026-08-31
qwaterv2/qheatv2: AES-decrypt encrypted Qundis walk-by blocks (EN 13757-7 mode 5)

bartman081523:
## What

The 2026 Q Water/Heat 5.5 modules send the `0DFF5F` WalkByDataSet block **AES-128-CBC encrypted** (content `byte[4]=0x35`, per #2025 and the owner-supplied keys in #1497) while the driver grammar expects the legacy plaintext layout — so key-less parsers silently walk into ciphertext (the garbage in #1916). This PR adds the missing decode:

- **C++ decrypt hook**: `MeterCommonImplementation::tryDecodeQundisWalkByAes()` — when a configured meter key exists and the block header shows the encrypted marker, the 48-byte payload is decrypted **before** the (unchanged) ixml grammars run, then the marker byte is normalized to the legacy `0x00` layout so parsing continues transparently.
- **CI=0x78 IV detail** (verified with real keys from #1497): these frames carry **no TPL**, so the IV access counter comes from the walk-by block itself:

  ```
  IV = DLL[2:10] (M-field ‖ A-field ‖ version ‖ type) ‖ ACC (block byte[2]) × 8
  AES-128-CBC over the 48 bytes after the 5-byte content header
  00 82 <ACC> 00 <0x35 encrypted | 0x00 plaintext> <48 bytes>
  ```

- **Fail-closed grammar guards** for `qwaterv2` **and** `qheatv2` (the `qwaterv2` half matches #2070; `qheatv2` had the same exposure for real encrypted heat telegrams).

## Verification (real data, no synthetic-only claims)

- **Water 60170138** (telegram + owner's key from #1497): decodes **byte-identical** to the values the meter owner posted — `total_m3` 26.546, `target_m3` 24.086, `target_year_m3` 20.552, `target_date` 2026-01-31, `target_year_date` 2025-12-31.
- **Heat 68369440** (telegram + owner's key from #1497): `total_kwh` 3052, `target_kwh` 2771 (2026-01-31), monthly history 517/457/375/364/370/354/334 kWh.
- **Fail-closed**: all **11** real encrypted walk-by telegrams collected from #1497/#1588/#2025/#2051 produce no invented values with this PR applied (water **and** heat).
- New tests: `AESTest` (synthetic roundtrip), `Kaltwasser` (real ciphertext, no key → fail closed), `HeatEnc55` (real heat ciphertext, no key → fail closed). Full `drivers` test suite green.
- A serial-key-derivation sweep over ~1100 standard derivations (padded/hashed/xor serial forms) against the two published real keys produced **no** key — so keys stay per-meter random and the `--key` configuration remains the correct input, no auto-derivation is attempted (`#281` documents the same conclusion for newer Diehl meters).

## Out of scope / follow-ups

- `qcaloric` / `lse_07_17` carry the same `0DFF5F` grammar and would benefit from the same guard; no real encrypted sample at hand, so left untouched here.
- If wanted, a real-key test vector (key + telegram already public in #1497) can be added to the test block — happy to include one if maintainers prefer that over the synthetic-only vector.

Related: #1916, #2025, #2070, #1497, #1588


weetmuts:
Many thanks! 

Yes, please add the real-key test vectors (key + telegram already public in https://github.com/wmbusmeters/wmbusmeters/issues/1497)

weetmuts:
I added so that the driver has to register transform_payload=try_qundis_decode
otherwise it will not try this decode process. And modified the qwaterv2 and qheatv2 drivers with this.

I also moved the actual code for the tryDecodeQundisWalkByAes into manufacturer_specifities.cc

weetmuts:
Open a new PR for the test vector.

PR #2078, opened 2026-09-08
Add driver for Ebeco EB-Therm 500 floor heating thermostat

magnfris:
Added a new driver for Ebeco EB-Therm 500 floor heating thermostat with EB-Connect wM-Bus module.

Decoded fields:

Total energy consumption (kWh)
Current temperature (°C)
On-time / Operating time (hours)
Status and error flags (bitmask mapping according to Ebeco specifications)


src/drivers/ebeco500.xmq
```diff
@@ -0,0 +1,123 @@
+// Copyright (C) 2026 Magnus Friskytt (gpl-3.0-or-later)
+driver {
+    name           = ebeco500
+    meter_type     = ElectricityMeter
```
jktjkt, line 4: Interesting -- and that thing apparently *really* measures electricity as well.

jktjkt (commented):
I only eyeballed the code, I have not compiled the code, hence not an approval. But a +1 from me, and thanks for the contribution!

weetmuts:
Looks, great, but should they not be in drivers/src ? I.e. not src/drivers 
also there is already a hydrus.xmq driver can you check if your version improves on the existing one?

weetmuts:
The golvvärme meter has a non-compliant id, "000001e3" good thing wmbusmeters accepts that. But why....

magnfris:
> Looks, great, but should they not be in drivers/src ? I.e. not src/drivers also there is already a hydrus.xmq driver can you check if your version improves on the existing one?

ah yes I put it in src/drivers since I was testing with my "own" drivers. I guess I should change it then?
yes it is an improvement of the existing one (hydrus).

magnfris:
> The golvvärme meter has a non-compliant id, "000001e3" good thing wmbusmeters accepts that. But why....

Yes good question. I can't find any documentation on the Ebeco-website about this or their setup, but I got some info about what their status (error-codes) mean from technical sales support guy.

magnfris:
My setup is HomeAssistant with a https://iotlabs.pl/en/wm-bus-gateway and I use that with ESPHome Builder to send data via MQTT to Home Assistant so I can use the data as sensors. I want to see not only how much water I use, but also how much electricity my heated floors in my two bathrooms and laundry-room uses up. 
I have compiled and setup wmbusmeters on my Mac just to test out the new drivers before I try to incorporate them into the repository and hopefully they end up in the wm-bus-gateway eventually. I can read the most important values already now with others drivers but not all (less important) values. Thanks for a great repository by the way :)

weetmuts:
Fint, jag tog din kod och checkade in separat. Gjorde några fixar i din ebeco500 drivis. 

https://wmbusmeters.org/drivers/ebeco500.xmq.html

Sedan tittade jag på ditt hydrus telegram och förde in förändringar manuellt i den existerande hydrus drivisen.

Vet du var det är för skillnad på storage 8 som du bidrar med och storage 1 som redan fanns? Båda verkar rapportera 
någon slags förbrukning vid ett visst tidpunkt.

weetmuts:
Har du någon mer info om vad:   ROOM_OC,
                    label = ROOM_SC
                    label = FLOOR_OC
                    label = FLOOR_SC
                    label = RELAY_OC
                    label = RELAY_SC
                    label = RELAY_OT

betyder?

weetmuts:
Jag tror att du använde en AI för att generera drivisen. Lite mycket fria associationer. :-D

PR #2081, opened 2026-09-12
amiplus: add tariff 4, power upper limit and gama350 alias for Elgama Gama 350

mru12:
Follow-up to #2038 for the Elgama Elektronika Gama 350 (EGM,01,02), based on telegrams from a meter installed by PGE (so not only Tauron uses this meter).

The Gama 350 sends records that amiplus did not decode:

- `8E801003` / `8E8010833C`: energy consumption / production on tariff 4. Added `total_energy_consumption_tariff_4_kwh` and `total_energy_production_tariff_4_kwh`, same shape as the tariff 1-3 fields.
- `0BABC8FC10`: Power W + UpperLimit + AccumulationOfAbsoluteValue. Added as `power_upper_limit_kw`. It is constant per meter: 70 kW on both Gama 350 telegrams (this one and the one from #2036) and 0 on the Apator meters in the existing tests. It is not the contractual power limit (that is about 8 kW for this meter), so it is possibly the rated maximum power of the meter (3 x 230 V x 100 A). The meaning is not confirmed, happy to drop the field if you prefer not to expose it.
- Added `aliases = gama350` so the meter model can be used as driver name in the config, like `multical21` for kamwater.

`default_fields` is left untouched. Existing test expectations are updated where the telegrams already contained these records (5 Apator tests gain `power_upper_limit_kw: 0`, the Otus and the existing Gama 350 test gain the tariff 4 fields). `simulations/simulation_t1.txt` and `tests/test_list_envs.sh` are updated for the new fields.

Decrypted telegram (meter id anonymized), showing all three records:

BE44ED147856341201027A2700B0052F2F_066D7432945737800C78785634120E038301210200008E10038301210200008E20030000000000008E30030000000000008E8010030000000000000E833C0000000000008E10833C0000000000008E20833C0000000000008E30833C0000000000008E8010833C0000000000000B2B8901000BAB3C0000000BFB143200000BFB943C4901000AFDC9FC0126020AFDC9FC0225020AFDC9FC0326020BABC8FC100000072F2F2F2F2F2F2F2F2F2F2F2F

Verified against the real meter over several days: tariff 4 registers stay 0 (single tariff contract), the upper limit register stays 70 kW, all other values match the meter display. `make test` and `cd drivers && make pretty` are clean.


weetmuts:
Nice, but:

drop the alias name. Aliases are only used when merging an old driver into another driver (usually new) to maintain backward compatibilitity.

The difvifkey for  total_energy_production_tariff_4 can be replaced with the same matcher as total_energy_consumption...
just: add_combinable = BackwardFlow
It looks better if we decompose the difvifkeys into meaningful settings instead of just the hex.

rename power_upper_limit to  absolute_max_power_consumption_kw
the field was defined in the buggy sanxing meter which wants to be an amiplus compatible meter,
but I cannot see how it can be that with a non-compliant encryption....

Anyway, the meaning is unclear. Can you find out the meaning? Perhaps power_upper_limit is better.
But in the meantime use absolute_max_power_consumption
        field {
            name     = absolute_max_power_consumption
            quantity = Power
            info     = 'The maximum demand indicator for all times?'
            match {
                measurement_type = Instantaneous
                vif_range        = AnyPowerVIF
                add_combinable   = UpperLimit
                add_combinable   = AccumulationOfAbsoluteValue
            }
        }

mru12:
@weetmuts Thanks for the review, all three points are addressed in a3b48849:

- alias dropped
- `total_energy_production_tariff_4` now uses `AnyEnergyVIF` + `BackwardFlow` + `tariff_nr = 4`. While at it I converted the other production fields (total, current power, tariffs 1-3) from `difvifkey` to the same style, so the whole block reads consistently. Easy to revert if you prefer to keep that out of this PR.
- `power_upper_limit` renamed to `absolute_max_power_consumption` with the sanxing matcher and info text.

On the meaning: the register is 70 kW on both Gama 350 meters we have telegrams from, 0 on all Apator meters in the tests and 99 kW on the sanxing meter. The contractual limit for my meter is about 8 kW, so it is not that. 3 x 230 V x 100 A = 69 kW, which matches the rated maximum of a 100 A three phase meter, so I believe it is the meter's rated maximum power and not a measurement. I have no documentation confirming it though.

weetmuts:
Thanks!

Issue #2051, opened 2026-08-05
Qundis Qwater 5.5


docbobo:
### Type of request

Existing device improvement

### Meter description

I am having issues decoding the telegrams from my Qundis Qwater 5.5 devices. Basically, the decrypted telegram has zero data from the actual sensor readings. Just the status and dates.

### Logged telegrams from meter in decrypted format

```shell
telegram=|_49449344016135141907780DFF5F3500824800356D4BAA7A099116CB70DB31FBF87B7EBE135413AB9DBF83FF99646CFD520D445A4AFCF8883AE150462901DF4C125BA5E0046D1C0B4538|+4379
telegram=|_49449344016135141907780DFF5F35008248003576DB925B98559C67858C3FEEBB7C524FBBBA721D1579DD1E7AEB5B20CAB348823806F3F1442579B0BCD0D42F15828132046D1E0B4538|+4488
telegram=|_49449344016135141907780DFF5F350082480035FD16740E2C1243CED97E42F6226D171C45EBF99942B4F25E6D122C611A18395369593B089297F791D2FD2A4FDA0265F5046D1F0B4538|+4602
telegram=|_484493440161351419077A49003025EB2C30FA4A6199A716921643B57D4ED15E74894E5B6841D71E582DC26285C8BDB7C6C0AD7EDAEFC872E0DC50D39A72F6326CFFFF046D200B4538|+4618
telegram=|_49449344016135141907780DFF5F3500824900352F02472DA19B1E67C5324184A24A21E5DB54094EB9F36FD18B7F6B6C65F56D46892DC48A2CB6AD95F44BBB0A33E44741046D210B4538|+4715
```

folfy:
Seems to be a duplicate of #1916 (please check for already open issues, before creating a new one)

docbobo:
I did - that issue looked a) as if the problem was supposed to be solved a while ago and b) was marked stale. 

If a problem is not solved, I am not going to reopen and old, stale discussion - apologies if you would handle that differently. 

That said, Claude was suggesting that this might be an encryption issue with the walk-by packets. If someone can confirm that, I’ll happily close this right away.

folfy:
Your packets look totally the same as in #1916, and are decoding fine when thrown into the decoder (which is supposed to use the same modules), including the readings:
[...]

PR #2083, opened 2026-09-13
fix: warn once when an encrypted Qundis WalkByDataSet is found but no meter key is configured

bartman081523:
Addresses the gap noted in #2051: when a Qundis WalkByDataSet is encrypted (header byte[4]=0x35) and no meter key is configured, `try_qundis_decode` failed completely silently — the decode helper returned false, the guarded ixml grammar rejected the value, and no message was emitted anywhere. Users only saw empty walk by fields and, in `--analyze`, a bare `ixml:1:9: syntax error` pointing at the marker byte.

Encrypted walk-by telegram (from the existing AESTest vector in `drivers/src/qwaterv2.xmq`, synthetic key, self-contained):

```
49449344016135141907780DFF5F350082C100353E44AAFEDB237BDBA569404984A09F6DD12E32C2F03DD335E259DD446711256309CC4E894B4C7D8CC427C7491B94FF05046D160A4F38
```

With this PR, without a key, stderr now gets — once per meter (unless verbose or debug), mirroring the existing `"(wmbus) WARNING! no key to decrypt payload!"` idiom:

```
(qds) WARNING! no key to decrypt the encrypted WalkByDataSet! Walk by values are not decoded for id: 14356101
```

The warning also appears in `--analyze` runs, where until now the only indication was the ixml syntax error on the encrypted marker.

**Tested** with the new `tests/test_qwds_walkby_nokey.sh` (registered in test.sh), using the same telegram as `simulations/simulation_qwds_walkby.txt`:

- the same encrypted telegram sent twice with `NOKEY` warns exactly once (dedup, like test_key_warnings.sh),
- with the meter key the block decrypts and nothing is printed.

`qwaterv2`/`qheatv2` driver tests and `test_key_warnings.sh` still pass; the full test suite is green.

I verified against my own captured Q water 5.5 walk-by telegrams: encrypted frame without key → warning once, walk by fields stay empty (fail closed); with the key → decoded values, no warning.

🤖 This PR was prepared with [Claude Code](https://claude.com/claude-code); opened by @bartman081523.


weetmuts:
Very nice! Thanks!

PR #2088, opened 2026-09-14
Give compact profile points an actuality duration when no base time is sent

cgommel:
Follow-up to #2056 and #2055.

TAF7 profiles carry no valid base time — OMS Spec Vol.2 Annex R, R.3.2 makes the actuality duration of the base value the time reference instead. The spacing of the profile was read and then dropped after the expansion, so there was no way to place the points in time.

Each generated point now gets an actuality duration: the one of the base value plus one spacing per step. It only appears when the profile is spaced in seconds, minutes, hours or days, carries no base time, and has an actuality duration on its base storage, so profiles with a base time (`caltose`, `fhkvdataiv`, `uiws`, the Annex G test) are unaffected.

```
wmbusmeters --analyze=ehzp AE44a8153412711300027a010000002F2F0700285EB5050000000007803C34080000000000000728905F01000000000004208F84160101FD971D00C1017402C401208D841601C701031C76010000000000CD01831324410800010000000001000000000100000001000000000100000000010000000001000000C701833C0200000000000000CD0183933C244108000000000000000000000000000000000000000000000000000000000000000000
```

master gives `actuality_duration_at_1_s: 2` and nothing else. With this PR `at_1` to `at_8` read 2, 10, 18, 26, 34, 42, 50 and 58 seconds.

The telegram is from a real EMH meter; the 8 s spacing agrees with the 90 W the same telegram reports as current power. Covered by the `Elen4` case in `ehzp` and three assertions in `testinternals`; `make test` is green.


weetmuts:
Very nice! Thank you!

